PR #299, opened 2015-01-05
small fix

nikivazou:


ranjitjhala:
@nikivazou in case you missed it, this breaks the build...

PR #304, opened 2015-01-12
Constraints on Class Methods

nikivazou:
See [tests/neg/LiquidClass.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/d32f45167aaa336ff874f622a8993a7e67a51c61/tests/neg/LiquidClass.hs), [tests/pos/LiquidClass.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/d32f45167aaa336ff874f622a8993a7e67a51c61/tests/pos/LiquidClass.hs)
and [tests/neg/StateConstraints00.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/d32f45167aaa336ff874f622a8993a7e67a51c61/tests/neg/StateConstraints00.hs)

### DIFF
--- a/tests/neg/LiquidClass1.hs
+++ b/tests/neg/LiquidClass1.hs
@@ -11,7 +11,7 @@ class Compare a where
 instance Compare Int where	
 {-@ instance Compare Int where 
 	cmax :: Odd -> Odd -> Odd ;
-	cmin :: Int -> Int -> Int
+	cmin :: Int -> Int -> Odd
   @-}
 
 	cmax y x = if x >= y then x else y
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,11 @@
 - Experimental support for lifting haskell functions to measures
 If you annotate a Haskell function `foo` with {-@ measure foo @-}, LiquidHaskell will attempt to derive an equivalent measure from `foo`'s definition. This should help eliminate some boilerplate measures that used to be required.
 
+- Refining class instances. For example
+
+{-@ instance Compare Int where 
+    cmax :: Odd -> Odd -> Odd @-}
+
 ## 0.2.0.0
 
 - Move to GHC-7.8.3
--- a/README.md
+++ b/README.md
@@ -465,6 +465,24 @@ above the type class definition. For example (tests/pos/Class.hs)
 Any measures used in the refined class definition will need to be
 *generic* (see [Specifying Measures](#specifying-measures)).
 
+
+As an alternative, you can refine class instances. 
+For example (tests/pos/LiquidClass.hs)
+
+instance Compare Int where  
+
+{-@ instance Compare Int where 
+    cmax :: Odd -> Odd -> Odd
+  @-}
+
+    cmax y x = if x >= y then x else y
+
+When `cmax` method is used on `Int`, liquidHaskell will give it 
+the refined type `Odd -> Odd -> Odd`. 
+
+Note that currently liquidHaskell does not allow refining instances of
+refined classes. 
+
 Refinement Type Aliases
 -----------------------
 
--- a/README.md
+++ b/README.md
@@ -469,6 +469,7 @@ Any measures used in the refined class definition will need to be
 As an alternative, you can refine class instances. 
 For example (tests/pos/LiquidClass.hs)
 
+~~~~
 instance Compare Int where  
 
 {-@ instance Compare Int where 
@@ -476,6 +477,7 @@ instance Compare Int where
   @-}
 
     cmax y x = if x >= y then x else y
+~~~~
 
 When `cmax` method is used on `Int`, liquidHaskell will give it 
 the refined type `Odd -> Odd -> Odd`. 
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -180,6 +180,7 @@ Library
                     Language.Haskell.Liquid.Misc, 
                     Language.Haskell.Liquid.CoreToLogic,
                     Language.Haskell.Liquid.Variance,
+                    Language.Haskell.Liquid.Dictionaries,
                     Language.Haskell.Liquid.Qualifier, 
                     Language.Haskell.Liquid.TransformRec, 
                     Language.Haskell.Liquid.Tidy, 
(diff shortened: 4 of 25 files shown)


tests/neg/LiquidClass1.hs
```diff
@@ -0,0 +1,39 @@
...
+-- | Typing classes
+-- | Step 1: Refine type dictionaries:
+
+class Compare a where
+	cmax :: a -> a -> a
+	cmin :: a -> a -> a
+
+instance Compare Int where	
+{-@ instance Compare Int where 
+	cmax :: Odd -> Odd -> Odd ;
```
gridaphobe: Nitpick, I really dislike the `;` :/

tests/neg/LiquidClass1.hs
```diff
@@ -0,0 +1,39 @@
...
+  	cmin y x = if x >= y then x else y
+
+-- | creates dictionary environment:
+-- | * add the following environment
+-- | dictionary $fCompareInt :: Compare Int
+-- |               { $ccmax :: Odd -> Odd -> Odd
+-- |               , $ccmin :: Int -> Int -> Int
+-- |               }
+-- |
+-- | Important: do not type dictionaries, as preconditions
```
gridaphobe: I don't understand this comment. Does this mean that we don't check the instance definition against the strengthened types?

nikivazou: @gridaphobe we do check the instance definition. We do not check the code that creates the dictionary. 
In this example, the code that creates the dictionary is the following:

~~~
LiquidClass.$fCompareInt [InlPrag=[ALWAYS] CONLIKE]
   :: LiquidClass.Compare GHC.Types.Int
 [LclIdX[DFunId],
  Str=DmdType,
  Unf=DFun: \ ->
        LiquidClass.D:Compare TYPE GHC.Types.Int $ccmax $ccmin]
 LiquidClass.$fCompareInt =
   LiquidClass.D:Compare @ GHC.Types.Int $ccmax $ccmin,
~~~

gridaphobe:
How does this interact with the existing type-class logic, e.g. in

```haskell
class Compare a where
  max :: x:a -> y:a -> {v:a | v >= x && v >= y}

instance Compare Int where
  max :: Odd -> Odd -> Odd
```

what is the actual type given to `max @ Int`? I guess it would be

```
max @ Int :: x:Odd -> y:Odd -> {v:Odd | v >= x && v >= y}
```

though I'm not sure I like this. I think the above should actually be rejected as

```
Odd -> Odd -> Odd
  </:
x:a -> y:a -> {v:a | v >= x && v >= y}
```

gridaphobe:
Also, please update the README :)

tests/neg/LiquidClass1.hs
nikivazou: What is the alternative? 
We need to know where the instance types stop. Indentation would be a solution..... but it seems quite complicated

gridaphobe: Take a look at the parser for refined class declarations, I don't remember what I did there, but it seems to work without the `;`

nikivazou: Do you allow more that one refinement signatures though?

tests/neg/LiquidClass1.hs
gridaphobe: I don't follow, in the original type-class PR I checked the instance definition **by** checking that code, i.e. I refined `LiquidClass.D:Compare`.

nikivazou: Interesting, because now I just ignore that code, and nothing seems to break....

Have you tried in the original type classes to give in class methods preconditions?

tests/neg/LiquidClass1.hs
gridaphobe: Yep!

https://github.com/ucsd-progsys/liquidhaskell/blob/master/include/GHC/Real.spec

tests/neg/LiquidClass1.hs
gridaphobe: Yes, see https://github.com/ucsd-progsys/liquidhaskell/blob/master/tests/pos/Class.hs#L47-L49

tests/neg/LiquidClass1.hs
nikivazou: Where is this checked?

On Mon, Jan 12, 2015 at 11:05 AM, Eric Seidel <notifications@github.com>
wrote:

> In tests/neg/LiquidClass1.hs
> <https://github.com/ucsd-progsys/liquidhaskell/pull/304#discussion-diff-22812841>
> :
>
> > @@ -0,0 +1,39 @@
> > +module LiquidClass where
> > +
> > +
> > +-- | Typing classes
> > +-- | Step 1: Refine type dictionaries:
> > +
> > +class Compare a where
> > +	cmax :: a -> a -> a
> > +	cmin :: a -> a -> a
> > +
> > +instance Compare Int where	
> > +{-@ instance Compare Int where
> > +	cmax :: Odd -> Odd -> Odd ;
>
> Yep!
>
>
> https://github.com/ucsd-progsys/liquidhaskell/blob/master/include/GHC/Real.spec
>
> —
> <https://github.com/ucsd-progsys/liquidhaskell/pull/304/files#r22812841>.
>



-- 
Niki Vazou

gridaphobe: I'm not sure, I think `div` and `mod` are used in a few tests, but I doubt we have an instance declared anywhere. Regardless, it parses properly :)

nikivazou: The reason why I believe we need the ';' is that

If you go to your GHC.Real.Spec and move the type specification of
`fromIntegral` AFTER the Class specification, then the type specification
for `fromIntegral` would be interpreted as if it belonged to the class
methods,
which is not what we want.

On Mon, Jan 12, 2015 at 11:09 AM, Eric Seidel <notifications@github.com>
wrote:

> In tests/neg/LiquidClass1.hs
> <https://github.com/ucsd-progsys/liquidhaskell/pull/304#discussion-diff-22813162>
> :
>
> > @@ -0,0 +1,39 @@
> > +module LiquidClass where
> > +
> > +
> > +-- | Typing classes
> > +-- | Step 1: Refine type dictionaries:
> > +
> > +class Compare a where
> > +	cmax :: a -> a -> a
> > +	cmin :: a -> a -> a
> > +
> > +instance Compare Int where	
> > +{-@ instance Compare Int where
> > +	cmax :: Odd -> Odd -> Odd ;
>
> I'm not sure, I think div and mod are used in a few tests, but I doubt we
> have an instance declared anywhere. Regardless, it parses properly :)
>
> —
> <https://github.com/ucsd-progsys/liquidhaskell/pull/304/files#r22813162>.
>



-- 
Niki Vazou

PR #306, opened 2015-01-15
Break Bare.hs out into multiple modules

spinda:
This takes the big `Bare` module and splits it out into multiple submodules. The diff looks huge, but most of this is just taking things from `Bare.hs` and putting them in different places. Each submodule has been organized and exposes only the functions that other submodules need to be using, so different parts of the code don't end up depending on each other's internal functions.

The main functional changes are in consolidating `ofBareType`, `mkSpecType`, `expandAlias`, and so on into `Type.hs`. The traversal over `RType` is implemented once in `ofBRType`, which take two functions as parameters for how alias should be applied and how `r` should be resolved. This should all be backwards-compatible, with one exception: instead of silently discarding refinements applied to refined type aliases in sorts, **an error is generated on attempting to use these aliases in sort signatures**. This can be rolled back if necessary.

In order to simplify this code, the state passed around to keep track of cyclic type aliases was removed and replaced with a graph-based approach to detect cyclic aliases in `Bare/RTEnv.hs` before alias expansion begins. A graph of the type aliases is built, then errors are generated for each cyclic strongly connected component. This has the advantage of generating errors for all cyclic aliases instead of one at a time. It may also perform better&mdash;or at least, it shouldn't perform worse&mdash;since these checks are done all at once instead of duplicated over each expansion recursion. An extra crash test was added to make sure cylic aliases are caught properly.

There are a few rough patches that could be ironed out:
 - A few things I haven't found good places for yet, so they're in `Bare/Misc.hs`.
 - `Bare/Sort.hs` and possibly `Bare/Existential.hs` should be renamed, but we weren't able to come up with good names for them at this time.

All [tests](https://travis-ci.org/spinda/liquidhaskell/builds/44037766) passed the last time they ran, but I made a few changes after the last run that haven't been retested yet. If the Travis build breaks, I'll go and fix it.


src/Language/Haskell/Liquid/Bare/ConType.hs
```diff
@@ -0,0 +1,153 @@
+{-# LANGUAGE TupleSections #-}
+
```
spinda: rename to dataType?

src/Language/Haskell/Liquid/Bare/RTEnv.hs
```diff
@@ -0,0 +1,169 @@
...
+buildAliasNode :: AliasTable -> RTAlias Symbol BareType -> Node Symbol
+buildAliasNode table alias
+  = (rtName alias, rtName alias, traceShow (show $ rtName alias) $ buildAliasEdges table $ rtBody alias)
+
+buildAliasEdges :: AliasTable -> BareType -> [Symbol]
+buildAliasEdges table
+  = ordNub . go
+  where go (RApp (Loc _ c) ts rs _)
+          = curr ++ concat (map go ts ++ map go (mapMaybe ref rs))
+          where
```
nikivazou: not nested where:)

src/Language/Haskell/Liquid/Bare/RTEnv.hs
```diff
@@ -0,0 +1,169 @@
...
+        go (ROth _)
+          = []
+        go (RExprArg _)
+          = []
+        go (RHole _)
+          = []
+
+        go (RRTy _ _ _ _)
+          = errorstar "Bare.RTEnv.buildAliasEdges used with RRTy"
+
```
nikivazou: go (RRTy env r _ t) = concatMap (second go) env ++ go t

src/Language/Haskell/Liquid/Bare/RTEnv.hs
```diff
@@ -0,0 +1,169 @@
...
+    err scc@(rta:_)
+      = ErrAliasCycle { pos    = fst $ locate rta
+                      , acycle = map locate scc
+                      }
+    err []
+      = errorstar "Bare.RTEnv.checkCyclicAliases: No type aliases in reported cycle"
+
+    locate sym
+      = (sourcePosSrcSpan $ rtPos alias, pprint sym)
+      where (_, alias) = fromAliasSymbol table sym
```
nikivazou: same, no nested where

src/Language/Haskell/Liquid/Bare/Sort.hs
```diff
@@ -0,0 +1,79 @@
...
+import Control.Applicative ((<$>))
+
+import qualified Data.List as L
+
+import Language.Fixpoint.Misc (errorstar, safeZip, fst3, snd3)
+import Language.Fixpoint.Types (meet)
+
+import Language.Haskell.Liquid.RefType (appRTyCon, strengthen)
+import Language.Haskell.Liquid.Types
+
```
nikivazou: rename to SymSort (or Normalize)

src/Language/Haskell/Liquid/Bare/Spec.hs
```diff
@@ -0,0 +1,194 @@
...
+makeTExpr vs (_, spec) = varSymbols id "TermExpr" vs $ Ms.termexprs spec
+
+varSymbols :: ([Var] -> [Var]) -> Symbol ->  [Var] -> [(LocSymbol, a)] -> BareM [(Var, a)]
+varSymbols f _ vs  = concatMapM go
+  where lvs        = M.map L.sort $ group [(sym v, locVar v) | v <- vs]
+        sym        = dropModuleNames . symbol . showPpr
+        locVar v   = (getSourcePos v, v)
+        go (s, ns) = case M.lookup (val s) lvs of 
+                     Just lvs -> return ((, ns) <$> varsAfter f s lvs)
+                     Nothing  -> ((:[]).(,ns)) <$> lookupGhcVar s
```
nikivazou: you can remove unused arguments

src/Language/Haskell/Liquid/Bare/Type.hs
```diff
@@ -0,0 +1,229 @@
+{-# LANGUAGE TupleSections #-}
+
+module Language.Haskell.Liquid.Bare.Type (
+    ofBareType
```
nikivazou: rename module to OfType

src/Language/Haskell/Liquid/WiredIn.hs
```diff
@@ -11,14 +11,30 @@ import Language.Fixpoint.Names                  (hpropConName, propConName)
 import Language.Fixpoint.Types
 import Language.Fixpoint.Misc                   (mapSnd)
 
-import TyCon
```
nikivazou: maybe this can go inside Bare

spinda:
The build seems to have errored due to an issue with [Travis](https://travis-ci.org/ucsd-progsys/liquidhaskell/jobs/47183394#L312). It'll re-run once I make the changes noted in the comments and rebase to `master`.

spinda:
@nikivazou I've made the changes we went over today. Tomorrow (well, later today) I'll do the rebasing.

spinda:
Rebased! Also added in a tiny cleanup of `Dictionaries` and a cleanup of the "not nice" `ReplaceM` in `Bare/GhcSpec.hs`.

spinda:
Yay!

nikivazou:
@spinda, I get some tracing 

~~~
Trace: ["BNat"] : ["Nat"]

Trace: ["Odd"] : []

Trace: ["Even"] : []

Trace: ["IncrListD"] : []

Trace: ["Nat"] : []

Trace: ["LeInt"] : []

Trace: ["Nat64"] : []

Trace: ["GeInt"] : []

Trace: [symOrder: ] : ["Nat","BNat","Odd","Even","IncrListD","LeInt","Nat64","GeInt"]
~~~

Is it leftover from your edits? *If so*, can you please remove it?
Btw, since this is usual, it is good to give more "unique" names to tracing, so that it is easier to detect them.

spinda:
Oh, wow, very sorry about that! I thought I got rid of that already. Removed in #309.

PR #307, opened 2015-01-16
Improve error for refined class/instances conflict

spinda:
```haskell
module Crash where

class Foo a where
{-@ class Foo where
      foo :: x:a -> {v:a | v = x}
  @-}
  foo :: a -> a       


instance Foo Int where
  {-@ instance Foo Int where
       foo :: x:Int -> {v:Int | v = x + 1} @-}
  foo x = x + 1

instance Foo Integer where
  {-@ instance Foo Integer where
       foo :: x:Integer -> {v:Integer | v = x + 1} @-}
  foo x = x + 1
```

```
 tests/crash/RClass.hs:4:11: Error: Refined classes cannot have refined instances
     Refined class definition for: Foo
     Defined at: tests/crash/RClass.hs:4:11
      
     Refined instance for: Int
     Defined at: tests/crash/RClass.hs:11:16
      
     Refined instance for: Integer
     Defined at: tests/crash/RClass.hs:16:16
```
vs

```
 tests/crash/RClass.hs:4:11: Error: You cannot refine instances of refined classes Foo
```

### DIFF
--- a/src/Language/Haskell/Liquid/Bare.hs
+++ b/src/Language/Haskell/Liquid/Bare.hs
@@ -45,6 +45,7 @@ import Text.Printf
 import Control.Monad.State      (get, gets, modify, put, State, evalState, evalStateT, execState, execStateT, StateT)
 import Data.Traversable         (forM)
 import Control.Applicative      ((<$>), (<*>), (<|>))
+import Control.Arrow            ((&&&))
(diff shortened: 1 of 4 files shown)


nikivazou:
Thanks!!!

gridaphobe:
Excellent, thanks!

PR #318, opened 2015-01-23
Maps: adding support for select/store

ranjitjhala:
**Do not merge** until travis says ok!

See 

https://github.com/ucsd-progsys/liquidhaskell/blob/maps/tests/pos/maps.hs
https://github.com/ucsd-progsys/liquidhaskell/blob/maps/tests/pos/maps1.hs
https://github.com/ucsd-progsys/liquidhaskell/blob/maps/tests/neg/maps.hs

also depends on matching PR in liquid-fixpoint https://github.com/ucsd-progsys/liquid-fixpoint/pull/43

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
  - export PATH="$HOME/.cabal/bin:/opt/ghc/7.8.3/bin:/opt/cabal/1.18/bin:$PATH"
  - cabal update
  - git clone git://github.com/ucsd-progsys/liquid-fixpoint.git /tmp/fixpoint
- - pushd /tmp/fixpoint && cabal install && popd
+ - pushd /tmp/fixpoint && cabal install -fbuild-external && popd
  - curl "http://goto.ucsd.edu/~gridaphobe/$SMT" -o "$HOME/.cabal/bin/$SMT"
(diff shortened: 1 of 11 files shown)


ranjitjhala:
I can get this working as shown in the tests above -- i.e. by DIRECTLY defining the measures and embedding, but NOT if I put the relevant definitions in `include/Data/Map.spec` and then importing,
as in 

https://github.com/ucsd-progsys/liquidhaskell/blob/maps/tests/todo/maps.hs

after renaming 

https://github.com/ucsd-progsys/liquidhaskell/blob/maps/include/Data/Map.hiddenspec

There is some weirdness going on with the name resolution?

ranjitjhala:
@gridaphobe the travis thing fails because I guess its running against the OLD fixpoint and not the new one. Does this mean I need to first merge fixpoint before running the tests here?

gridaphobe:
@ranjitjhala ah yes, that would make sense, travis is configured to always build fixpoint/master. I'd go ahead and merge the fixpoint PR since it doesn't look like it could break anything on LH's end. Then we can retrigger the tests here.

gridaphobe:
@ranjitjhala regarding the name-resolution, that's very odd indeed. We'll have to look at the verbose log to figure out what's going on..

Issue #320, opened 2015-01-28
Allow `CoreToLogic` definitions inside type specifications


ranjitjhala:
See:

https://github.com/ucsd-progsys/liquidhaskell/blob/master/tests/todo/coretologic.hs

We can use `empty` in measure definitions but NOT in vanilla predicates and types.

nikivazou:
fixed with #321

PR #321, opened 2015-01-30
Core to logic

nikivazou:
fix for #320


ranjitjhala:
Thanks!

Can / did you move the test into pos/ ?

PR #323, opened 2015-01-31
Strengthened the AVL insert function to remove unnecessary guards

michaelbeaumont:
I added a predicate that strengthens the postconditions on `insert` so, to satisfy the totality checker, we don't need the case checks I removed.

I feel like the type on line 21 shouldn't be required somehow, since by the definition of `balanced`, the absolute value of `bFac` for an `AVLTree` must be 0 or 1.

Probably related to that, the following type checks:

    force :: Tree a -> Tree a
    {-@ force :: forall <p :: Tree a -> Prop>. t:AVLTree a<p> -> {v:AVLTree a<p> | bFac v <= 1 && bFac v >= -1} @-}
    force t@(Nil) = t
    force t@(Tree v l r) = t

while this does not:

    force :: Tree a -> Tree a
    {-@ force :: forall <p :: Tree a -> Prop>. t:AVLTree a<p> -> {v:AVLTree a<p> | bFac v <= 1 && bFac v >= -1} @-}
    force t = t

I guess this is because on the constructors Branch and Nil, `balanced` is expanded to the whole definition, while on a general `Tree a`, we only know `balanced`. Could it be possible to try to automatically expand `t` to both constructors and then deduce it that way?

### DIFF
--- a/tests/pos/AVLRJ.hs
+++ b/tests/pos/AVLRJ.hs
@@ -26,7 +26,6 @@ bFac (Tree _ l r) = ht l - ht r
 htDiff :: Tree a -> Tree a -> Int
 htDiff l r = ht l - ht r
 
-
 -- | Empty 
 {-@ empty :: {v: AVLTree a | ht v == 0} @-}
 empty = Nil
@@ -36,7 +35,10 @@ empty = Nil
 singleton a = Tree a Nil Nil
 
 -- | Insert 
-{-@ insert :: a -> s: AVLTree a -> {t: AVLTree a | EqHt t s || HtDiff t s 1 } @-}
+
+{-@ predicate PostInsert S T = ((bFac T == 0) => (EqHt T S || ht T == 1)) && (EqHt T S || HtDiff T S 1) @-}
+
+{-@ insert :: a -> s: AVLTree a -> {t: AVLTree a | PostInsert s t } @-}
 insert :: (Ord a) => a -> Tree a -> Tree a
 insert a Nil = singleton a
 insert a t@(Tree v _ _) = case compare a v of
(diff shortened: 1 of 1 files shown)


nikivazou:
@michaelbeaumont this is indeed an issues we have come across many times. 
In your second example (@force t = t@) liquidHaskell only knows that the result is equal to the input. 
While in the first the tree elements (ie, @v@, @l@, and @r@) come into scope and liquidHaskell is aware of all the properties that you defined on trees. 

Automatic expansion of your first example to the second is a good suggestion, but for now we should do it manually...

Also, the abstract refinement @p@ that you have in your types is not required: it is implied by the polymorphic type variable @a@ that whatever properties the elements of the inputs tree have will also hold for the elements of the output tree. Does this make sense?

michaelbeaumont:
I was working with a non polymorphic version and I kind of just copied it over to this version so it might not be necessary anymore, but `p` here is more general than a refinement on `a`, it's a refinement on `Tree a` right?

So if I'm in a function and hand `force` a `Tree a` whose height I know is less than 2, for example, if I don't have that extra `p`, don't I "lose" that knowledge on the return value of `force`, because `force` could conceivably just return `Nil`, which is an `AVLTree a` with `bFac == 0`.

Probably I screwed up the syntax with `AVLTree a<p>` when I copy and pasted...

nikivazou:
If you want the `p` to be on the tree, then you should write `AVLTree <p> a` (I know this syntax is a bit weird...)

You are right that you will loose this information, but you could just say that `force` returns its argument:

~~~
force :: t:AVLTree a -> {v:AVLTree a | bFac v <= 1 && bFac v >= -1 && v = t}
~~~~

which is a type more precise than the one with the abstract refinement `p`.

PR #341, opened 2015-02-14
Uncommented GHC.Num.negate spec

michaelbeaumont:
I don't know why this was commented out but it fixes the issue of having to write `(0-1)` everywhere in code instead of `-1`.

This is faster than fixing it at the CoreToLogic level. Were there some kind of problems with the spec?


ranjitjhala:
Thanks @michaelbeaumont ! We'll look into this, I can't remember why it was commented out...

@gridaphobe why does travis say "build failed" and yet  "All 320 tests passed (464.30s)" ? 

https://travis-ci.org/ucsd-progsys/liquidhaskell/jobs/50776054

PR #342, opened 2015-02-15
Wchecking

nikivazou:
Adding some well formedness checking on Abstact Refinements


ranjitjhala:
Super!!!! I think this may make it easier to do the "while" and other examples.

PR #343, opened 2015-02-15
Added ENeg type

michaelbeaumont:
LH wasn't able to lift functions with expressions containing `(-1)` into measures (see `tests/pos/AVLRJ.hs` for an example), so I added a type constructor `ENeg` to represent this to `Liquid.Fixpoint.Types.Expr`. See ucsd-progsys/liquid-fixpoint#50

### DIFF
--- a/src/Language/Haskell/Liquid/Bare/Expand.hs
+++ b/src/Language/Haskell/Liquid/Bare/Expand.hs
@@ -104,6 +104,8 @@ expandExpr (EApp f@(Loc l f') es)
          Nothing ->
            EApp f <$> mapM expandExpr es
 
+expandExpr (ENeg e)
+  = ENeg <$> expandExpr e
 expandExpr (EBin op e1 e2)
   = EBin op <$> expandExpr e1 <*> expandExpr e2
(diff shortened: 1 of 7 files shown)


gridaphobe:
This seems fine to me (but travis will fail to build this PR until the related fixpoint PR is merged)

ranjitjhala:
Thanks Michael!!

@gridaphobe : will Travis pick up the new fixpoint or do we need to prod it somehow?

gridaphobe:
@ranjitjhala we'll need to restart the build once after we merge the fixpoint PR.

@michaelbeaumont btw, I take it your motivating example is https://github.com/ucsd-progsys/liquidhaskell/blob/master/tests/pos/AVLRJ.hs#L96 ? Would you mind adding a separate test file with the predicate rewritten as a haskell function, so we have a concrete test for the new functionality?

michaelbeaumont:
It was with functions lifted to measures (CoreToLogic.hs), for example: https://github.com/michaelbeaumont/liquidhaskell/blob/fixNegate/tests/pos/AVLRJ.hs#L101

PR #348, opened 2015-02-22
String

nikivazou:
Some string interpretation: 
Strings are grapped from the code and modeled as Uninterpreted Functions, see

[tests/pos/Strings.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/b0d752540012d7991590973f357451017b8bb0df/tests/pos/Strings.hs)


spinda:
Wow, nice! Would it be possible/better to use quotes for "foo" in the type
signature? On a quick glance, I wonder how this handles things like spaces
without them.

spinda:
I'm also curious about how this affects name resolution.

nikivazou:
Well spaces etc are not handled for now.....

If you want you can play around, otherwise lets discuss after Friday:)

ranjitjhala:
Niki -- this can cause problems with name resolution as pointed out. It may be better to use the new support for "literals" in fixpoint ...

gridaphobe:
@nikivazou @ranjitjhala I have a branch that adds support for strings via fixpoint literals from long ago when I was looking at the dynamic dictionary stuff.

https://github.com/ucsd-progsys/liquidhaskell/commits/fay

(I also agree that the strings need to be quoted to avoid wreaking havoc with name resolution)

PR #353, opened 2015-03-04
makeSyms

nikivazou:
DO NOT MERGE!

Trying to find out what `makeSymbols` does


spinda:
For the `mkVarExpr` substitution, instead of going through `makeSymbols`, we can just ask GHC for all of the data constructor IDs: `getModuleInfo` -> `modInfoTyThings` -> filter for `AConLike` and then `RealDataCon`.

spinda:
`freeSyms` can be most likely populated in the same way, as it only seems to include data constructors as well.

PR #356, opened 2015-03-08
Resolve Exprs in RExprArgs

spinda:
We do name resolution on `Expr`s inside `UReft`s, but not in `RExprArg`s in the `RType`s.

### DIFF
--- a/src/Language/Haskell/Liquid/Bare/OfType.hs
+++ b/src/Language/Haskell/Liquid/Bare/OfType.hs
@@ -46,15 +46,15 @@ import Language.Haskell.Liquid.Bare.Resolve
 
 ofBareType :: SourcePos -> BareType -> BareM SpecType
 ofBareType l
-  = ofBRType expandRTAliasApp (resolve l <=< expandReft)
+  = ofBRType expandRTAliasApp (resolve l <=< expandReft) (resolve l <=< expandExpr)
 
 ofMeaSort :: BareType -> BareM SpecType
 ofMeaSort
-  = ofBRType failRTAliasApp return
+  = ofBRType failRTAliasApp return return
 
 ofBSort :: BSort -> BareM RSort
 ofBSort
-  = ofBRType failRTAliasApp return 
+  = ofBRType failRTAliasApp return return
 
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


src/Language/Haskell/Liquid/Bare/OfType.hs
```diff
@@ -100,9 +102,10 @@ rtypePredBinds = map uPVar . ty_preds . toRTypeRep
 ofBRType :: (PPrint r, Reftable r)
          => (SourcePos -> RTAlias RTyVar SpecType -> [BRType r] -> r -> BareM (RRType r))
          -> (r -> BareM r)
+         -> (Expr -> BareM Expr)
          -> BRType r
          -> BareM (RRType r)
-ofBRType appRTAlias resolveReft
+ofBRType appRTAlias resolveReft resolveExpr
```
gridaphobe: Looks to me like this extra parameter is not needed, why not just inline the `resolve <=< expandExpr` bit?

spinda: Because `resolve` takes a `SourcePos` parameter which you don't always have. I looked into removing it a while back but concluded that that needed a small change to Fixpoint (which I could do, if it's wanted). A bit fuzzy on the details there at the moment.

spinda: If I remember correctly, the same change to Fixpoint is needed to supply the `SourcePos` information in all locations where the `OfType` family of functions is called. I'll look into it in more detail tomorrow.

spinda: An alternative is to add location information to the `MSort` datatype in `Measure.hs` (or to `RExprArg` directly...), which would not require a FIxpoint change. Which do you prefer? Or would you go with something else entirely?

gridaphobe: Ah quite right, I saw the `Loc` in the `RApp` case and I guess I stopped reading :)

I suspect the `Loc` is used for error reporting, in which case we shouldn't just remove it outright. (There's another issue here somewhere that talks about adding `Loc`s to all of the datatypes we use and making `Located` a type-class, but I can't find it in github's mobile site..)

spinda: What I mean by "remove" it is that, if some change were made to Fixpoint
(still need to retrace back and figure out what it is exactly), we could
derive the location information in all cases from just the data being
`resolve`d.

I think it had to do with changing one of the `Symbol`s in `Expr` to
`LocSymbol`.

spinda: It's changing `EVar` to take a `LocSymbol` instead of a `Symbol`, but now I remember why I decided it wouldn't work out. It ends up requiring a large, cascading series of changes to Fixpoint, adding location information requirements where there weren't any before.

spinda:
I've updated it to embed the needed location information in `RExprArg`.

PR #357, opened 2015-03-09
Fix name resolution for local specifications in imported modules

### LINKED ISSUE #354: Name resolution problem with non-top level functions
This is a major problem that is blocking the 230 Final :( 

see `tests/todo/MM.hs` (which imports `tests/todo/LL.hs`) and fails with:

```
 /Users/rjhala/research/liquid/liquidhaskell/tests/todo/LL.hs:8:9: Error: GHC Error
     Not in scope: variable `LL.go'
```

(the problem goes away if I just make `go` a top-level function.)

I'm guessing this has to do with @spinda's recent changes?

spinda:
Fixes #354.

### DIFF
--- a/src/Language/Haskell/Liquid/Bare/Check.hs
+++ b/src/Language/Haskell/Liquid/Bare/Check.hs
@@ -5,7 +5,6 @@
 module Language.Haskell.Liquid.Bare.Check (
     checkGhcSpec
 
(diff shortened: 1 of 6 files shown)


spinda:
This should be ready to merge as soon as Travis gives the okay.

PR #358, opened 2015-03-10
Adding better syntax for bounds

nikivazou:
See for the new bound syntax [benchmarks/icfp15/pos/Overview.lhs](https://github.com/ucsd-progsys/liquidhaskell/blob/bounds/benchmarks/icfp15/pos/Overview.lhs)

### DIFF
--- a/src/Language/Haskell/Liquid/Bare/OfType.hs
+++ b/src/Language/Haskell/Liquid/Bare/OfType.hs
@@ -98,7 +98,7 @@ rtypePredBinds = map uPVar . ty_preds . toRTypeRep
 
 --------------------------------------------------------------------------------
 
-ofBRType :: (PPrint r, Reftable r)
+ofBRType :: (PPrint r, UReftable r)
          => (SourcePos -> RTAlias RTyVar SpecType -> [BRType r] -> r -> BareM (RRType r))
          -> (r -> BareM r)
          -> BRType r
--- a/src/Language/Haskell/Liquid/Bare/OfType.hs
+++ b/src/Language/Haskell/Liquid/Bare/OfType.hs
@@ -103,8 +103,8 @@ ofBRType :: (PPrint r, UReftable r)
          -> (r -> BareM r)
          -> BRType r
          -> BareM (RRType r)
-ofBRType appRTAlias resolveReft t
-  = go t
+ofBRType appRTAlias resolveReft
+  = go
   where
     go t@(RApp _ _ _ _)
       = do aliases <- (typeAliases . rtEnv) <$> get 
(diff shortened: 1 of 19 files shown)


src/Language/Haskell/Liquid/Parse.hs
```diff
@@ -417,6 +418,22 @@ funArgsP  = try realP <|> empP
...
 
+boundP = do 
+  name   <- locParserP upperIdP
+  params <- many (parens tyBindP)
+  reservedOp "="  
+  args   <- bargsP
+  body   <- predP
+  return $ Bound name params args body  
+ where 
+    bargsP = try ( do reservedOp "\\"
```
gridaphobe: You probably don't need the `try` here.

src/Language/Haskell/Liquid/Bare/OfType.hs
```diff
@@ -148,6 +143,24 @@ ofBRType appRTAlias resolveReft
     go_syms
       = secondM ofBSort
 
+    goRFun bounds _ (RApp c ps _ _) t | Just bnd <- M.lookup c bounds 
+      = makeBound bnd (bar <$> ps) <$> go t
+    goRFun _ x t1 t2
+      = rFun x <$> go t1 <*> go t2
+
+    bar (RVar x _) = x
+    bar _          = errorstar "BAR"
```
gridaphobe: Do we have code anywhere to `catch` the exceptions thrown by `errorstar` so we can print out the panic-please-report-the-bug message?

spinda: `errorstar` is just a [wrapper around `error`](https://github.com/ucsd-progsys/liquid-fixpoint/blob/b1592e1d8189a8f8ba5c2920e0c8eaffdc9e607d/src/Language/Fixpoint/Misc.hs#L88).

`ofBRType` is already in `BareM`, so we should throw a proper `Error` via `throwError` instead of using `errorstar` here.

gridaphobe: Right, but we could also have a "top-level" `catch` somewhere to catch the exception from `error`.

spinda: Ah, right. I misread your comment.

ranjitjhala: + @spinda is correct, can we please **not** use error/errorstar **ever**,
especially inside BareM, where `throwError` is there.

On Tue, Mar 10, 2015 at 2:24 PM, Michael Smith <notifications@github.com>
wrote:

> In src/Language/Haskell/Liquid/Bare/OfType.hs
> <https://github.com/ucsd-progsys/liquidhaskell/pull/358#discussion_r26167800>
> :
>
> > @@ -148,6 +143,24 @@ ofBRType appRTAlias resolveReft
> >      go_syms
> >        = secondM ofBSort
> >
> > +    goRFun bounds _ (RApp c ps _ _) t | Just bnd <- M.lookup c bounds
> > +      = makeBound bnd (bar <$> ps) <$> go t
> > +    goRFun _ x t1 t2
> > +      = rFun x <$> go t1 <*> go t2
> > +
> > +    bar (RVar x _) = x
> > +    bar _          = errorstar "BAR"
>
> errorstar is just a wrapper around error
> <https://github.com/ucsd-progsys/liquid-fixpoint/blob/b1592e1d8189a8f8ba5c2920e0c8eaffdc9e607d/src/Language/Fixpoint/Misc.hs#L88>
> .
>
> ofBRType is already in BareM, so we should throw a proper Error via
> throwError instead of using errorstar here.
>
> --
> <https://github.com/ucsd-progsys/liquidhaskell/pull/358/files#r26167800>.
>



-- 
Ranjit.

PR #361, opened 2015-03-12
Use Travis's standard copy of GHC 7.8 instead of installing our own

spinda:
I've been communicating with Travis-CI support, trying to work out the issue we're having with `cabal install` randomly erroring out during builds. They're not sure what the problem is, but they recommend using the standard Travis's standard copies of GHC and Cabal instead of installing our own, to eliminate that as a possible source.

It looks like we're currently using a modified version of [hvr/multi-ghc-travis](https://github.com/hvr/multi-ghc-travis), but it doesn't look like we need to anymore. We're not currently testing against multiple versions of GHC, and [Travis supports that natively now](http://docs.travis-ci.com/user/languages/haskell/) anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-# NB: don't set `language: haskell` here
+language: haskell
+
 env:
  - SMT=z3 TESTS=Unit/pos
  - SMT=z3 TESTS=Unit/neg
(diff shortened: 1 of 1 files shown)


gridaphobe:
When I originally set up travis they were using ghc 7.8.2, but we need >=7.8.3 for API compatibility reasons, but I see they're on 7.8.4 now :)

My only concern would be accommodating the upcoming 7.10 release. I have a branch that removes our custom version of GHC's desugarer and I'd like to merge it in once the release is live.

spinda:
In that case, you can always roll back these changes when you merge in 7.10
support. Until then, we can find out whether or not this setup makes our
Travis issues go away.

spinda:
Is this okay to merge, or does it need further discussion?

PR #362, opened 2015-03-19
Temporarily enable verbose logging for Cabal commands on Travis

spinda:
I believe the cause of our random build failures is an interuption of network connectivity while we're downloading packages from Hackage. Unfortunately, Cabal doesn't print out a nice error message for this by default, and I can't reproduce the issue outside of Travis. Thus, to confirm that this is the case, we'd need to temporarily enable verbose logging for our Cabal commands in Travis builds.

If this *is* the cause of the issue, it can be worked around by adding `travis_retry` before our Cabal commands. First, though, I'd like to be able to provide Travis support with verbose log data from an errored-out build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,12 +23,12 @@ before_install:
  - travis_retry sudo apt-get install ocaml camlidl
  - travis_retry cabal update
  - git clone git://github.com/ucsd-progsys/liquid-fixpoint.git /tmp/fixpoint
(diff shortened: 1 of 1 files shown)


gridaphobe:
Let's go ahead and add travis_retry, I've seen a number of network failures in the build logs.

spinda:
#363

PR #363, opened 2015-03-19
Work around Travis issues with travis_retry

spinda:
Random network failures during Cabal dependency installs break Travis builds. Work around this by adding travis_retry to commands that install dependencies.

See #362.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,8 @@ script:
  - cabal build
  - cabal copy
  - cabal register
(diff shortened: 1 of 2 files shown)


spinda:
I'm submitting this as a pull request first to make sure that `.travis.yml` is still valid after my changes, instead of potentially messing up builds on `master`.

spinda:
I've added another workaround to prevent Travis from timing out tests that take > 10 minutes to complete.

spinda:
This should be ready to merge now. I've split out the Travis code into a separate script and added some helper functions to work around the issues we've been running into.

gridaphobe:
Cool, thanks @spinda!

PR #364, opened 2015-03-27
Ghc 7.10

gridaphobe:
**DON'T MERGE YET**

This branch makes LH compatible with GHC 7.10 and removes our custom desugarer (yay!), opting instead to use the new SourceNote Ticks to extract locations. Furthermore, since the GHC api didn't change very much this time, I was able to insert a small shim between us and GHC so we're compatible with 7.8 **and** 7.10.

There's some ugliness yet to resolve:
1. For some reason looking up `GHC.Integer.Type.Integer` fails, so I've had to remove the qualification..
2. MANY warnings, mostly from changes in GHC's api and the AMP. I don't think -Werror is going to be feasible if we want to support two GHCs.


src/Language/Haskell/Liquid/Types.hs
```diff
@@ -458,8 +458,9 @@ instance Hashable (PVar a) where
 ------ Strictness --------------------------------------------------
 --------------------------------------------------------------------
 
-instance NFData Var
-instance NFData SrcSpan
+instance NFData Var where rnf _ = ()
```
gridaphobe: I forgot about this, the new version of `NFData` requires a `Generic` instance for automatic derivation, but GHC doesn't supply them for most types :disappointed:

spinda:
> I was able to insert a small shim between us and GHC so we're compatible with 7.8 and 7.10.

Nice!

src/Language/Haskell/Liquid/Types.hs
spinda: Would a standalone deriving instance help here?

gridaphobe: Unfortunately not, because standalone deriving requires the data constructors to be in scope, which they are not (at least for `Var`).

spinda: Ah, here's something from the `deepseq` docs:

> If you need the previously used exact default 'rnf' method implementation semantics, use
>
> ```haskell
> instance NFData Colour where rnf x = seq x ()
> ```
>
> or alternatively
>
> ```haskell
> {-# LANGUAGE BangPatterns #-}
> instance NFData Colour where rnf !_ = ()
> ```

gridaphobe: Oh wow, no wonder they changed it, that's not even correct...

nikivazou: IF we have to choose between two GHC versions and Werrors, the Werrors flag
is more useful, especially in long term.

gridaphobe: I disagree, I don't find Werror to be particularly useful in Haskell where most of the warnings are cosmetic or stylistic. 

In contrast, not everyone will rush to upgrade to 7.10; so moving to 7.10 will prevent people from using LH.

nikivazou: So when I added the Werrors flag I came across some real errors, ie not
stylistic warnings, and now that they are fixed I don't want to allow them
again.

Just create a 7.10 branch for the ones that rash to upgrade.

gridaphobe: I take that back. Looking at the list of warnings that GHC supports, the standard warnings seem completely reasonable. However, the additional warnings enabled by `-W` (-fwarn-unused-binds, -fwarn-unused-matches, -fwarn-unused-imports, -fwarn-incomplete-patterns, -fwarn-dodgy-exports, -fwarn-dodgy-imports) don't seem particularly useful with the exception of incomplete-patterns.

Incidentally, most of the new warnings are only triggered by `-W`, so maybe we can just remove that :)

spinda: If I remember correctly, -fwarn-unused-matches and -fwarn-unused-binds have
caught a few bugs in the past. We should at least keep those around.

gridaphobe:
Ok, I've updated it so we only disable warnings about unused imports and deprecations (the latter is necessary to avoid having to deal with the changes to overlapping instances).

gridaphobe:
@spinda how's the modular verification stuff coming along? It appears that GHC 7.10 is wreaking havoc with our name resolution hacks, and I'd rather not fiddle with that stuff unless I need to :)

spinda:
I'm still working away at it. It's going to be a bit still before I can get
everything resolved.

gridaphobe:
@nikivazou 2 of the remaining test failures are due to partiality in `Constraint.Generate.splitC`. `tests/pos/StateConstraints0.hs` triggers https://github.com/ucsd-progsys/liquidhaskell/blob/ghc-7.10/src/Language/Haskell/Liquid/Constraint/Generate.hs#L527-L528

```
**** ERROR *********************************************************************
**** ERROR *********************************************************************
**** ERROR *********************************************************************
(Another Broken Test!!!) splitc unexpected: ds_dVC:{VV#1360 : (Compose.ST {VV#1343 : s^"lq_tmp_x1345" | k_1344} {VV#1346 : a^"lq_tmp_x1348" | k_1347})^"lq_tmp_x1362" | k_1361}
-> f#aS7:{VV#1381 : (Compose.ST {VV#1364 : s^"lq_tmp_x1366" | k_1365} {VV#1367 : b^"lq_tmp_x1369" | k_1368})^"lq_tmp_x1383" | k_1382}
-> {VV#1402 : (Compose.ST {VV#1385 : s^"lq_tmp_x1387" | k_1386} {VV#1388 : b^"lq_tmp_x1390" | k_1389})^"lq_tmp_x1404" | k_1403}

{x :: {VV : s | papp1 pre VV}, y :: {VV : s | papp2 postg VV x} |- {VV : b | papp2 rf VV y} <: {VV : b | papp2 r VV x}} =>
{xx :: {VV : s | papp1 pre VV}, w :: {VV : s | papp2 postg VV xx} |- {VV : s | papp2 postf VV w} <: {VV : s | papp2 post VV xx}} =>
{ww :: {VV : s | papp1 pre VV} |- {VV : s | papp2 postg VV ww} <: {VV : s | papp1 pref VV}} =>
lq_tmp_db48:(Compose.ST s a)
-> lq_tmp_db49:(Compose.ST s b) -> (Compose.ST s b)
**** ERROR *********************************************************************
**** ERROR *********************************************************************
**** ERROR *********************************************************************
```

`tests/pos/propmeasure.hs` triggers https://github.com/ucsd-progsys/liquidhaskell/blob/ghc-7.10/src/Language/Haskell/Liquid/Constraint/Generate.hs#L498-L499

```
**** ERROR *********************************************************************
**** ERROR *********************************************************************
**** ERROR *********************************************************************
Predicate in lhs of constraint:{lq_tmp_x232 : [a] | (null lq_tmp_x232 <=> true) && len lq_tmp_x232 == 0 && length lq_tmp_x232 == 0 && (nonEmpty lq_tmp_x232 <=> false) && (lenEqFive lq_tmp_x232 <=> false) && (lenNEqFive lq_tmp_x232 <=> true) && (lenGEFour lq_tmp_x232 <=> false)}
<:
[a]
**** ERROR *********************************************************************
**** ERROR *********************************************************************
**** ERROR *********************************************************************
```

Any idea what might have changed in ghc 7.10 that could cause these errors? (I've barely touched the LH code itself.)

nikivazou:
It seems that they changed the generation of the intermediate code, and that I was making some assumptions about the intermediate code and these assumptions do not hold anymore :(

spinda:
FYI, Travis has built-in support for GHC 7.10. You have to put it in quotes, though, so `ghc: '7.10'` (otherwise it's interpreted as a decimal `7.1`).

gridaphobe:
@nikivazou do you have a link that describes the changes?

nikivazou:
What changes?

gridaphobe:
In the intermediate code generation

gridaphobe:
@spinda I tried reverting to travis' regular haskell infrastructure, but it doesn't seem to find ghc 7.10

```
$ export PATH=/usr/local/ghc/$(ghc_find 7.10)/bin/:$PATH
ghc_find: error, no such version 7.10
```

Perhaps 7.10 is only available on the newer container-based bots?

spinda:
Can you try "7.10.1"?

gridaphobe:
@spinda I tried using `'7.10.1'` too but it doesn't work, even on the container-based infrastructure (see also https://github.com/travis-ci/travis-ci/issues/3785)

gridaphobe:
There are a few lingering issues, described [here](https://github.com/ucsd-progsys/liquidhaskell/blob/master/TODO.markdown#ghc-710), but everything compiles and the tests pass now, so I'm merging before bitrot sets in.

ranjitjhala:
whoa! so are we officially declaring support for ghc 7.10 then?

On Fri, May 22, 2015 at 4:01 PM, Eric Seidel <notifications@github.com>
wrote:

> There are a few lingering issues, described here
> <https://github.com/ucsd-progsys/liquidhaskell/blob/master/TODO.markdown#ghc-710>,
> but everything compiles and the tests pass now, so I'm merging before
> bitrot sets in.
>
> --
> <https://github.com/ucsd-progsys/liquidhaskell/pull/364#issuecomment-104785793>
> .
>



-- 
Ranjit.

gridaphobe:
The big issue is that the HTML output is missing some annotations (so the emacs mode probably is too). If you look at GHC's output, you can see all of the source information, so I think we're losing some SrcSpan's along our pipeline. But I'm not sure where that would happen...

ranjitjhala:
So whats the story now? shall I attempt to move to 7.10?

gridaphobe:
@ranjitjhala I haven't dug any further into the annotation issue, so it will probably break your emacs mode (not necessarily the flycheck piece, but certainly the get-refined-type-at-point piece).

ranjitjhala:
Because ? I got the sense that only infix operators were broken?

Btw it may have to do with your tossing out the custom desugarer -- that was put in precisely to add src spans everywhere.

At any rate do you have it working so I can see what it looks like without fiddling with installing 7.10 yet?

gridaphobe:
Ah crap, it turns out that `ghc -g` actually isn't providing quite enough source information.

Take, e.g. `tests/pos/Abs.absN`:
```
absN =
  \ (@ a_aEJ)
    ($dNum_aNB :: Num a_aEJ)
    ($dOrd_aNC :: Ord a_aEJ)
    (x_aDY :: a_aEJ) ->
    src<tests/pos/Abs.hs:4:1-34>
    case src<tests/pos/Abs.hs:4:13-17>
         > @ a_aEJ
           $dOrd_aNC
           (src<tests/pos/Abs.hs:4:13> x_aDY)
           (src<tests/pos/Abs.hs:4:17> fromInteger @ a_aEJ $dNum_aNB 0)
    of _ [Occ=Dead] {
      False ->
        src<tests/pos/Abs.hs:4:31-34>
        negate @ a_aEJ $dNum_aNB (src<tests/pos/Abs.hs:4:33> x_aDY);
      True -> src<tests/pos/Abs.hs:4:24> x_aDY
    }
```
The application of `x > 0` has a tick, but `>` itself does not. Since `>` is applied infix, we don't know it's exact source location. I'll file a bug with GHC and see if we can get that fixed.

gridaphobe:
@ranjitjhala yes, the breakage is entirely due to killing our custom desugarer. I'll try to add back our desugarer in the meantime, hopefully 7.10 didn't break it too badly..

> At any rate do you have it working so I can see what it looks like without fiddling with installing 7.10 yet?

I have it running on my laptop (I switched to 7.10 a while ago)

ranjitjhala:
Couldn't quite follow this; so the problem is indeed with the ghc desugarer?

gridaphobe:
Yes, unfortunately so.

Issue #292, opened 2015-01-01
LH accepts function with malformed spec


ranjitjhala:
See https://github.com/ucsd-progsys/liquidhaskell/blob/master/tests/todo/badspec.hs
See https://github.com/ucsd-progsys/liquidhaskell/blob/master/tests/todo/badspec1.hs

This should be rejected...

ranjitjhala:
Another variant. The actual bugs are different of course, in the latter case I think we probably just take a prefix of the args and drop the rest which is why `badspec1.hs` is safe. 

Silly but very severe bug, we shouldn't be **accepting** programs with malformed specs!

nikivazou:
I think @spinda (Michael) has a fix for the second variant (not merged yet.)

spinda:
I believe the cause for the second variant is the same as for #287.

spinda:
~~I believe the cause for the second variant is the same as for #287.~~

Looking at it again, that looks like a totally different issue. The problem is that it's accepting `Foo 1000` instead of `Fooz 1000`, right?

ranjitjhala:
Yes the 1000 is somehow ignored in the application...

spinda:
Just as an update, I'm working on fixing this right now.

ranjitjhala:
Folks, what is going on with this? 

We need to fix it ASAP, because its a rather horrible and urgent bug. 

I have added another version, this time in `tests/neg` to remind ourselves 
that the build is broken until this is fixed.

spinda:
Ack, ack, ack. Right, this is the plugHoles issue.

ranjitjhala:
Is it the case that plug holes is going to take a while to fix?

spinda:
No, I believe it can be fixed on its own. I'll work on it asap if no one
else has gotten to it before then.

spinda:
Working on this right now. (Well, first I'm waiting for my system to update so that I can compile the latest `fixpoint`...)

spinda:
For `badspec.hs`, `plugHoles` turns

```haskell
module Goo where

{-@ foo :: xs:_ -> {v:_ | this = rubbish } @-}
foo _ _ = 0
```

into

```haskell
module Goo where

{-@ foo :: Num a => xs:t -> {v:(t -> a) | this = rubbish } @-}
foo :: Num a => t -> t -> a
foo _ _ = 0
```

and then LiquidHaskell seems to ignore refts on `(->)`, as that passes on its own.

[...]

PR #368, opened 2015-04-01
Error on type synonyms applied to too many args

spinda:
Addresses tests/todo/badspec{1,2}.hs from #292 (minus the binder-as-tyvar issue with badspec2)


nikivazou:
Thanks!

PR #378, opened 2015-04-09
Dump logs of failed tests when running on Travis

spinda:
@nikivazou and I were discussing this earlier today. When a test fails on Travis, right now, we have to re-run it on our own systems to see log output. Besides being inconvenient, sometimes we have test failures on Travis that we can't reproduce ourselves. This PR modifies the test and Travis scripts to dump the logs of failed tests when running on Travis. It also tightens a few things up in the Travis script.

You can view an example here:

https://travis-ci.org/spinda/liquidhaskell/jobs/57744542#L832-L836

with this at the end:

```
(...)
The command "scripts/travis do_test "$TESTS" "$SMT" && scripts/travis test_source_pkg" exited with 1.
$ scripts/travis dump_fail_logs
./benchmarks/bytestring-0.9.2.1/Data/ByteString/Lazy/.liquid/Char8.hs.log.fail:
    http://sprunge.us/FQeJ
./benchmarks/bytestring-0.9.2.1/Data/ByteString/.liquid/Char8.hs.log.fail:
    http://sprunge.us/bhQT
```

I originally had the Travis script dump the logs to stdout, but their length caused Travis to break in strange ways. So I have the script upload them to the http://sprunge.us pastebin instead, printing out links to where the logs can be viewed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,7 @@ install:
 
 script:
  - scripts/travis do_build
- - scripts/travis do_test "$TESTS" "$SMT"
- - scripts/travis test_source_pkg
+ - scripts/travis do_test "$TESTS" "$SMT" && scripts/travis test_source_pkg
 
 after_failure:
  - scripts/travis dump_fail_logs
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ install:
 script:
  - scripts/travis do_build
  - scripts/travis do_test "$TESTS" "$SMT"
- - scripts/travis dump_fail_logs
  - scripts/travis test_source_pkg
+
+after_failure:
+ - scripts/travis dump_fail_logs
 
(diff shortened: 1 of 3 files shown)


spinda:
Note that the Travis tests are *supposed* to fail here. `bytestring` is still broken on `master`; this shouldn't prevent the PR from being merged.

nikivazou:
Thanks @spinda !

nikivazou:
The log for `WhieM.hs` is not super useful: http://sprunge.us/BWGj

It seems that it just time-outs?

spinda:
@nikivazou The `Killed` at the end indicates that something else is killing the process before it completes. Not sure what would be doing that...

spinda:
Exit code 137 means that the process received SIGKILL. The test runner
isn't getting killed, so it can't be the Travis script, and there's nothing
that would kill the liquid process in the test runner.

Only thing I can come up with is that we're running out of memory and
Linux's OOM killer is killing liquid.

PR #380, opened 2015-04-14
Multi parameter measures

nikivazou:
A first example with two parameter measures!
See [tests/pos/HasElem.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/multiparams/tests/pos/HasElem.hs) and  [tests/neg/HasElem.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/multiparams/tests/neg/HasElem.hs).

Also, this branch stops support for `exist` in types, as it can always be expressed with bounded refinement types.


ranjitjhala:
Nice!! Before we merge can you check whether this has any impact on the performance numbers? Just compare the times for master vs this branch for a sample of benchmarks?

nikivazou:
Good idea. I will do it. Thought given that that imported build-in types do not have multi-parameter measures, I do not expect the time to increase. But verification is expected to be slow for programs that actually use multi-parameter measures. 

@spinda, @gridaphobe  is there any way to tell Travis to log time?

ranjitjhala:
Actually any way to use tasty as well? Perhaps there is some infrastructure for measuring the performance of tests?

spinda:
Isn't time already being logged? See: https://travis-ci.org/ucsd-progsys/liquidhaskell/jobs/58532879

```
​Tests
  Unit
    pos
      AVL.hs:                     OK (5.21s)
```

`AVL.hs` took 5.21 seconds.

ranjitjhala:
Good point. Is there a "built-in" way to persist that? And compare over
runs etc.? Or do folks just hand roll it in an ad-hoc fashion?

On Thu, Apr 16, 2015 at 7:18 AM, Michael Smith <notifications@github.com>
wrote:

> Isn't time already being logged? See:
> https://travis-ci.org/ucsd-progsys/liquidhaskell/jobs/58532879
>
> Tests
>   Unit
>     pos
>       AVL.hs:                     OK (5.21s)
>
> AVL.hs took 5.21 seconds.
>
> --
> <https://github.com/ucsd-progsys/liquidhaskell/pull/380#issuecomment-93615265>
> .
>



-- 
Ranjit.

spinda:
You can upload data to various locations after a run:

http://docs.travis-ci.com/user/deployment/

The logs can be pulled from Travis as well. I'm not aware of a ready-made way of aggregating and comparing this information, though. @gridaphobe?

spinda:
I'm not sure how reliable the Travis timing numbers would be, given the variance in hardware that build VMs might be allocated to, varying load from other VMs on the same hardware, and so on. Might it be better to get these timing values by running the tests on a regular machine?

gridaphobe:
1. I sent a patch to tasty a while ago to log timing data. There's no built-in support for persisting the numbers in a machine-readable format AFAIK, but I can easily throw together a custom reporter that will save a json or csv file.
2. I've used travis before to compare performance between different versions of Target. As @spinda points out, you have to expect some variance in the timings due to hardware/load changes, but in my experience it's consistent enough to detect big regressions.
3. I don't know of an existing service for logging/analyzing performance data of the top of my head, but I'm sure we can find one :)

PR #382, opened 2015-04-19
Use Travis containers for stabler and 2x faster builds

spinda:
Travis has two build systems: a VM-based system and a [Linux containers-based system](http://docs.travis-ci.com/user/workers/container-based-infrastructure/). The latter can run your builds if you don't need `sudo` support. It's much faster to start up and run builds, it has more memory available, it's stabler, and it allows the caching of files between builds. This is the biggest win for us, as caching our Cabal dependencies instead of recompiling them every time cuts total build time in half ([before](https://travis-ci.org/ucsd-progsys/liquidhaskell/builds/57073862) / [after](https://travis-ci.org/spinda/liquidhaskell/builds/59090800)).

Travis can install `apt` packages before removing privileges from your container, as long as they're on a [whitelist](https://github.com/travis-ci/apt-package-whitelist), for security reasons. [I got `ocaml` and `camlidl` whitelisted](https://github.com/travis-ci/travis-ci/issues/3639), and since we're now using Travis's built-in GHC support instead of installing our own copy, we don't need `sudo` anymore.

Also, I noticed that the [`vector-algorithms` benchmarks haven't actually been running](https://travis-ci.org/ucsd-progsys/liquidhaskell/jobs/57073869#L808): the Travis script tried to run `Benchmarks/vector-algorithms` instead of `Benchmarks/vect-algs`. I went and fixed that here. Fortunately, they're still passing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ env:
  - SMT=z3 TESTS=Benchmarks/text
  - SMT=z3 TESTS=Benchmarks/bytestring
  - SMT=z3 TESTS=Benchmarks/esop
- - SMT=z3 TESTS=Benchmarks/vector-algorithms
+ - SMT=z3 TESTS=Benchmarks/vect-algs
  - SMT=cvc4 TESTS=Unit/pos
  - SMT=cvc4 TESTS=Unit/neg
  - SMT=cvc4 TESTS=Unit/crash
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,17 +33,17 @@ env:
  # - TESTS=hscolour
 
 before_install:
- - scripts/travis setup_cabal
+ - scripts/travis clean_cache "$SMT"
  - scripts/travis clone_fixpoint
- - scripts/travis install_smt "$SMT"
 
 install:
  - scripts/travis install_cabal_deps
+ - scripts/travis install_smt "$SMT"
 
 script:
  - scripts/travis do_build
  - scripts/travis do_test "$TESTS" "$SMT" && scripts/travis test_source_pkg
- - scripts/travis do_cleanup
+ - scripts/travis clean_cache "$SMT"
 
 after_failure:
  - scripts/travis dump_fail_logs
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,8 +41,7 @@ install:
  - scripts/travis install_smt "$SMT"
 
 script:
- - scripts/travis do_build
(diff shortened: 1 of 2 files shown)


spinda:
Unforunately, this doesn't fix the issue we've been having with `WhileM` getting killed on Travis.

.travis.yml
```diff
@@ -1,5 +1,21 @@
...
+  apt:
+    packages:
+    - ocaml
+    - camlidl
+
+sudo: false
+
+cache:
+  apt: true
+  directories:
```
gridaphobe: I've been reluctant to enable caching for cabal packages because then we have to worry about dependency conflicts as time moves on.

At the very least we need to add `--upgrade-dependencies` to the `cabal install` invocations to ensure we're always testing against new versions of packages. Is there another flag we can pass to ensure a more sandbox-like environment?

(I've been considering setting up the nix package manager as a first step on travis to improve build times, since it has binary caches for haskell packages.)

spinda: I know [Halcyon](https://halcyon.sh/) handles this well, but I don't know about Cabal. It's hacky, but the Travis script could throw away the cache and reinstall if Cabal says there will be a conflict.

gridaphobe: Wiping the cache when there's a conflict sounds fine to me. It's still a net win for us and travis :)

spinda:
This should be ready to merge now, as long as everyone's okay with it.

ranjitjhala:
lets do it!

@gridaphobe are you back in SD?

On Mon, Apr 20, 2015 at 10:28 PM, Michael Smith <notifications@github.com>
wrote:

> This should be ready to merge now, as long as everyone's okay with it.
>
> --
> <https://github.com/ucsd-progsys/liquidhaskell/pull/382#issuecomment-94642438>
> .
>



-- 
Ranjit.

gridaphobe:
@ranjitjhala not yet, I'll be back on Monday.

spinda:
I need to debug another Cabal issue, so don't merge this yet after all.

gridaphobe:
@spinda have you tried using `--reorder-goals` instead of `--max-backjumps`? I always forget what it does, but it often helps when I run into dependency conflicts..

spinda:
@gridaphobe I'm doing a few runs with `--reorder-goals` now. Hopefully it'll fix this up.

spinda:
That seems to have fixed it. Should be okay to merge now.

PR #383, opened 2015-04-22
Improvements to `--diff` mode

ranjitjhala:
This causes many changes rippling across liquid-fixpoint (see the `diff` branch there).

@spinda -- should we merge this into `master` **now** rather than later? I fear this will cause you major headaches otherwise?

@nikivazou and @gridaphobe  please advise.

related to: https://github.com/ucsd-progsys/liquid-fixpoint/pull/60

### DIFF
--- a/Liquid.hs
+++ b/Liquid.hs
@@ -54,9 +54,10 @@ liquidOne target info =
                     putStrLn "*************** Slicing Out Unchanged CoreBinds *****************"
      dc <- prune cfg cbs' target info
      let cbs'' = maybe cbs' DC.newBinds dc
-     let cgi   = {-# SCC "generateConstraints" #-} generateConstraints $! info {cbs = cbs''}
+     let info' = maybe info (\z -> info {spec = DC.newSpec z}) dc
+     let cgi   = {-# SCC "generateConstraints" #-} generateConstraints $! info' {cbs = cbs''}
      cgi `deepseq` donePhase Loud "generateConstraints"
-     out      <- solveCs cfg target cgi info dc
+     out      <- solveCs cfg target cgi info' dc
      donePhase Loud "solve"
      let out'  = mconcat [maybe mempty DC.oldOutput dc, out]
      DC.saveResult target out'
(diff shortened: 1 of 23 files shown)


nikivazou:
We should merge as soon as it passes Travis (except `whileM`).

spinda:
That's a lot of changes! We should get them merged sooner rather than later.

ranjitjhala:
Grr this is that odd thing where we need to use a special branch of fixpoint to even compile... 

What's the protocol?

gridaphobe:
Unfortunately there's not a great solution at the moment, beyond testing locally (or on goto). The best we can do at the moment is merge the fixpoint branch and restart travis :/

ranjitjhala:
connected to https://github.com/ucsd-progsys/liquid-fixpoint/pull/60

spinda:
@gridaphobe Couldn't we add `-b diff` to the fixpoint clone command in the Travis script?

ranjitjhala:
Folks (@spinda and @gridaphobe) what is one to make of the these travis reports? Should I merge this branch or not? Also, can we fix this ASAP? Not much point in using Travis if we have to read the tea leaves to figure out if its actually broken or not ... :-/ 

`The command "cabal update" failed and exited with 127 during .`
https://travis-ci.org/ucsd-progsys/liquidhaskell/jobs/59820728
https://travis-ci.org/ucsd-progsys/liquidhaskell/jobs/59820722

`whileM` fails
https://travis-ci.org/ucsd-progsys/liquidhaskell/jobs/59820730

gridaphobe:
That failure appears to a screw-up by travis' provisioning system as it complains that it can't find cabal. Not much we can do about that...

@spinda, I was hoping to find a more systematic way of determining which branch to fetch, not just hard coding it. The problem is that Travis fetches the current PR by number, not by branch-name, so we can't just assert that the branch names must match. I haven't investigated too deeply though, so there may be some bit of git hackery we can apply :smile:

ranjitjhala:
Ok, I'm merging this in.

Issue #395, opened 2015-05-19
URGENT: Incomprehensible error / measure bug


ranjitjhala:
@nikivazou can you look into this ASAP, seems to be some strange error with the lifting code.

See `tests/todo/fixme.hs` -- in brief:

```
{-@ measure keys @-}
keys :: (Ord k) => [(k, v)] -> S.Set k
keys []       = S.empty
keys (kv:kvs) = (S.singleton (fst kv)) `S.union` (keys kvs)
```

but this is fine.

```
{-@ measure okeys  :: [(a, b)] -> (S.Set a)
    okeys ([])     = (Set_empty 0)
    okeys (kv:kvs) = (Set_cup (Set_sng (fst kv)) (okeys kvs))
  @-}
```

nikivazou:
@ranjitjhala was this example working in  the past?

ranjitjhala:
Don't know -- just wrote it -- but I have seen this same crash before.

ranjitjhala:
Btw even if you can't fix it now, it would be good to get a decent error message pointing to some source location that causes the problem.

ranjitjhala:
ok, less desperate now because i could work around it by using

   data Map k v = Emp | Bind k v (Map k v)

but still an error message would be nice prior to the update...

nikivazou:
I pull requested a fix, but lets wait for travis.

nikivazou:
It is fixed. The test is moved here

https://github.com/ucsd-progsys/liquidhaskell/blob/master/tests/pos/Keys.hs

@ranjitjhala  note how I redefined `myfst` (instead of using `fst`) because `fst` has a true type.

ranjitjhala:
Thanks @nikivazou !!!

On Tue, May 19, 2015 at 6:55 PM, Niki Vazou <notifications@github.com>
wrote:

> Closed #395 <https://github.com/ucsd-progsys/liquidhaskell/issues/395>.
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/ucsd-progsys/liquidhaskell/issues/395#event-309086412>
> .
>



-- 
Ranjit.

PR #396, opened 2015-05-19
fir for #395

nikivazou:


ranjitjhala:
Thanks @nikivazou ... lets wait for travis.

Issue #273, opened 2014-12-07
Liquid Haskell doesn't accept Haskell names containing ' (single-quote)


david-christiansen:
In the Idris code base, we have a datatype `Err' :: * -> *` where `Err' t` is a type of error messages containing terms of type `t`. Liquid Haskell couldn't parse the specification:

```
{-@ data Err' [errSize] @-}
```

because of this. @gridaphobe and I worked around the issue by manually applying it at use sites. Additionally, the same thing happened when using the name `n'` in the name of a Liquid Haskell binder.

nikivazou:
Assigned to @Kyly

nikivazou:
Assigned to @Kyly

nikivazou:
Fixed at #397

PR #397, opened 2015-05-25
Accept Haskell names containing ' 

Kyly:
Now accepts data type names with single-quote. #273


src/Language/Haskell/Liquid/Parse.hs
```diff
@@ -1,5 +1,4 @@
 {-# LANGUAGE NoMonomorphismRestriction #-}
-{-# LANGUAGE FlexibleContexts          #-}
```
gridaphobe: This extension is required by GHC 7.10, please add it back.

src/Language/Haskell/Liquid/Parse.hs
```diff
@@ -750,13 +749,18 @@ tyBodyP ty
           outTy (RFun _ _ t _) = Just t
           outTy _              = Nothing
 
+upperIdP' :: Parser Symbol
+upperIdP' = try $ symbol <$> condIdP alphanums (isUpper . head)
+  where
+    alphanums = ['A' .. 'Z'] ++ ['a'..'z'] ++ ['0'..'9'] ++ "'"
```
gridaphobe: Won't this allow parsing things like `'Foo`, which is not a valid Haskell name?

Also, we should also accept `'`s in identifiers, not just constructors, so it would probably be best to have one parser that accepts *any* valid Haskell name, and a variation that checks that the first character is upper-case.

spinda: `isUpper` would prevent that, but the point about `'` in identifiers is still true.

spinda:
Could you rebase those merge commits out, or at least squash some of them together?

src/Language/Haskell/Liquid/Parse.hs
nikivazou: We should also allow unicodes and generally follow haskell's grammar https://www.haskell.org/onlinereport/syntax-iso.html

But lets merge this (once rebased and Travis accepts it) and @Kyly will fix it later.

PR #398, opened 2015-05-26
Data types now take unicode characters and single quotes.

Kyly:
This allows for Unicode character and single quotes in datatype name. For example `data Λλ'` is excepted but `data λ'` would not. Should be easy to apply this same definition to variable ids. I'll get on started that once I have a chance to meet with @nikivazou. #273


ranjitjhala:
Very nice!! Thanks @kyly!!

PR #402, opened 2015-06-09
Autosize

nikivazou:
Automatically infer sizing (termination) info for ADT: the user does not have to write the sizing functions.


ranjitjhala:
Nice! Can you update the readme and change log?

ranjitjhala:
Why is the build broken?

gridaphobe:
Unused variables in `Language.Haskell.Liquid.Constraint.Generate` (https://travis-ci.org/ucsd-progsys/liquidhaskell/jobs/66142967#L1120-L1127)

ranjitjhala:
I don't understand, how does it build locally then?

On Tue, Jun 9, 2015 at 7:53 PM, Eric Seidel <notifications@github.com>
wrote:

> Unused variables in Language.Haskell.Liquid.Constraint.Generate (
> https://travis-ci.org/ucsd-progsys/liquidhaskell/jobs/66142967#L1120-L1127
> )
>
> —
> <https://github.com/ucsd-progsys/liquidhaskell/pull/402#issuecomment-110568506>
> .
>



-- 
Ranjit.

gridaphobe:
You're probably not using `-Werror` locally, so the unused variables would just present as a warning.

ranjitjhala:
I'm just hitting "make" (and presumably so is Niki), and I think `-Werror`
happens by default in that case (its certainly happening to me, its caught
a bunch of little things which have made it worth the irritation)

(i.e. how would one even *not* use it locally given that its in the .cabal
file by default?)

​

gridaphobe:
`-Werror` is excluded by default (https://github.com/ucsd-progsys/liquidhaskell/blob/master/liquidhaskell.cabal#L45-L48)

That's probably my fault as I have to disable it when I upload to hackage, and sometime accidentally commit the result..

ranjitjhala:
! then how come I get it whenever I hit `make` ?
​

ranjitjhala:
ah you can circumvent it with 'cabal install' instead of 'make'. tsk. tsk.
​

ranjitjhala:
anyways, @nikivazou please fix the errors, the unused var thing is super
annoying, but also super useful (and on the whole positive IMO.)

​

gridaphobe:
Also of note, `make` defaults to the first target which is

```
fast:
	$(CABAL) install -fdevel --ghc-options=$(FASTOPTS)
```

the `-fdevel` turns on `-Werror` regardless of defaults.

nikivazou:
So, I am doing `cabal install` and I had this issue before (building locally and then failing on travis). I remember I turned on `-Werror` then. Lets keep the flag on always.

PR #404, opened 2015-06-10
Paths

ranjitjhala:
Added `--cabaldir` option to let LH use the `.cabal` file and sandbox information for a target to suitably beef up the search paths and GHC options. This lets us actually run LH conveniently (if slowly!) on itself, and presumably other non-trivial projects. However, looks like the GHC -> Core times are quite bothersome.


src/Language/Haskell/Liquid/CmdLine.hs
```diff
@@ -242,12 +256,41 @@ parsePragma   :: Located String -> IO Config
...
 
 ---------------------------------------------------------------------------------------
+withCabal :: Config -> IO Config
+---------------------------------------------------------------------------------------
+withCabal cfg
+  | cabalDir cfg = withCabal' cfg
+  | otherwise    = return cfg
+
+withCabal' cfg = do
+  putStrLn $ "addCabalDirs: " ++ tgt
```
gridaphobe: We should try to respect the verbosity settings, using `whenNormal` (or perhaps in this case `whenVerbose`) to guard `print` expressions.

src/Language/Haskell/Liquid/CmdLine.hs
```diff
@@ -242,12 +256,41 @@ parsePragma   :: Located String -> IO Config
...
+  io <- cabalInfo tgt
+  case io of
+    Just i  -> return $ fixCabalDirs' cfg i
+    Nothing -> do putStrLn "Cannot find .cabal information, proceeding without."
```
gridaphobe: Ditto here

liquidhaskell.cabal
```diff
@@ -271,4 +274,6 @@ test-suite test
                      tasty >= 0.10,
                      tasty-hunit >= 0.9,
                      tasty-rerun >= 1.1,
-                     transformers
+                     transformers,
+                     Cabal >= 1.18
```
gridaphobe: We should only need `Cabal` as a dependency in the `library` section since we don't import any of its modules in `Liquid.hs` or `test.hs`.

ranjitjhala: We _do_ need it to build Liquid.hs --> CmdLine.hs --> Cabal.hs

That is, we need it in the executable too.

I'll let you fiddle with the Cabal file, I can't get it to build without
putting the damn thing everywhere :)

have fixed the putStrLn issues...
​

gridaphobe: Yes, but the executable is a separate build target from the library, so it doesn't directly depend on cabal. Because of this separation, the executable target should only depend on packages that it imports **directly**, but I'll take a look :)

Issue #399, opened 2015-05-26
Paring Refined Data Definitions


nikivazou:
See 

[tests/todo/Parse.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/406e046c88eeb4edb5802017394154e4f97890ac/tests/todo/Parse.hs)

nikivazou:
Assigned to @Kyly

gridaphobe:
@nikivazou can this be closed?

nikivazou:
Sure!

PR #405, opened 2015-06-16
Changed reading space to reading blanks in dataDeclFullP

Kyly:
Just a small change but it seems to fix the issue of paring refined data definitions. #399

### DIFF
--- a/src/Language/Haskell/Liquid/Parse.hs
+++ b/src/Language/Haskell/Liquid/Parse.hs
@@ -864,7 +864,7 @@ dataDeclFullP
        spaces
        fsize <- dataSizeP
        spaces
-       ts  <- sepBy tyVarIdP spaces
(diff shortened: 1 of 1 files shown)


ranjitjhala:
I wow that's it? Did it break any tests?
Thanks!

Kyly:
Well it seems to have broken a few things. Not sure what the issue is. I can't seems to reproduce the problem locally.

ranjitjhala:
Unfortunately there are some broken things on TRAVIS because of various
reasons (some tests need to be updated to ghc 7.10).

Is anything NEW broken by this change? (The parser, unfortunately, is a
real house of cards...)
​

nikivazou:
There are some failing tests with ghc 7.8 but they have exit code 137. I think this means something went wrong on the cloud, so I rebuild.

nikivazou:
Thanks!

Issue #407, opened 2015-06-18
Look into possibly supporting Stack


spinda:
Brief: https://www.reddit.com/r/haskell/comments/3a6p9h/how_stack_lowers_the_barrier_to_open_source/

[Stack](https://github.com/commercialhaskell/stack#readme) is a new build tool from the commercial-haskell project that might be worth looking into. It handles build caching, dependency management, multi-package projects (useful for our case), and so on, without the fuss of Cabal sandboxes, linking, and recompiles. Despite the name, it's not locked to Stackage, and can pull packages from both Hackage and arbitrary git commits.

ranjitjhala:
Yes, I was also pondering this (having spent the better part of the last
few days
burning cycles recompiling sandboxes a billion times...)
​

gridaphobe:
Hehe, I (mostly) haven't had to deal with cabal nonsense since I switched to using the nix package manager :)

Still, `stack` seems like a useful tool, and it looks like you just need to add a `stack.yaml` file with some configuration options. But I think if we're actually going to add a `stack.yaml` file to the repo, one of us should actually be using it. Otherwise it might bitrot..

spinda:
I can try it out. I gave up on sandboxes and have just been cabal
install-ing everything, so stack would been an improvement.

spinda:
I have a test of one method of setting up a workspace, in one repository with the other repositories as submodules:

https://github.com/spinda/liquid

After installing stack, running

```bash
git clone git@github.com:spinda/liquid.git
cd liquid
git submodule init
git submodule update
```

to set up the repository and

```bash
stack build
```

to build `liquid-fixpoint` and `liquidhaskell` worked perfectly. Then

```bash
stack exec -- liquid tests/pos/Abs.hs
stack exec -- liquid tests/neg/alias00.hs
```

to verify that running `liquid` worked.

`stack test` built everything successfully but failed in the test runner, which assumes the `liquid` binary will be located at `dist/build/liquid/liquid`.

Overall, a very nice first impression.

spinda:
The "megarepository" model isn't required for using Stack, but putting a `stack.yaml` file in each repository doesn't work as nicely for development right now due to https://github.com/commercialhaskell/stack/issues/348.

ranjitjhala:
In perfectly happy to use it, once we can set it up...

gridaphobe:
@spinda where does stack put the `liquid` binary?

spinda:
[...]

PR #408, opened 2015-06-20
Determine liquid path relative to test runner

spinda:
re: #407

This should work for both `cabal test` and `stack test`.


spinda:
If everything looks okay, I'd like to merge this in quickly so I can use `stack test` instead of running my tests manually.

ranjitjhala:
Assuming that plain old `make test` also works?

spinda:
Yes, `make test` works.

PR #411, opened 2015-06-30
Use createDirectoryIfMissing for .liquid dirs

spinda:
This squelches the

    Warning: Couldn't do create temp directory: .liquid: createDirectory: already exists (File exists)

message that shows up sometimes.


ranjitjhala:
thanks!

Issue #412, opened 2015-07-02
Another broken test (GHC 7.10.1)


m0davis:
I'm new to LH. Is this a known issue? Given the following input file, TestLiquid.hs:

> module TestLiquid where
> data Free f a = Pure a | Free (f (Free f a)) 
> oneForFree :: Free [] Int
> oneForFree = Free [Pure 1] 

I get the following error when trying to `liquid TestLiquid.hs`:

LiquidHaskell Copyright 2009-15 Regents of the University of California. All Rights Reserved.

**** DONE:  Extracted Core using GHC *******************************************
 
liquid: 
**** ERROR *********************************************************************
**** ERROR *********************************************************************
**** ERROR *********************************************************************
(Another Broken Test!!!) splitc unexpected:
{VV#259 : a^"lq_tmp_x_261" | $k__260[lq_tmp_x_258:=lq_tmp_x_317][VV#268:=VV#348][lq_tmp_x_256:=lq_anf__dwY][lq_tmp_x_257:=lq_anf__dwZ][lq_tmp_x_271:=lq_tmp_x_317][lq_tmp_x_274:=VV#348][lq_tmp_x_253:=VV#345][lq_tmp_x_332:=VV#345][lq_tmp_x_326:=VV#348][lq_tmp_x_323:=lq_tmp_x_317]}

{lq_tmp_x_305 : (TestLiquid.Free [] GHC.Types.Int)^"lq_tmp_x_24" "lq_tmp_x_313" | $k__312[lq_tmp_x_310:=lq_tmp_x_317][VV#311:=lq_tmp_x_305][lq_tmp_x_318:=VV#345]}
**** ERROR *********************************************************************
**** ERROR *********************************************************************
**** ERROR *********************************************************************

ranjitjhala:
Oops, no this seems to be something new thanks for pointing it out! We will try to get to it as soon as possible!

nikivazou:
Thanks for the report @m0davis! I am taking a look at it right now.

nikivazou:
Fixed! Thank for the report @m0davis, and let us know for anything else you want!

ranjitjhala:
thanks @nikivazou!!

On Fri, Jul 3, 2015 at 5:03 PM, Niki Vazou <notifications@github.com> wrote:

> Closed #412 <https://github.com/ucsd-progsys/liquidhaskell/issues/412>.
>
> —
> Reply to this email directly or view it on GitHub
[...]

PR #413, opened 2015-07-02
Non fully applied TyCons 

nikivazou:
should not create constraints on their abstract refinements

fix for #412


ranjitjhala:
Do all local tests pass?

nikivazou:
Now all tests pass locally, lets see if Travis agrees......

ranjitjhala:
Looks like. 137 failure so merging thanks @nikivazou

PR #416, opened 2015-07-12
fix link to Z3 in README.md

michaelficarra:


ranjitjhala:
Thanks @michaelficarra !

Issue #417, opened 2015-07-14
.dir-locals.el should be added to .gitignore and removed from repo


christetreault:
There is a `.dir-locals.el`file in the project root. This file is evaluated by emacs every time a file is visited in the directory, and gives me the following output:

```
The local variables list in [prefix]/liquidhaskell/ contains values that may not be safe (*).

... stuff about prompt options ...

* eval : (setq default-directory (locate-dominating-file buffer-file-name ".dir-locals.el"))
```

Maybe this is useful to some people, but others may not like this file changing their `default-directory`. Since this is a personal choice type thing, please add this file to the .gitignore and remove it from the repository.

ranjitjhala:
Can you go ahead and do this? (Submit a pr) thanks!

christetreault:
Done

nikivazou:
Thanks!

gridaphobe:
Ack, sorry I missed this issue, but there was a reason for adding the `.dir-locals.el` to the repo. Setting the `default-directory` is the only way I know of to make flycheck run `hlint` from the project root, in order to pick up the settings in `HLint.hs`.

There are other things I would like to put in `.dir-locals.el`, like disabling tabs and trimming whitespace. These things are best specified per-project IMO as they can make for noisy commits in other projects.

christetreault:
I guess the real question is: How critical is it that all emacs users get the specialized HLint settings?

If the special rules are important (organization coding standards, performance, etc...) maybe the build system should require that HLint pass before it will build?

spinda:
If we really wanted, we could add it to Travis's tests...

gridaphobe:
Well, ideally a vim user would provide a similar set of project settings, etc. :)

Adding HLint to travis at the moment would be a bad idea because we certainly don't follow most of the suggestions, but maybe in the future...

PR #418, opened 2015-07-14
Delete dir locals

christetreault:
Delete .dir-locals.el and add to .gitignore. Closes issue #417


nikivazou:
Thanks @christetreault!

Issue #431, opened 2015-08-24
Change how Configs are merged


christetreault:
Currently, the Liquid Haskell `Config` type has a `Monoid` instance and this is used to merge configurations (environment variable + command linea args = `Config`). 

After [some discussion](https://github.com/ucsd-progsys/liquid-fixpoint/issues/90#issuecomment-133581372) it seems that this might not be the best way. @gridaphobe suggested a new function `mergeConfig :: Config -> Config -> Config`, not bound by the `Monoid` laws that can be used to merge the configs an a sane manner. This function would merge the configs such that higher priority items override lower priority with `priority = env < file < cmdline`.

gridaphobe:
That's not quite what I meant, so let me elaborate. My suggestion is that instead of having the config-parser return a full `Config` it should return a function from `Config -> Config`, e.g.

```haskell
parseFlag :: String -> (Config -> Config)
-- parseFlag "--no-termination" = \cfg -> cfg { noTermination = True }
-- parseFlag "--idirs src" = \cfg -> cfg { idirs = "src" : idirs cfg }
-- etc.
```

The comments are examples of how this function should behave, not a suggested implementation :) Then we could run the parser on the three different sources of flags and combine the resulting functions (incidentally, these `Config -> Config` functions *do* form a `Monoid`, where `mappend = (.)` and `mempty = id`), giving us a single `Config -> Config` function that we can apply to the default `Config`. So the top-level parser would look something like

```haskell
getOpts :: FilePath -> IO Config
getOpts file = do
  envT <- parseFlags <$> lookupEnv "LIQUIDHASKELL_OPTS"
  fileT <- parseFlags <$> readPragmas file
  argsT <- parseFlags <$> getArgs
  return (argsT . fileT . envT $ defaultConfig)
```

I don't know how difficult it would be to rewrite our config-parser in this manner, it probably depends on the API the underlying parser library provides. Currently we use http://hackage.haskell.org/package/cmdargs.

ranjitjhala:
This seems too complicated.
Can someone remind me of what the problem is with chris' monoid suggestion?

christetreault:
My proposal is to get rid of the Monoid instance and just have a function that combines Configs where one takes precedence. This way we can not have to worry about being good people and [obey the law](https://hackage.haskell.org/package/base-4.8.1.0/docs/Data-Monoid.html)

That said, now that I read the monoid laws, I guess there's no real reason we can't obey them and still do what I suggest...

[...]

PR #436, opened 2015-08-26
Merge configs such that File > CmdLine > Env

christetreault:
Change how configs are merged. Priority is _File > Command Line > Environment_ per today's discussion with @ranjitjhala and @gridaphobe.

Will close issue #431


christetreault:
:+1:

PR #437, opened 2015-08-27
Sorts

ranjitjhala:
Various changes to the `Sort` data type and related operations needed to untangle the mess whose symptoms are in #75 (in liquid-fixpoint).

local tests pass, both for liquid-fixpoint and liquidhaskell, but would be good to run on travis.


ranjitjhala:
I just eagerly merged on the command line to get travis to get this running.

Issue #443, opened 2015-09-05
Function call to non-constant


bmcfluff:
See:

https://github.com/ucsd-progsys/liquidhaskell/blob/master/tests/pos/ResolvePred.hs

In the fq file, the function ResolvePred.C#rpw is not a Constant but it is applied in refinements.

ranjitjhala:
Should be fixed by:

https://github.com/ucsd-progsys/liquidhaskell/commit/9dcaaae944611145898cf39dc89a6b5e340d69b1#diff-6409f1f998798a468b29e8bb8e2a82e3R203

in particular, [this PR](https://github.com/ucsd-progsys/liquidhaskell/pull/444).

bmcfluff:
The fix worked; closing. (ResolvePred still fails but for an unrelated reason now.)

PR #444, opened 2015-09-05
Add First-Order Data Constructors as Fixpoint Constants

ranjitjhala:
Earlier, given things like:

```
data L a = N | C a (L a)
```

were adding just things like `N` but not the `C` -- deferring the latter to `binds`. In general *all* symbols that appear as applications inside refinements *must* be declared as constants. See #443. 

This particular PR should fix that, by adding all first-order data-constructors.


ranjitjhala:
Since @christetreault has confirmed lack of performance degradation, am merging.

PR #451, opened 2015-09-11
Check Well Sortedness of Qualifiers

nikivazou:
See ucsd-progsys/liquid-fixpoint#98


nikivazou:
@spinda, I remember you added a link to Travis that reports why each test fails. It is still there? I cannot find it

PR #469, opened 2015-09-25
Resolve expr args to aliases (fixes #468)

### LINKED ISSUE #468: Name Resolution and Type Alias Expansion
In `tests/todo/append-bug-name.hs` the check for
`prop_app_nil_ok` works fine, but `prop_app_nil_bad` doesn't.

The only difference is the type of axiom_append_nil_{bad, ok},
in which the only difference is that one uses the {...N...} while
the other does not. For some reason this yields two **different** 
refinements, one with `Append.N` and one with `N`, which 
causes the problem. 

In short, there is some odd name resolution bug that interacts 
with the type alias expansion code. 

@spinda -- any idea whats going on?

spinda:
Expression arguments weren't being resolved before being substituted into aliases. This fixes that.

@nikivazou could you please test this?


nikivazou:
Looks goor @spinda, thanks!

ranjitjhala:
Thanks @spinda!!

PR #470, opened 2015-09-25
Add a stricter check in the type specifications

nikivazou:
checking with Travis


ranjitjhala:
@nikivazou is this for issue #468 or something else? (And either way should we wait or merge?)

nikivazou:
It is something else: function binders do not go into fixpoint environment, but when we check type signatures we do insert function binders in the check-environment. So a signature

```
foo :: f:(a -> b) -> x:a -> {v:b | v == f x }
```

will be treated as well-formed. Which is wrong because pruning of unsorted refinements will discard the unsorted `v == f x`, since `f` will be out of scope. And this can lead to unsoundness.

ranjitjhala:
Eeks. So merge?

nikivazou:
Yes! I was waiting for Travis

PR #475, opened 2015-09-30
Performance monitor

christetreault:
For my testing, I've had some kludges in place due to this groundwork not being in master. Due to this, I'm only ~99% certain it actually works right as things tend to break when you're testing with random extra lines and commented out lines. I'd like to merge this into master, and give it a few more dry runs to make sure it works right.

This merge should be perfectly safe; with the exception of one minor change to test.hs (I create the summary csv file prior to archiving the test, rather than after archiving it), all of this stuff is perfectly content to sit there and not bother anybody.


christetreault:
PS: don't delete the branch. I'm going to keep working on this branch for any additional work that needs to get done.

nikivazou:
Thanks!

PR #476, opened 2015-10-02
Use a submodule for liquid-fixpoint

spinda:
@ranjitjhala, @gridaphobe, @christetreault, and I discussed this today. This sets up `liquid-fixpoint` as a submodule of `liquidhaskell`, and updates the `stack.yaml` file to build from that local copy. (For Cabal users, the sandbox instructions on the README have been updated.)

The process of updating and keeping track of the `liquid-fixpoint` submodule has been documented in the README (under the section "Working With Submodules").

Besides forcing us to keep track of which version of `liquid-fixpoint` we're building against, this lets @christetreault's performance dashboard tie the data it collects to a specific `liquid-fixpoint` commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,8 @@ _build
 *.swp
 *.swo
 
+/.stack-work/
+/.vagrant/
 TAGS
 a.out
 *.json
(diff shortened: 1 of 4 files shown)


ranjitjhala:
travis seems super unhappy about this, should I still merge?

spinda:
Yes, this doesn't touch the Travis scripts at all. I manually cancelled the
builds because Travis was acting up again, and I wanted to free up the test
runners to check my Travis PR.

PR #477, opened 2015-10-03
Use stack and Docker to run tests on Travis

spinda:
This reworks the Travis build setup to use `stack` for installing GHC, installing dependencies,, and building `liquidhaskell`. This lets us test with multiple GHC versions without relying on either Travis's official GHC support or `multi-ghc-travis`, meaning we can switch back to Travis's Docker system for our builds. That means our builds will run much, much quicker, as (a) the Docker containers take much less time to provision than the full VMs we're currently using, and (b) with the Docker system, we can cache our GHC install and built dependencies across runs, instead of having to rebuild them every single test.

(Note&mdash;this depends on #476 being merged first.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,8 @@ _build
 *.swp
 *.swo
 
+/.stack-work/
+/.vagrant/
 TAGS
 a.out
 *.json
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "benchmarks/xmonad"]
 	path = benchmarks/xmonad
 	url = https://github.com/nikivazou/xmonad.git
+[submodule "liquid-fixpoint"]
+	path = liquid-fixpoint
+	url = https://github.com/ucsd-progsys/liquid-fixpoint.git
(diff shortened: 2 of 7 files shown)


ranjitjhala:
Awesome will merge when at my laptop!

spinda:
Sorry, please don't merge quite yet&mdash;I'm still testing this.

gridaphobe:
I don't mind switching to stack for travis, especially if it means we can use the container bots. But can we please test against hackage instead of stackage? I want travis to also serve as a guard against upgraded dependencies, and stackage prevents that.

spinda:
@gridaphobe - after some head-scratching I've managed to get this to use raw Hackage instead of Stackage.

This should be ready to merge now (or, once #476 is merged). Based on my testing, with a populated cache, this cuts total testing time roughly in half&mdash;not even counting the long boot-up times for the VM-based system, which is instantaneous for the new Docker one. With the cache, it now takes ~1-2min to set up the environment and dependencies and start building and running tests.

spinda:
Also, the tests might show up as "failed" here, but that's because there's one test (`eq_pos`) that's failing on `master` right now. Everything else seems to be passing.

gridaphobe:
Excellent, thanks!

PR #481, opened 2015-10-06
Updates to perf-mon resulting from friday meeting

christetreault:
Implements changes resulting from our Friday meeting.

1) uses the git submodule for fixpoint. Now this will correlate a specific fixpoint with a specific liquid.
2) Added readme entry for using the scripts I wrote
3) Minor best-practicy-type updates to the scripts


ranjitjhala:
Ok, one last thing -- is it possible to set up

1. a cron job or some such that just generates (extends) the graphs every week. So, every say, sunday at midnight, it creates a fresh build etc. generates the stats for ALL benchmarks and updates the database?

2. rebuild the site & post the (incrementally built up) results to some URL e.g.

   http://goto.ucsd.edu/liquid/perf-mon/

christetreault:
1) The command `generate-site.bash -s [some start hash]` should be suitable for use in a cron job. This will only generate logs for new commits prior to the last run (newer than `[some start hash]`) Obviously, if we do it on a weekly basis, this will be a lot of commits, so this will need to happen on some always-on computer.

2) There is a results folder in the site folder, which contains a bunch of .json files. This is the only thing that actually changes when you regenerate the site (as far as I can tell). To host it somewhere, it's just a matter of copying the folder to some internet-facing path. We could probably whip a script to sftp the results folder to the web server, but I imagine we don't want that one on github for security reasons. Once we get the specifics of the web server and test server, we can work out what that would look like, but I'm sure thats a 5-liner.

christetreault:
Before we deploy this to goto, we should merge this pull request. The change to use submodules is in this PR, among other things.

ranjitjhala:
go for it.

On Wednesday, October 7, 2015, Chris Tetreault <notifications@github.com>
wrote:

> Before we deploy this to goto, we should merge this pull request. The
> change to use submodules is in this PR, among other things.
>
> —
> <https://github.com/ucsd-progsys/liquidhaskell/pull/481#issuecomment-146417849>
> .
>


-- 
Ranjit.

PR #510, opened 2015-11-14
AppendArrow moved back into TODO

christetreault:
Appendarrow never terminates, moving it back into TODO


nikivazou:
Thanks @christetreault but apparently something is wrong with building fixpoint. Also I fixed `AppendArrow` in my new pull request, so no need to merge this anymore.

PR #520, opened 2015-11-24
Bump fixpoint submodule; add --real flag to tests

bmcfluff:
Because of uninterpreted * and / in fixpoint


bmcfluff:
Fixed failures and merged manually

PR #524, opened 2015-12-03
Free vars in qualifiers now an Error

bmcfluff:
This breaks a couple tests. Should I make it some kind of Warning instead, or shall we fix those tests?


bmcfluff:
(The two newly-broken tests are tests/pos/Resolve.hs and tests/pos/LocalHole.hs)

ranjitjhala:
Why are those tests breaking? are they due to recent issues? something else?

On Thu, Dec 3, 2015 at 1:22 PM, bmcfluff <notifications@github.com> wrote:

> (The two newly-broken tests are tests/pos/Resolve.hs and
> tests/pos/LocalHole.hs)
>
> —
> <https://github.com/ucsd-progsys/liquidhaskell/pull/524#issuecomment-161772045>
> .
>



-- 
Ranjit.

bmcfluff:
Those tests have free vars in some of their qualifiers. Until now this was going unnoticed (free vars don't necessarily crash fixpoint if we get lucky) but now I'm explicitly searching for them and throwing an error when I find them.

ranjitjhala:
Puzzled. I don't see any `qualif` lines in those files. Presumably the
errors are due to the **generated** qualifiers? Can you print out the:

1. offending qualifiers,
2. source positions from which they are generated (both in the information),

so we can figure out whats going on?


 ​

bmcfluff:
qualif Auto(v$35$0 : ResolveB.Bar): (v$35$0 != ResolveB.A$35$rlx) // "tests/pos/Resolve.hs" (line 8, column 1)
qualif Auto(v$35$16 : int): (v$35$16 <= n$35$a18B) // "tests/pos/LocalHole.hs" (line 10, column 5)

ranjitjhala:
The first is a globally declared constant, so should be allowed.

The second is a bug that should be fixed in LH ...

bmcfluff:
Are you sure? When I run this, I get a second qualifier `qualif Auto(v$35$0 : ResolveB.Bar): (v$35$0 = ResolveB.B$35$rly) // "tests/pos/Resolve.hs" (line 8, column 1)`; this ResolveB.B$35$rly is a globally defined constant but ResolveB.A$35$rlx from the qualifier I flagged earlier is not.

ranjitjhala:
I see, BOTH should be global constants, I suspect only one is because it is being used? 

The other should also be as it appears in the spec.

Can you file two separate LH 
issues about this?

And leave them as currently failing till they are fixed. (Do not move to todo)

bmcfluff:
Issues filed; merged.

PR #533, opened 2015-12-08
Consistency for `Integral` method refinements

Gabriel439:
The `quot` and `rem` functions have fewer refinements than `quotRem` and the `divMod` function has fewer refinements than `div`/`mod` so this change makes them behave more consistently


Gabriel439:
The context for this change is that I had some code that used `quot` and failed to type-check when I thought it should, so I checked the `include/` file for `Integral` and saw that `quotRem` had the necessary refinements that I needed.  I was able to work around this by using `fst . quotRem` instead of `quot` but I thought I should contribute the fix upstream

ranjitjhala:
Super! Thanks a bunch @Gabriel439 -- much appreciated!

Gabriel439:
You're welcome!

PR #535, opened 2015-12-11
Update to new organization of LF

ranjitjhala:
this is the matching PR for https://github.com/ucsd-progsys/liquid-fixpoint/pull/170


ranjitjhala:
waiting for tests to finish locally before merging.

PR #537, opened 2015-12-12
Lhi

ranjitjhala:
First version of Lhi seems to be working, next: fix the output so its the real thing.


ranjitjhala:
In some mysterious fashion, the relevant bits from this appear to be merged into master even though the PR is open? Oh well.

PR #538, opened 2015-12-15
Module reorganization

christetreault:
Reorganized modules per the scheme laid out in https://github.com/ucsd-progsys/liquidhaskell/blob/master/TODO.markdown#todo

- Modules moved/import lists fixed
- Some module name changes (i.e. `Language.Haskell.Liquid.GhcMisc` -> `Language.Haskell.Liquid.GHC.Misc`)
- Some minor formatting changes here and there


ranjitjhala:
Thanks! (Does this call for a version bump?)

christetreault:
Version bump not needed because we've already done a major version bump since the last deployment to hackage

Issue #547, opened 2015-12-24
Inscrutable Termination Error


ranjitjhala:
See

https://github.com/ucsd-progsys/liquidhaskell/blob/master/tests/todo/decr.hs

for details.

ranjitjhala:
This is issue remains, except that the structural checker can now verify the code on its own... 👍

PR #551, opened 2015-12-28
Split `Split` related functions into separate module

ranjitjhala:


ranjitjhala:
Lets merge ASAP after travis has done its thing.

ranjitjhala:
This is attempting to (clean up code) and fix issue #547 but just shows that more work needs to happen because the `Oblig` there is clearly bogus.

PR #556, opened 2016-01-03
Error Handling

ranjitjhala:
1. Big reshuffle of error handling modules and other things that get in the way, however, now one can import `Language.Haskell.Liquid.Types.Errors` almost anywhere (and throw errors anywhere.)

2. Added support for printing source-context with errors.


ranjitjhala:
@christetreault can you merge this in? It should be possible to go and remove all the calls to `error/star` now, ideally without `todo` or `panic` but with `panicError` (which is in `UX.Tidy` but should be importable 
in most places.)

PR #568, opened 2016-01-11
todo and impossible take a (Maybe SrcSpan)

christetreault:
After our last meeting, it occurred to me that `todo` and `impossible` should take an optional `SrcSpan` so that we can have more useful messages. 

If we expect users to open bug reports and such regarding these we should give them the ability to give us more useful feedback.


nikivazou:
Sorry, @christetreault cannot be automatically merged :( Can you update it?

ranjitjhala:
I get some weird GH error when merging!

PR #572, opened 2016-01-13
New parsing code

spinda:
**Please don't merge this - I'm still working on it.** I'm creating this PR now just to track progress.

Outline:
- _Language.Haskell.Liquid.Parse2_ - New parser based on the code I developed in GSoC '15, adapted for the current state of the codebase.
- _Language.Haskell.Liquid.Unparse_ - Pretty-printer exclusively for generating formatted code from AST. Should be the inverse of `Parse2`'s parser, in that the output AST of `parse ∘ unparse ∘ parse` should be semantically equivalent to the input AST. Featuring colored output (via [`ansi-wl-pprint`](https://hackage.haskell.org/package/ansi-wl-pprint-0.6.7.3/docs/Text-PrettyPrint-ANSI-Leijen.html)) for prettier user messages.


ranjitjhala:
Nice! I've been wanting the colored output. However note that that stuff was also added to HughesPJ per a GitHub comment I saw...

gridaphobe:
Very excited about the new parser! But can we please not call the printer `unparse`? :D

spinda:
@ranjitjhala - Oh, cool! I guess that's what [Text.PrettyPrint.Annotated.HughesPJ](https://hackage.haskell.org/package/pretty-1.1.3.2/docs/Text-PrettyPrint-Annotated-HughesPJ.html) is. It looks like it allows you to attach arbitrary metadata (i.e. colors) to `Doc`s and then convert them into `String`s during [`renderDecorated`](https://hackage.haskell.org/package/pretty-1.1.3.2/docs/Text-PrettyPrint-Annotated-HughesPJ.html#v:renderDecorated). Do you know if there's already a package implementing color codes on top of this system? Otherwise I can implement them myself.

@gridaphobe - Sure, I had a hard time coming up with what to call it and am not attached at all to `Unparse`. Any suggestions?

gridaphobe:
> I guess that's what Text.PrettyPrint.Annotated.HughesPJ is. It looks like it allows you to attach arbitrary metadata (i.e. colors) to Docs and then convert them into Strings during renderDecorated.

Yes, also common are type annotations and other useful metadata. Idris does a ton of neat things with the annotated printers.

> Any suggestions?

`Pretty` or `PrettyPrint`! GHC uses `Outputable` which is not bad either.

ranjitjhala:
We cannot use a GHC dependency on liquid-fixpoint.

Is "Outputable" a package? It was part of GHC last I checked...

I think @spinda wants to know if there are any libraries for implementing
color.

Short answer: I don't know, there is the ansi-terminal but perhaps it would
be
best to see the docs for Annotate.HughesPJ? We could switch to
ansi-wl-pprint too
(but that may cause changes all across LH, but probably easy to fix...)

spinda:
I'm pretty sure @gridaphobe meant that `Outputable` is an example of a name we could use for this typeclass, instead of `Unparse`. Which would work, but since we're also using the GHC API we would run into issues with naming clashes. I didn't want to call it `PrettyPrint` because we already have `PPrint` which isn't necessarily an inverse of the parser and is used for other things besides source printing.

I was asking if there were any pre-made libraries out there that use Annotate.HughesPJ annotations to implement color code annotations (like `Red`, `Green`, etc). It doesn't look like there are any at the moment, so I'll use `ansi-terminal` and write something LiquidHaskell-specific for that. Something like:

```haskell
data TokenType = Keyword | Symbol | Literal
```

which will be converted to color codes in the render step. At the moment I have blue for keywords (like `not`, `int`, `real`...), yellow for symbols (parentheses, brackets, commas...), and green for literals (ints, floats, string literals...).

gridaphobe:
Yes I was just referring to `Outputable` as an example name.

> I didn't want to call it PrettyPrint because we already have PPrint which isn't necessarily an inverse of the parser and is used for other things besides source printing.

Ah, I thought that `Unparse` was meant to replace `PPrint`. I would just fix the `PPrint` instances then, they *should* be inverses of `Parse`. That we also `PPrint` some things that we don't parse is irrelevant IMO.

spinda:
Okay. I'll call it `PPrint2` for now (while I'm working on it) to match `Parse2`.

chrisdone:
@spinda what's the status on this? I wouldn't go as far to say Liquid Haskell is unusable at this point, but I can't seem to write anything without hitting `Leftover while parsing`. It's been one year so far, did you make any progress?

alanz:
Question: If I want to look at reworking the parser, should I start here, or from the current develop branch?

ranjitjhala:
This may be a bit old, so it may be easier to look at develop -- OTOH this
is probably much clearer than the develop Parser...

nikivazou:
Closing this :(

PR #580, opened 2016-01-15
Termreal

ranjitjhala:
This time, based off the **old** fixpoint (against which master builds).

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,6 @@
 # GHC 7.10
+# resolver: nightly-2015-09-18
+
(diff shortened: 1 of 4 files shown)


ranjitjhala:
Why does it say checks haven't passed even though travis says "passed" ?

ranjitjhala:
I have tweaked FP so that the interpreted DIV is `div :: Int -> Int -> Int` (not `Real -> Real -> Real`) and updated tests accordingly. The *proper* fix will be to support some form of overloading.

PR #582, opened 2016-01-18
add --scrape-imports to enable scraping qualifiers from imports

### LINKED ISSUE #576: Refinements lost when refined functions moved to another module
This module type-checks successfully:

```haskell
module Test1 where

import Control.Monad.ST (ST)

import qualified Data.Vector
import qualified Data.Vector.Mutable

{-@
example :: ST s ({ o : Data.Vector.Vector Int | vlen o = 0 })
@-}
example :: ST s (Data.Vector.Vector Int)
example = do
    l0sMutable <- newST 0
    freezeST l0sMutable

{-@ measure mlen :: Data.Vector.Mutable.MVector s a -> Int @-}

{-@
assume newST
    :: n : Int -> ST s ({ v : Data.Vector.Mutable.MVector s a | mlen v == n })
@-}
newST :: Int -> ST s (Data.Vector.Mutable.MVector s a)
newST = Data.Vector.Mutable.new
{-# INLINE newST #-}

{-@
assume freezeST
    :: mv : Data.Vector.Mutable.MVector s a
    -> ST s ({ v : Data.Vector.Vector a | vlen v == mlen mv })
@-}
freezeST :: Data.Vector.Mutable.MVector s a -> ST s (Data.Vector.Vector a)
freezeST = Data.Vector.freeze
{-# INLINE freezeST #-}
```

... but if I move everything except `example` to another module named `Test2`:

```haskell
module Test2 where

import Control.Monad.ST (ST)

import qualified Data.Vector
import qualified Data.Vector.Mutable

{-@ measure mlen :: Data.Vector.Mutable.MVector s a -> Int @-}

{-@
assume newST
    :: n : Int -> ST s ({ v : Data.Vector.Mutable.MVector s a | mlen v == n })
@-}
newST :: Int -> ST s (Data.Vector.Mutable.MVector s a)
newST = Data.Vector.Mutable.new
{-# INLINE newST #-}

{-@
assume freezeST
    :: mv : Data.Vector.Mutable.MVector s a
    -> ST s ({ v : Data.Vector.Vector a | vlen v == mlen mv })
@-}
freezeST :: Data.Vector.Mutable.MVector s a -> ST s (Data.Vector.Vector a)
freezeST = Data.Vector.freeze
{-# INLINE freezeST #-}
```

... and then change module `Test1` to import them:

```haskell
module Test1 where

import Control.Monad.ST (ST)
import Test2

import qualified Data.Vector
import qualified Data.Vector.Mutable

{-@
example :: ST s ({ o : Data.Vector.Vector Int | vlen o = 0 })
@-}
example :: ST s (Data.Vector.Vector Int)
example = do
    l0sMutable <- newST 0
    freezeST l0sMutable
```

... then type-checking module `Test1` fails with the following error:

```
 Test1.hs:(13,1)-(15,23): Error: Liquid Type Mismatch
   Inferred type
     VV : {VV : (Vector Int) | 0 <= vlen VV}
  
     VV : {VV : (Vector Int) | vlen VV == 0}
  
   In Context
     VV : {VV : (Vector Int) | 0 <= vlen VV}
```

Note that if I reduce `example` to just `newST` or just `freezeST` then the problem disappears, so this issue has something to do with their interaction.

gridaphobe:
Partially fixes #576 by allowing users to override the defaults and scrape qualifiers from imported modules. It's off by default because of the performance implications (I've seen up to 2x overhead on some modules, eg `Data.Text`).

### DIFF
--- a/scripts/travis
+++ b/scripts/travis
@@ -126,7 +126,7 @@ function setup_ghc {
 function install_dependencies {
   echo "Solving dependency constraints..."
   loud stack update
-  loud stack solver --modify-stack-yaml
+  loud stack solver --update-config
 
   echo "Installing dependencies..."
   loud stack build liquidhaskell --only-dependencies --test --no-run-tests
--- a/src/Language/Haskell/Liquid/Constraint/Qualifier.hs
+++ b/src/Language/Haskell/Liquid/Constraint/Qualifier.hs
@@ -31,10 +31,16 @@ specificationQualifiers :: Int -> GhcInfo -> SEnv Sort -> [Qualifier]
 -----------------------------------------------------------------------------------
 specificationQualifiers k info lEnv
   = [ q | (x, t) <- (tySigs $ spec info) ++ (asmSigs $ spec info) ++ (ctors $ spec info)
-        -- FIXME: this mines extra, useful qualifiers but causes a significant increase in running time
-        -- , ((isClassOp x || isDataCon x) && x `S.member` (S.fromList $ impVars info ++ defVars info)) || x `S.member` (S.fromList $ defVars info)
-        , x `S.member` (S.fromList $ defVars info)
+        , x `S.member` (S.fromList $ defVars info ++
+                                     if info `hasOpt` scrapeImports
+                                     -- NOTE: this mines extra, useful qualifiers but causes
+                                     -- a significant increase in running time, so we hide it
+                                     -- behind `--scrape-imports`
+                                     then impVars info
+                                     else [])
         , q <- refTypeQuals lEnv (getSourcePos x) (tcEmbeds $ spec info) (val t)
+        -- NOTE: large qualifiers are VERY expensive, so we only mine
+        -- qualifiers up to a given size, controlled with --max-params
         , length (q_params q) <= k + 1
     ]
     -- where lEnv = trace ("Literals: " ++ show lEnv') lEnv'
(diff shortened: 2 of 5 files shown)


nikivazou:
@gridaphobe,  since `Text` is a pathological case, why don't we have this on by default and turn it off for `Text`?

ranjitjhala:
Is Text pathological? (or does this happen elsewhere too?)

gridaphobe:
It's not pathological, there are examples even in the regression tests. `tests/pos/Map0.hs` runs 5 seconds slower with the flag and only imports `Liquid.Prelude`.

nikivazou:
The problem with this flag, as well as with the pruning of qualifiers done by `--max-params`, is that it would take a lot of thinking before the user thinks that there may be a qualifier issue and activate them. How about running liquidHaskell with these flags on, if the program is UNSAFE?

ranjitjhala:
btw, the import stuff imports *literally everything*
from those libraries, can we filter out only those
that are _used_ in the module being checked?

gridaphobe:
> The problem with this flag, as well as with the pruning of qualifiers done by --max-params, is that it would take a lot of thinking before the user thinks that there may be a qualifier issue and activate them. How about running liquidHaskell with these flags on, if the program is UNSAFE?

I would rather educate users on the inference process, so they can decide for themselves whether these flags might help.

gridaphobe:
> btw, the import stuff imports *literally everything* from those libraries, can we filter out only those that are _used_ in the module being checked?

So only scrape qualifiers from a spec for `foo` only if `foo` appears in the core bindings? I can add another flag for that, sure. It's not clear how much of a difference that will make in terms of performance (or "completeness" for that matter), but it's worth a look.

PR #584, opened 2016-01-18
L.H.L.Types.Errors no longer imports any LiquidHaskell module

christetreault:
This is good, because no we can actually use it everywhere without creating dependency cycles.


gridaphobe:
Excellent!

Issue #594, opened 2016-01-27
Partially applied abstract refinements panic


abakst:
If I (naively) attempt to partially apply an abstract refinement à la:

~~~~{.haskell}
module Boo where

{-@ data X <q :: Int -> Int -> Prop> = X (x0 :: Int) (x1 :: Int<q x0>) @-}  
data X = X Int Int  

{-@ data T <p :: Int -> Int -> Int -> Int -> Prop> = C { x :: Int, y :: Int, z :: X<p x y> } @-}
data T = C { x :: Int, y :: Int, z :: X }  
~~~~

I am met with the error `liquid: <no location info> Unexpected panic (!)
    1st argument of Boo.X is "p" that expects 1 arguments, but it has 2` which is quite confusing.

The solution to the above appears to be to rewrite the application of p as 
~~~~
X<\i1 -> {v:Int<p x y i1> | true}>
~~~~
so perhaps a message suggesting this transform would be helpful.

ranjitjhala:
Now looks like:

```

**** RESULT: ERROR *************************************************************


 tests/crash/issue594.hs:7:10-41: Error: Malformed Predicate Application
 
 7 | data T = C { x :: Int, y :: Int, z :: X }  
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
     The 1st argument of Boo.X is predicate p
     which expects 2 arguments but is given only 1
     Abstract predicates cannot be partially applied, see 
       https://github.com/ucsd-progsys/liquidhaskell/issues/594
     for possible fix.
```

PR #597, opened 2016-01-28
UserErrors

ranjitjhala:
1. Create a `UserError` type which can be thrown from almost anywhere (and caught at the top level to give a decent error message, instead of raw strings.)

2. Create a corresponding `uError :: UserError -> a` function 

3. Use the above to "fix" issue #594.

@christetreault (and everyone else) instead of `panic` we should use `uError` if it is an error related to malformed inputs.


christetreault:
Why were these guys in Misc.hs switched back to using `errorstar`?

* https://github.com/ucsd-progsys/liquidhaskell/pull/597/files#diff-958889cf92489f6b938eff72f9b360c6L35
* https://github.com/ucsd-progsys/liquidhaskell/pull/597/files#diff-958889cf92489f6b938eff72f9b360c6L80
* https://github.com/ucsd-progsys/liquidhaskell/pull/597/files#diff-958889cf92489f6b938eff72f9b360c6L84

PR #605, opened 2016-02-03
Split `REnv` into Global and Local Envs

ranjitjhala:
To enable printing of just the "local" part, the global one being the binders that correspond to module imports.


tests/test.hs
```diff
@@ -158,7 +158,7 @@ knownToFail Z3   = [ "tests/pos/linspace.hs" ]
 testCmd :: FilePath -> FilePath -> FilePath -> SmtSolver -> LiquidOpts -> String
 ---------------------------------------------------------------------------
 testCmd bin dir file smt (LO opts)
-  = printf "cd %s && %s --verbose --smtsolver %s %s %s" dir bin (show smt) file opts
+  = printf "cd %s && %s --smtsolver %s %s %s" dir bin (show smt) file opts
```
gridaphobe: please don't remove the verbose output, it can be very useful on travis!

ranjitjhala: has it ever been?

ranjitjhala: Plus:

+ Don't want to have the numbers generated by `stack/cabal test` to be corrupted by debugging, as they are also used to measure performance over time,

+ We can always add the `--verbose` as an extra parameter to `stack/cabal test` if we really want to generate the verbose logs on travis, no need to hardwire the thing inside the script...

gridaphobe: Fair enough, let's just add `--verbose` to the travis scripts then. It has been invaluable on the (rare) occasion that I can't reproduce the failure locally.

ranjitjhala:
Some more changes, to *avoid* throwing the VV into the environment (and hence decrease the number of non-linear KVars).

ranjitjhala:
merging because travis is ok with it...

PR #623, opened 2016-03-02
periodic develop merge

gridaphobe:


ranjitjhala:
ugh. is this because i merged #624 first (yuck, sorry!)

gridaphobe:
I think it was fussing already due to the lingering `tests/pos/Gradual.hs` failure, but I've marked that as known-to-fail for the time being (the gradual stuff is highly experimental).

PR #629, opened 2016-03-07
Rework GHC API setup and configuration

spinda:
This:

- Cleans up and organizes `GHC/Interface.hs`
- Tells GHC to put the .hi/.o/etc files in a temporary directory which is automatically cleaned up on exit, instead of letting it put them next to the .hs files and deleting them every time.
- Ensures DynFlags tweaks that LiquidHaskell relies on aren't overridden by other GHC options.
- Extracts additional .spec/.hquals search paths from the provided `ghc-option`s.
- Prints GHC messages in Loud (verbose) mode for debugging.


gridaphobe:
> Tells GHC to put the .hi/.o/etc files in a temporary directory which is automatically cleaned up on exit, instead of letting it put them next to the .hs files and deleting them every time.

Nice, but I wonder if we shouldn't save the artifacts instead of deleting them. This could cut down LH runtimes substantially for small files. And GHC has a bunch of logic for determining when an existing .hi file can be reused, I think it's all encapsulated in the CompManager build-style that we use.

> Prints GHC messages in Loud (verbose) mode for debugging.

I'm not sure about this. We already produce a TON of output in verbose mode (too much IMO). Does GHC add anything useful here?

spinda:
>Nice, but I wonder if we shouldn't save the artifacts instead of deleting them. This could cut down LH runtimes substantially for small files. And GHC has a bunch of logic for determining when an existing .hi file can be reused, I think it's all encapsulated in the CompManager build-style that we use.

This would be great except that at the moment we need to extract info from imported modules that is only available when GHC does a fresh compile. The `.lqhi` stuff (which *will* get merged in, it's my next goal now that I've wrapped up most of the Cabal stuff) would fix this.

Either way, for the moment using a temporary directory is much cleaner than what we're doing now.

> Prints GHC messages in Loud (verbose) mode for debugging.

Seeing warnings/errors from GHC helped to debug an issue with compiling C files as I was working on this. It's not essential though, and can be taken out if it's not wanted.

gridaphobe:
Ok, sounds good then. 

(At some point we may want to split the verbose/debug messages into different flags to make `-v` more usable :)

Issue #612, opened 2016-02-16
Weird import behavior when using stack


rpglover64:
Consider two simple files in a stack project:

`FileA.hs`:
```haskell
module FileA where
import Data.List (foldl')
```

`FileB.hs`:
```haskell
module FileB where
import Data.List
foo = foldl'
```

Running `stack exec -- liquid [...]` gives the following output:
```
% stack exec -- liquid FileA.hs
LiquidHaskell Copyright 2009-15 Regents of the University of California. All Rights Reserved.


**** DONE:  annotate ***********************************************************


**** ERROR: Invalid Source *****************************************************


 FileA.hs:2:19-24: Error: GHC Error
     Module ‘Data.List’ does not export ‘foldl'’
```
```
% stack exec -- liquid FileB.hs
LiquidHaskell Copyright 2009-15 Regents of the University of California. All Rights Reserved.


**** DONE:  annotate ***********************************************************


**** ERROR: Invalid Source *****************************************************


 FileB.hs:3:7-12: Error: GHC Error
     Not in scope: ‘foldl'’
```

As far as I can tell, `liquid` is opening the appropriate interface files (at least according to `strace`), but not using them properly (whatever that means)

rpglover64:
There's a relevant IRC conversation [here](http://ircbrowse.net/day/haskell/2016/02/16?id=22288739&timestamp=1455596409#t1455596409)

(copied and cleaned up version):
```
20:20:58 <rpglover64> I'm having some issues with running liquidhaskell using `stack exec`.
20:21:28 <synergistics> shachaf: I got that, but I didn't see how s and t are themselves identities
20:21:32 <rpglover64> MyModule.hs:6:22-27: Error: GHC Error     Module ‘Data.List’ does not export ‘foldl'’
20:21:52 <rpglover64> which makes no sense to me
20:24:31 <rpglover64> (I do have an "import Data.List (foldl')" line in my file)
20:28:50 <tommd> rpglover64: You misunderstand the error.  It is telling you that Data.List does not define the function foldl', not that you aren't importing it.
20:29:05 <rpglover64> But Data.List does define foldl'
20:29:13 <tommd> rpglover64: Which isn't true for any distributed module named Data.List I know.
20:29:17 <tommd> Yes agreed.
20:29:30 <rpglover64> The error points to the import line
20:29:35 <tommd> rpglover64: So, do you have something in a Data directory called List.hs?
20:29:41 <rpglover64> no
[...]

PR #630, opened 2016-03-07
Make sure GHC versions match

spinda:
This should fix the issue described in #612.

I haven't actually verified that this detects the error case correctly&mdash;could someone help with that?


gridaphobe:
I think the trick is to `stack install` this branch and then `stack exec liquid` in a project with a different version of ghc.

ranjitjhala:
So shall we merge this?

spinda:
No, after testing it needs some more work.

nikivazou:
Closing this :(

Issue #631, opened 2016-03-07
Refine type error when haskell measure does not satisfy singleton type


nikivazou:
In the following example, the  singleton type for the haskell measure `foo` cannot be verified

[tests/neg/BadHMeas.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/0d19cfbde1dbf695b57d17573eb448985b1417dc/tests/neg/BadHMeas.hs)

 because the type of `fst` is not singleton `fst :: (a, b) -> a`

Currently, a type error is thrown for the body of `foo`. 

Instead, we should throw the following error:

```
 tests/pos/Fixme.hs:4:1-3: Error: Cannot promote Haskell function Main.foo to lo
```

PR #632, opened 2016-03-07
Hmeas error

nikivazou:
fix for #631

### DIFF
--- a/src/Language/Haskell/Liquid/Constraint/Generate.hs
+++ b/src/Language/Haskell/Liquid/Constraint/Generate.hs
@@ -89,7 +89,8 @@ import Language.Haskell.Liquid.Types.Visitors         hiding (freeVars)
 import Language.Haskell.Liquid.Types.PredType         hiding (freeTyVars)
 import Language.Haskell.Liquid.Types.Meet
 import Language.Haskell.Liquid.GHC.Misc          ( isInternal, collectArguments, tickSrcSpan
-                                                 , hasBaseTypeVar, showPpr, isDataConId)
+                                                 , hasBaseTypeVar, showPpr, isDataConId
+                                                 , sourcePosSrcSpan)
 import Language.Haskell.Liquid.Misc
 import Language.Fixpoint.Misc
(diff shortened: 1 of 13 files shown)


src/Language/Haskell/Liquid/Constraint/Generate.hs
```diff
@@ -167,16 +168,17 @@ initEnv info
        f2       <- refreshArgs' $ assm info                  -- assumed refinements      (for imported vars)
        f3       <- refreshArgs' $ vals asmSigs sp            -- assumed refinedments     (with `assume`)
        f40      <- refreshArgs' $ vals ctors sp              -- constructor refinements  (for measures)
+       f5       <- refreshArgs' $ vals inSigs sp            -- assumed refinedments     (with `assume`)
```
gridaphobe: the comment here seems to be wrong

gridaphobe:
I'm a bit skeptical of this patch, it looks to me like the following

```haskell
{-@ measure len @-}
{-@ len :: [a] -> {v:Int | v > 0} @-}
len :: [a] -> Int
len [] = 0
len (x:xs) = 1 + len xs
```

would produce a "cannot promote `len`" error, when the error is *not* that we cannot promote `len` but rather that the `Pos` refinement is bogus.

nikivazou:
@gridaphobe the above is just producing a type error

nikivazou:
So, I will fix the comment and merge?

gridaphobe:
My example is effectively giving two types to `len`

```
len :: xs:[a] -> {v:Int | v = len xs}
len :: xs:[a] -> {v:Int | v > 0}
```

which means we should be checking that `len` has the combined type

```
len :: xs:[a] -> {v:Int | v = len xs && v > 0}
```

Can you point out where we're `join`ing the two types? Or how we would end up checking both specifications? 

This is a subtlety that I think should be pointed out in comments because we're either keeping **two** types for `len` around (when I would expect we have an invariant that we only have **one** type for `len` in the environment), or we're joining the two types (in which case we must need to track the types of obligations each conjunct represents).

PR #650, opened 2016-03-23
Don't drop refinements for int-real mismatch

### LINKED ISSUE #602: Refinement dropped when specializing type class constraint
The following program incorrectly type-checks:

```haskell
{-@
class Foo a where
    foo :: { x : a | x /= 0 } -> a
@-}

class Foo a where
    foo :: a -> a

instance Foo Double where
    foo = id

-- example :: (Num a, Foo a) => a
example :: Double
example = foo 0
```

If you use the commented out type signature for `example` then type-checking correctly fails.

I tested this against a freshly checked out repository following the `stack` build instructions.

gridaphobe:
depends on https://github.com/ucsd-progsys/liquid-fixpoint/pull/216

fixes #602 

----

Also includes minor fixes/improvements to the class specifications


ranjitjhala:
Thanks!!!

PR #653, opened 2016-03-26
Add refinement types for the `bytestring` library

Gabriel439:
The general rules I applied for refinements were:

* Preconditions were only added for things that would cause crashes or
  exceptions.  I did not protect against silly but valid inputs (like
  `Data.ByteString.take (-1)`)
* I added as many postconditions as humanly possible
* I did not add refinements for deprecated functions in order to avoid Liquid
  Haskell failing when the `bytestring` library eventually eliminates them

I didn't cover every single module in the `bytestring` library.  The modules that
I haven't done yet are all the `Builder`-related modules:

* `Data.ByteString.Builder`
* `Data.ByteString.Builder.Extra`
* `Data.ByteString.Builder.Prim`
* `Data.ByteString.Lazy.Builder`
* `Data.ByteString.Lazy.Builder.ASCII`
* `Data.ByteString.Lazy.Builder.Extras`

### DIFF
--- a/include/Data/ByteString.spec
+++ b/include/Data/ByteString.spec
@@ -1,0 +1,380 @@
+module spec Data.ByteString where
+
+measure bslen :: Data.ByteString.ByteString -> { n : Int | 0 <= n }
+
+invariant { bs : Data.ByteString.ByteString  | 0 <= bslen bs }
+
+assume empty :: { bs : Data.ByteString.ByteString | bslen bs == 0 }
+
+assume singleton
+    :: Data.Word.Word8 -> { bs : Data.ByteString.ByteString | bslen bs == 1 }
+
+assume pack
+    :: w8s : [Data.Word.Word8]
+    -> { bs : Data.ByteString.ByteString | bslen bs == len w8s }
+
+assume unpack
+    :: bs : Data.ByteString.ByteString
+    -> { w8s : [Data.Word.Word8] | len w8s == bslen bs }
+
+assume cons
+    :: Data.Word.Word8
+    -> i : Data.ByteString.ByteString
+    -> { o : Data.ByteString.ByteString | bslen o == bslen i + 1 }
+
+assume snoc
+    :: i : Data.ByteString.ByteString
+    -> Data.Word.Word8
+    -> { o : Data.ByteString.ByteString | bslen o == bslen i + 1 }
+
+assume append
+    :: l : Data.ByteString.ByteString
+    -> r : Data.ByteString.ByteString
+    -> { o : Data.ByteString.ByteString | bslen o == bslen l + bslen r }
+
+head :: { bs : Data.ByteString.ByteString | 1 <= bslen bs } -> Data.Word.Word8
+
+assume uncons
+    :: i : Data.ByteString.ByteString
+    -> Maybe (Data.Word.Word8, { o : Data.ByteString.ByteString | bslen o == bslen i - 1 })
+
+assume unsnoc
+    :: i : Data.ByteString.ByteString
+    -> Maybe ({ o : Data.ByteString.ByteString | bslen o == bslen i - 1 }, Data.Word.Word8)
+
+last :: { bs : Data.ByteString.ByteString | 1 <= bslen bs } -> Data.Word.Word8
+
+tail :: { bs : Data.ByteString.ByteString | 1 <= bslen bs } -> Data.Word.Word8
+
+init :: { bs : Data.ByteString.ByteString | 1 <= bslen bs } -> Data.Word.Word8
+
+assume null
+    :: bs : Data.ByteString.ByteString
+    -> { b : Bool | Prop b <=> bslen bs == 0 }
+
+assume length :: bs : Data.ByteString.ByteString -> { n : Int | bslen bs == n }
+
+assume map
+    :: (Data.Word.Word8 -> Data.Word.Word8)
+    -> i : Data.ByteString.ByteString
+    -> { o : Data.ByteString.ByteString | bslen o == bslen i }
+
+assume reverse
+    :: i : Data.ByteString.ByteString
+    -> { o : Data.ByteString.ByteString | bslen o == bslen i }
+
+assume intersperse
+    :: Data.Word.Word8
+    -> i : Data.ByteString.ByteString
+    -> { o : Data.ByteString.ByteString | (bslen i == 0 <=> bslen o == 0) && (1 <= bslen i <=> bslen o == 2 * bslen i - 1) }
+
+assume intercalate
+    :: l : Data.ByteString.ByteString
+    -> rs : [Data.ByteString.ByteString]
+    -> { o : Data.ByteString.ByteString | len rs == 0 ==> bslen o == 0 }
+
+assume transpose
+    :: is : [Data.ByteString.ByteString]
+    -> { os : [{ bs : Data.ByteString.ByteString | bslen bs <= len is }] | len is == 0 ==> len os == 0}
+
+foldl1
+    :: (Data.Word.Word8 -> Data.Word.Word8 -> Data.Word.Word8)
+    -> { bs : Data.ByteString.ByteString | 1 <= bslen bs }
+    -> Data.Word.Word8
+
+foldl1'
+    :: (Data.Word.Word8 -> Data.Word.Word8 -> Data.Word.Word8)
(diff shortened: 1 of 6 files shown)


ranjitjhala:
Wow this is awesome thanks!!! 

@gridaphobe wonder how these compare to our sigs in the benchmarks...

At any rate can you merge? Or does it cause some odd conflict with the benchmarks that we need to resolve?

ranjitjhala:
@gridaphobe am I correct that these failures are due to some "duplicate"
specifications (both in the .spec files and in `benchmarks/bytestring/Data/Bytestring/**.hs` 
files themselves?) (am currently only on phone...)

include/Data/ByteString/Unsafe.spec
```diff
@@ -0,0 +1,28 @@
...
+    -> { n : Int | 0 <= n && n < bslen bs }
+    -> Data.Word.Word8
+
+assume unsafeTake
+    :: n : { n : Int | 0 <= n }
+    -> i : { i : Data.ByteString.ByteString | n <= bslen i }
+    -> { o : Data.ByteString.ByteString | bslen o == n }
+
+assume unsafeDrop
+    :: n : { n : Int | 0 <= n }
```
gridaphobe: FYI, you can write `n : { Int | 0 <= n }` for the above :)

include/Data/ByteString.spec
```diff
@@ -0,0 +1,380 @@
+module spec Data.ByteString where
+
+measure bslen :: Data.ByteString.ByteString -> { n : Int | 0 <= n }
```
gridaphobe: I think this line needs to be moved to `Data/ByteString/Unsafe.spec`, which should be the lowest file in the import hierarchy (`Data/ByteString/Internal.spec` would make sense too).

Right now the testsuite is failing with

```
 /home/ubuntu/liquidhaskell/include/Data/ByteString/Unsafe.spec:20:8-17: Error: Bad Type Specification
 assumed type Data.ByteString.Unsafe.unsafeTake :: n:{n : Int | 0 <= n} -> {i : ByteString | n <= bslen i} -> {o : ByteString | bslen o == n}
     Sort Error in Refinement: {i##11 : Data.ByteString.Internal.ByteString | n <= bslen i##11}
     Unbound Symbol bslen
 Perhaps you meant: plen, len, fplen
```

which makes sense as `Data.ByteString` imports `Data.ByteString.Unsafe` and thus `bslen` doesn't exist yet.

gridaphobe:
@ranjitjhala the current problem is with a measure being defined too late in the import hierarchy, though I wouldn't be surprised if we run into a duplicate issue once that's resolved.

Gabriel439:
Is there a way that I can run the same test locally?

gridaphobe:
Yes, you can run the test suite with `stack test`, but it's quite long so I'd use

```
stack test --test-arguments="-p bytestring"
```

to restrict it to the bytestring benchmarks (it'll still take about 10 minutes to run).

Gabriel439:
I didn't realize there were so many annotations in the `benchmarks/` section.  In that case what I may do is just transplant what is already there to the `include/` directory

Gabriel439:
Actually, I have one more question, which is how to make sure that `liquid` pulls in changes to `include/` directory.  I've notice on several occasions that if I've built `liquid` and later make changes to the `include/` directory the `liquid` tool doesn't reflect those changes, even if I reinstall it.

For example, after testing the changes on my branch I tried to checkout `master` and run the test again and it still failed.  Wiping the `.stack-work` directory and reinstalling doesn't fix the problem and the only solution I know that works is to delete the repository and do a fresh clone to get it to reset correctly.

gridaphobe:
Hmm that's odd. IIRC `stack build` does cabal's `install` step (`copy` + `register`) which should handle an updated `include`.

There's also the (poorly named) `-finclude` option to search for the `include` directory in the current directory, which is useful for development since you can change the `include`s without having to `stack build`.

Gabriel439:
So that helps if I'm running the `liquid` tool myself, but `stack test` still fails with a stale view of the `include` directory

gridaphobe:
Hmm, I'm afraid I can't reproduce your problem locally. Here's what I'm trying

First I run
```
stack test --fast liquidhaskell --test-arguments="-p Unit/pos/zipW2"
```
which passes as expected. Then I tweak https://github.com/ucsd-progsys/liquidhaskell/blob/master/include/Prelude.spec#L25
to say
```
assume GHC.Num.+                :: (GHC.Num.Num a) => x:{a | false} -> y:a -> {v:a | v = x + y }
```
making it unsafe to ever call `+`. Then I re-run
```
stack test --fast liquidhaskell --test-arguments="-p Unit/pos/zipW2"
```
and it fails as expected.

Have you made `cabal` aware of the new ByteString specs (see https://github.com/ucsd-progsys/liquidhaskell/blob/master/liquidhaskell.cabal#L24)? I just noticed that's not in the PR as it stands, which we'd need to fix regardless. If cabal doesn't know about the new specs, they won't be copied and things will probably break in strange ways :)

Gabriel439:
Alright, so I narrowed down the problem.  The issue was that I was that liquid haskell was completely ignoring my `include/Data/ByteString/Internal.spec` file.

The easiest way to reproduce this is to check out my `reproduce_1` branch: https://github.com/Gabriel439/liquidhaskell/tree/gabriel/reproduce_1

... which has this small addition:

```diff
diff --git a/include/Data/ByteString/Internal.spec b/include/Data/ByteString/Internal.spec
new file mode 100644
index 0000000..5d96625
--- /dev/null
+++ b/include/Data/ByteString/Internal.spec
@@ -0,0 +1,3 @@
+module spec Data.ByteString.Internal where
+
+******
diff --git a/liquidhaskell.cabal b/liquidhaskell.cabal
index b8c0bdd..64434a7 100644
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -21,6 +21,7 @@ data-files: include/*.hquals
           , include/Control/*.spec
           , include/Data/*.hquals
           , include/Data/*.spec
+          , include/Data/ByteString/*.spec
           , include/Data/Text/*.spec
           , include/Data/Text/Fusion/*.spec
           , include/Data/Text/Lazy/*.spec
```

I included a deliberate parse error in the `include/Data/ByteString/Internal.spec` file to see if it was getting used at all.  However, if I run the benchmark on the `benchmarks/bytestring-0.9.2.1/Data/ByteString/Internal.hs` file then my spec file is ignored:

```bash
$ stack test --fast liquidhaskell --test-arguments="-p Benchmarks/bytestring/Data/ByteString/Internal.hs"
liquidhaskell-0.6.0.0: test (suite: test, args: -p Benchmarks/bytestring/Data/ByteString/Internal.hs)

Tests
  Benchmarks
    bytestring
      Data/ByteString/Internal.hs: OK (13.74s)

```

Gabriel439:
Okay, so my guess for why this is happening is that Liquid Haskell only applies the `*.spec` file if you import the module of that name, but since `Data.ByteString.Internal` is not importing itself then Liquid Haskell doesn't check the `Data/ByteString/Internal.spec` file

gridaphobe:
@Gabriel439 Sorry, that's correct, we don't currently look for `.spec` files that match the current module we're checking. We only do that for imports. Though I don't think there's a good reason for this..

ranjitjhala:
@gridaphobe can we meet and resolve this PR on Friday (or tomorrow in HC)?

Looks like there are a couple of different things going on here...

gridaphobe:
@ranjitjhala sure either works for me.

Gabriel439:
I believe we should discard what I wrote and take the existing specs used for the benchmarks and move them upstream to the defaults `include`s.  The specs in the benchmarks appear to be much more comprehensive and more well thought out than my work

ranjitjhala:
@Gabriel439 -- perhaps, but it would be nice to use your files as a starting point! 

At the very least their mere presence has pointed out some glitches, which I 
am presently trying to address. My plan is to:

1. fix those problems & merge this PR, and then,
2. gradually pull in / replace your specs with the ones from  `benchmarks/bytestring` 

Does that sound reasonable? 

Thanks!

ranjitjhala:
Closed and merged via https://github.com/ucsd-progsys/liquidhaskell/pull/674 (Thanks @Gabriel439 !)

Gabriel439:
Thank you! :)

PR #654, opened 2016-03-28
strengthen record selectors with singleton refinement

### LINKED ISSUE #595: Dependent field refinements dropped when using field accessor functions
The following program correctly type-checks:

```haskell
import Data.Vector

{-@
data Test = Test
    { v  :: Vector ()
    , x0 :: { x0 : Bool | vlen v < 1 ==> Prop x0 }
    }
@-}

data Test = Test
    { v  :: Vector ()
    , x0 :: Bool
    }

example :: Test -> ()
example (Test v_ x0_) =
    if x0_
    then ()
    else v_ ! 0
```

However, if I change `example` to use field accessors instead of pattern matching on the constructor then the program fails to type-check:

```haskell
import Data.Vector

{-@
data Test = Test
    { v  :: Vector ()
    , x0 :: { x0 : Bool | vlen v < 1 ==> Prop x0 }
    }
@-}

data Test = Test
    { v  :: Vector ()
    , x0 :: Bool
    }

example :: Test -> ()
example t =
    if x0 t
    then ()
    else v t ! 0
```

gridaphobe:
This strengthens record selector types with the singleton refinement for the generated measure, ie given

```
data Foo = Foo { getFoo :: Int }
{-@ data Foo = Foo { getFoo :: Nat } @-}
```

we would produce the following type for `getFoo`

```
getFoo :: x:Foo -> {v:Int | v >= 0 && v = getFoo x}
```

Fixes #595

### DIFF
--- a/src/Language/Haskell/Liquid/Bare/DataType.hs
+++ b/src/Language/Haskell/Liquid/Bare/DataType.hs
@@ -1,10 +1,11 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TupleSections #-}
 
 module Language.Haskell.Liquid.Bare.DataType (
     makeConTypes
   , makeTyConEmbeds
-
+  , makeRecordSelectorSigs
   , dataConSpec
   , meetDataConSpec
   ) where
@@ -22,11 +23,11 @@ import Data.Maybe
 import qualified Data.List           as L
 import qualified Data.HashMap.Strict as M
 
-import Language.Fixpoint.Types (Symbol, TCEmb)
+import Language.Fixpoint.Types (Symbol, TCEmb, mkSubst, Expr(..), Brel(..), subst, mkEApp)
 
 import Language.Haskell.Liquid.GHC.Misc (sourcePos2SrcSpan, symbolTyVar)
 import Language.Haskell.Liquid.Types.PredType (dataConPSpecType)
-import Language.Haskell.Liquid.Types.RefType (mkDataConIdsTy, ofType, rApp, rVar, uPVar)
+import Language.Haskell.Liquid.Types.RefType (mkDataConIdsTy, ofType, rApp, rVar, strengthen, uPVar, uReft)
 import Language.Haskell.Liquid.Types
 import Language.Haskell.Liquid.Types.Meet
 import Language.Haskell.Liquid.Misc (mapSnd)
@@ -39,6 +40,7 @@ import Language.Haskell.Liquid.Bare.Env
 import Language.Haskell.Liquid.Bare.Lookup
 import Language.Haskell.Liquid.Bare.OfType
 
+-- import Debug.Trace
 
 -----------------------------------------------------------------------
 -- Bare Predicate: DataCon Definitions --------------------------------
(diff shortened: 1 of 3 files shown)


ranjitjhala:
This has been a long standing issue  !! Thanks !!!! Any idea why CI is fussing?

gridaphobe:
@ranjitjhala it looks like (part of) the issue is combining this with `--totality`. We now assert a type for the record selectors, but the types are not strong enough to prove totality. Eg

```
data Foo = Bar {getBar :: Int} | Baz
```

`getBar` is partial, and our generated type is not strong enough to prove that the `error` inside `getBar` is unreachable.

I'm not sure why this is only becoming an issue now, that suggests that before we must have been inferring FALSE(!) for record selectors.. But I suggest we don't check record selectors at all, they're generated by GHC after all.

src/Language/Haskell/Liquid/Bare/DataType.hs
```diff
@@ -156,3 +158,34 @@ makeTyConEmbeds' :: TCEmb (Located Symbol) -> BareM (TCEmb TyCon)
...
+    su   = mkSubst $ [ (x, EApp (EVar x) (EVar z)) | x <- xs ]
+    args = tyArgs dcp
+    xs   = map fst args
+    z    = "lq$recSel"
+    res  = dropPreds (tyRes dcp)
+
+    -- FIXME: this is clearly imprecise, but the preds in the DataConP seem
+    -- to be malformed. If we leave them in, tests/pos/kmp.hs fails with
+    -- a malformed predicate application. Niki, help!!
+    dropPreds = fmap (\(MkUReft r _ps ss) -> MkUReft r mempty ss)
```
gridaphobe: @nikivazou I'm not sure what's going on here, but we seem to create a malformed predicate application in datacon types, specifically in the result type.

It was already there prior to this patch (look at the type for the datacon in `tests/pos/kmp.hs`), but somehow it hasn't caused problems so far.

nikivazou: Can you give the examples of the malformed types?

gridaphobe: This is the error LH produces.
```
 tests/pos/kmp.hs:64:10-13: Error: Bad Type Specification
 assumed type KMP.alen :: forall a <rng :: Int-> a -> Prop>.
                          lq$recSel:(Arr <{VV : a<rng x1> | true}> a) -> {v : Int | v >= 0
                                                                                    && v == alen lq$recSel}
     Missing arguments on  rng
```
(Btw this looks more like an unbound variable rather than a "missing argument"..)

But if I disable the "missing argument" check, we get these final types.

```
KMP.A : forall a <rng :: GHC.Types.Int-> a -> Prop>.
         alen##KMP.A##r10i:{VV : GHC.Types.Int | VV >= 0} -> aval##KMP.A##r10i:(i:{VV : GHC.Types.Int | VV >= 0
                                                                                                        && VV < alen##KMP.A##r10i} -> {VV : a<rng i> | true}) -> {VV : (KMP.Arr <\lq_tmp$db##6 VV -> {VV : _<rng> | true}> a) | alen VV == alen##KMP.A##r10i}
```

```
 KMP.alen : forall a <rng :: GHC.Types.Int-> a -> Prop>.
            lq$recSel:(KMP.Arr <{VV : a<rng> | true}> a) -> {v##16 : GHC.Types.Int | v##16 >= 0
                                                                                     && v##16 == alen lq$recSel}
```

Ah, it turns out the `rng` application to `Arr` is not quite the same. Curious, when you construct the datacon type do you do any special processing of abstract predicates? I was just extracting the types directly from the DataConP.

nikivazou: Sorry, I did not have the chance to look at this yet. Is it still open?

gridaphobe: I merged the PR (it works just fine if you drop the predicates), but this would still be good to sort out.

Issue #649, opened 2016-03-23
Generic measure specialization appears to remain unused


j6carey:
With a file `t.hs` containing:
```haskell
module T where

import Data.Proxy
import Data.Word

{-@ class measure sizeOf :: forall a . Proxy a -> Int @-}
{-@
instance measure sizeOf :: Proxy Word16 -> Int
sizeOf (Proxy) = 2
@-}
{-@
instance measure sizeOf :: Proxy Word32 -> Int
sizeOf (Proxy) = 4
@-}

{-@
bar :: { x : Int | x = (sizeOf (Proxy :: Proxy Word32)) }
    -> ()
@-}
bar :: Int -> ()
bar _ = ()

{-@
qux :: { x : Int | x = 4 }
    -> ()
@-}
qux :: Int -> ()
qux = bar
```

I see this incorrect error:
```
$ stack exec liquid -- t.hs
LiquidHaskell Copyright 2009-15 Regents of the University of California. All Rights Reserved.


**** DONE:  Cleaned Files ******************************************************
 

**** DONE:  Parsed All Specifications ******************************************
 

**** DONE:  Loaded Targets *****************************************************
 

**** DONE:  Extracted Core using GHC *******************************************
 

**** DONE:  generateConstraints ************************************************
 
Working   0% [.................................................................]
Done solving.

**** DONE:  solve **************************************************************
 

**** DONE:  annotate ***********************************************************
 

**** RESULT: UNSAFE ************************************************************


 t.hs:28:7-9: Error: Liquid Type Mismatch
 
 28 | qux = bar
            ^^^
 
   Inferred type
     VV : {VV : (GHC.Types.Int <>) | VV == 4}
  
   not a subtype of Required type
     VV : {VV : (GHC.Types.Int <>) | VV == sizeOf (Data.Proxy.Proxy : (Data.Proxy.Proxy  GHC.Word.Word32))}
  
   In Context
$
```

ranjitjhala:
Could this have to do with the recently fixed Word literal issue..?

gridaphobe:
I think the problem is not so much that the instance measure is unused, but that we don't support type ascriptions in refinements, ie we don't support the

```
Proxy :: Proxy Word32
```

expression. I'm honestly surprised that it even parses.

j6carey:
Here is a variant that does not use `Proxy :: Proxy Word32`:
```haskell
module T where

import Data.Proxy
import Data.Word
import GHC.Ptr

[...]

PR #657, opened 2016-03-29
prevent declaring multiple instance measures for the same measure-tycon pair

gridaphobe:
Partial fix for #649.

This fixes the soundness issue by throwing an error if we have two instances of a measure for the same tycon, but doesn't allow for giving precise measure definitions for eg Ptr Word16 vs Ptr Word32.


ranjitjhala:
@gridaphobe -- i sense if you dig a bit more, you'll find why they get
generalized to `Ptr a` (and hence conjoined :))

PR #666, opened 2016-04-06
Develop

nikivazou:


src/Language/Haskell/Liquid/Bare/Check.hs
```diff
@@ -192,11 +199,14 @@ checkTy :: (Doc -> Error) -> TCEmb TyCon -> TCEnv -> SEnv SortedReft -> Located
 checkTy mkE emb tcEnv env t = mkE <$> checkRType emb env (val $ txRefSort tcEnv emb t)
 
 checkDupIntersect     :: [(Var, Located SpecType)] -> [(Var, Located SpecType)] -> [Error]
-checkDupIntersect xts mxts = concatMap mkWrn dups
+checkDupIntersect xts asmSigs = concatMap mkWrn $ trace msg dups
```
nikivazou: Should we remove this trace message?

gridaphobe: please! :)

ranjitjhala: Yes my bad sorry!

PR #669, opened 2016-04-11
Qualify `Vector` type in `Data/Vector.spec`

Gabriel439:
Liquid Haskell fails if you use the `Data.Vector` module due to the unqualified reference in `Data/Vector.spec` and succeeds after this change.

### DIFF
--- a/include/Data/Vector.spec
+++ b/include/Data/Vector.spec
@@ -2,7 +2,7 @@ module spec Data.Vector where
 
 import GHC.Base
 
-data variance Vector covariant
+data variance Data.Vector.Vector covariant
 
 
 measure vlen    :: forall a. (Data.Vector.Vector a) -> Int


gridaphobe:
Thanks! I'll merge as soon as the CI is happy.

sorawee:
This is weird. I was able to `import Data.Vector` prior your pull request without any problem...

ranjitjhala:
​Hmm.​ I ran into the exact same problem today. (Which needed this very
fix...)

Has something recently changed in how the qualified imports are resolved?

(Regardless, many thanks @Gabriel439!)

Gabriel439:
I believe the covariance annotation which was affected was only recently added, which might explain why this problem didn't strike before.

It also might require more than just importing `Data.Vector` to trigger the problem

PR #672, opened 2016-04-13
Diff

ranjitjhala:
Various crosscutting changes to fix the `--diff` mode, including:

1. Modify `DiffCheck` to ignore definitions in _other_ files 
  (undoubtedly will come back to bite later...)

2. Modify `Parse` and `Bare` and `GhcSpec` to ensure that the entities in 
    `GhcSpec` are `Located SpecType` (rather than `SpecType`).


ranjitjhala:
Merge if/when CI gives green light...

PR #679, opened 2016-04-18
Counter examples

gridaphobe:
Automatically add a counter-example to failed subtyping queries if possible.

I think this is stable enough to merge into develop now, by which I mean the failure to generate a counter-example (for whatever reason) should not break LH. Counter-example generation is automatic, though we could add a flag to toggle it if desired.

If we can find a counter-example, it will be displayed inline, eg

```
 tests/neg/ListElem.hs:12:1-8: Error: Liquid Type Mismatch

 12 | listElem _ []      = False
      ^^^^^^^^

   Inferred type
     VV : {VV : GHC.Types.Bool | VV == GHC.Types.True}
     VV = True

     VV : {VV : GHC.Types.Bool | Prop VV <=> Set_mem ?b (listElts ?a)}

   In Context
     ?a : {?a : [a] | len ?a >= 0}
     ?a = [1]

     ?b : a
     ?b = 0
```

Otherwise we log a (verbose-mode) warning with the error message.


ranjitjhala:
Very exciting!!!

Can we make it not default yet (until the dust settles a bit?)

PR #681, opened 2016-04-20
Fixing Type Sigs

ranjitjhala:
But also changes the generated Kuts -- c.f. `addKuts` but those *should* only be tickled by `--eliminate` (in the `graphs` branch of `liquid-fixpoint`, which is not yet merged.)


ranjitjhala:
Please pull asap. I don't want to be re-fixing these conflicts!

.ghci
```diff
@@ -1 +1,31 @@
 :set -isrc
+
+bind 0 x: {v: int | 0 < v}
```
gridaphobe: What's this about? Bad paste?

liquidhaskell.cabal
```diff
@@ -135,7 +135,7 @@ Library
                 , vector >= 0.10
                 , hashable >= 1.2
                 , unordered-containers >= 0.2
-                , liquid-fixpoint >= 0.5 && < 0.6
+                , liquid-fixpoint
```
gridaphobe: Please leave this bound in place, liquid-fixpoint changes API so much it's pretty much guaranteed the next hackage release won't work with current LH.

ranjitjhala: Feel free to fiddle and fix with the bounds -- I removed it as the thing wasn't building ...

.ghci
ranjitjhala: Haven't the slightest idea --- didn't even know that file existed!

PR #684, opened 2016-04-21
New ordering heuristics

ranjitjhala:
And changes for adding Kuts to FQ query, and code for --eliminate which is very promising but still blows up now and then...

Matching PR for:

ucsd-progsys/liquid-fixpoint#222


ranjitjhala:
(oops, need to unmask the tests that fail on `--eliminate`)

PR #688, opened 2016-04-21
Invariants

nikivazou:
Allow different invariants on TyCon with different argumements 

See [tests/pos/MultipleInvariants.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/e4bc6c6b165fbf81f6b865ca3efa8b3fc2c1860d/tests/pos/MultipleInvariants.hs)


src/Language/Haskell/Liquid/Constraint/Types.hs
```diff
@@ -233,30 +233,55 @@ elemHEnv x (HEnv s) = x `S.member` s
...
+      Nothing   -> Nothing
+      Just invs -> Just (catMaybes $ goodInvs ts <$> invs)
+lookupRInv _ _
+  = Nothing 
+
+goodInvs :: [SpecType] -> RInv -> Maybe SpecType
+goodInvs _ (RInv []  t) 
+  = Just t 
+goodInvs ts (RInv ts' t)
+  | ts' == (toRSort <$> ts)
```
gridaphobe: This will only work if the `Type`s match exactly. What if I have

```
invariant {v: Ptr a | ptrSize v >= 0}
invariant {v: Ptr Word16 | ptrSize v = 2}
```

A `Ptr Word16` would then only get the 2nd invariant, but it should get both. Furthermore, a `Ptr b` would probably get neither right?

nikivazou: No, the first invariant will apply to any `Ptr` because it is "Generic" that is the type arguments of the data con are distinct type variables. So, the first will apply to `Ptr b` too (and every Ptr)

gridaphobe: Oh, I see. I missed the comment above. Excellent!

nikivazou: Just added the comment:)

PR #689, opened 2016-04-23
Measure invariants

nikivazou:
Turn measure post condition into invariants

TODO: check that measure types do not have any pre conditions

### DIFF
--- a/src/Language/Haskell/Liquid/Bare/Check.hs
+++ b/src/Language/Haskell/Liquid/Bare/Check.hs
@@ -138,8 +138,8 @@ firstDuplicate = go . L.sort
                 | otherwise = go (x:xs)
     go _                    = Nothing
 
-checkInv :: Bool -> TCEmb TyCon -> TCEnv -> SEnv SortedReft -> Located SpecType -> Maybe Error
-checkInv allowHO emb tcEnv env t   = checkTy allowHO err emb tcEnv env t
+checkInv :: Bool -> TCEmb TyCon -> TCEnv -> SEnv SortedReft -> (Maybe Var, Located SpecType) -> Maybe Error
+checkInv allowHO emb tcEnv env (_, t)   = checkTy allowHO err emb tcEnv env t
   where
     err              = ErrInvt (sourcePosSrcSpan $ loc t) (val t)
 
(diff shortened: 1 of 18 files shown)


ranjitjhala:
Is the post condition (output refinement) being inferred in this example?

ranjitjhala:
Also: SUPER!!!

sorawee:
> Is the post condition (output refinement) being inferred in this example?

I think it's not. I run the code in my local machine (which is an old version), and it says SAFE.

A code that demonstrates that shouldn't have `{-@ data F [lenF] @-}`.

gridaphobe:
Nice! I'm also curious about whether the post-condition is being inferred.

Also, does this work for old-style measures too? IIRC we already support refinements in the measure syntax, so it shouldn't be too hard to add.

nikivazou:
@sorawee are you suggesting that the user should not write `{-@ data F [lenF] @-}.`? To me it seems ok to ask the user to define how to size a user-defined data type, i.e., it is ok to ask the user to write `{-@ data F [lenF] @-}.`, at least for now. 

The post condition is not inferred, but it could be!
So, it should be, if there is a user-defined post pondition, use that otherwise create a k_var.

sorawee:
Oh no. I think we misunderstood each other.

What I wrote is, I think that that test that you added (`tests/pos/AutoTerm.hs`) does not actually test the new feature that you added in this PR. We can see this by copying this test to goto.ucsd. When hitting the check button, LH says that it's SAFE. However, if this is truly a new feature, then goto.ucsd which doesn't support this feature yet should result in UNSAFE.

ranjitjhala:
That is @nikivazou it doesn't seem there is any post condition on the measure in this test -- it seems commented out?

nikivazou:
I made many fixes: 

@sorawee 's example [tests/pos/AutoTerm.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/measure_invariants/tests/pos/AutoTerm.hs) is now safe in this PR and UNSAFE in the demo. 

Turned out to be complicated to implement, because when we check the body of the measure, we cannot assume the invariant, but we need to assume it in the recursive call (i.e., for all inputs less than the arguments of the measure.)


Also, in GhcListSort, we produce 2 sigs for the local binder `descending` and `ascending` with incompatible haskell types (these types cannot meet). Any idea why?

ranjitjhala:
Great!

> Also, in GhcListSort, we produce 2 sigs for the local binder descending
 and ascending with incompatible haskell types (these types cannot meet).
Any idea why?

Not sure what you mean...?

PR #692, opened 2016-04-26
smarter monomorphization in model-generation

gridaphobe:
- use class constraints to drive monomorphization
- default unconstrained `(t :: *)` to `Int`
- default unconstrained `(t :: * -> *)` to `[]`


nikivazou:
@gridaphobe, what does this PL address?

gridaphobe:
It makes the counter-example generation work better for polymorphic types. 

We have to monomorphize the types in order to produce values, and the scheme was originally very naive (i.e. make everything an Int). The new scheme takes class constraints and kinds into account, meaning it's more likely to find a valid specialization to use to generate a model.

PR #695, opened 2016-04-28
Add Typing rule for `>>=`

gridaphobe:
to avoid creating a kvar for the lambda


ranjitjhala:
Oops I am working on this too...

PR #696, opened 2016-04-28
Add Typing rule for `>>=`

gridaphobe:
to avoid creating a kvar for the lambda


ranjitjhala:
Do the tests pass? (My version is similar but is failing tests due to some odd shadowed binder issue)

gridaphobe:
I had a similar shadowing issue, it seemed to be due to calling
'normalize' on the lambda body rather than 'stitch'.
 
It looks like there are still a few failing tests, but I haven't looked
at them yet.

ranjitjhala:
Yes, thought that might be it.

The icfp15 monad tests should fail with this ...

ranjitjhala:
*Don't merge* -- I have taken the relevant parts into the `patterns` branch which also has some other stuff that I will send a PR for in due course.

PR #698, opened 2016-04-28
ApiAnnotations

### LINKED ISSUE #617: Ignore commented out annotations
```
-- {-@ foo :: () @-}
-- foo = ()
```
and
```
{-
{-@ foo :: () @-}
foo = ()
-}
```

Both cause LH to complain about `foo` being undefined.

spinda:
This (a) reworks the GHC Interface pipeline to handle multiple target files at once and parse/load each module one-by-one with a dependency graph, and (b) uses that infrastructure to pull comments out of GHC's ApiAnnotations interface instead of using our own parsing for that (fixes #617).

(a) lays the groundwork for the addition of quasiquoters (coming very soon!), .lqhi files, and the GHC plugin interface. It also means LH compiles each module exactly once, which helps a lot with `liquidhaskell-cabal`.

One caveat: this isn't compatible with using `--idirs` inside `{-@ LIQUID @-}` file pragmas. I think the test suite is the only user of this functionality; to make it work, I moved those flags into an `extraOptions` map in the test runner itself.

### DIFF
--- a/src/Language/Haskell/Liquid/GHC/Interface.hs
+++ b/src/Language/Haskell/Liquid/GHC/Interface.hs
@@ -236,7 +236,7 @@ processModule cfg logicMap tgtFiles depGraph specEnv modSummary = do
   let mod              = ms_mod modSummary
   _                   <- liftIO $ whenLoud $ putStrLn $ "Module: " ++ showPpr (moduleName mod)
   file                <- liftIO $ canonicalizePath $ modSummaryHsFile modSummary
-  _                   <- load $ LoadDependenciesOf $ moduleName mod
+  _                   <- loadDependenciesOf $ moduleName mod
   parsed              <- parseModule $ keepRawTokenStream modSummary
   typechecked         <- typecheckModule $ ignoreInline parsed
   _                   <- loadModule typechecked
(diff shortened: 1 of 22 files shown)


gridaphobe:
If `--idirs` doesn't work anymore in a LIQUID pragma (fine by me), do we throw an error warning the user?

spinda:
Good point! We do now.

src/Language/Haskell/Liquid/GHC/Interface.hs
```diff
@@ -72,284 +79,339 @@ import Language.Fixpoint.Utils.Files
...
+getSpecComments :: ParsedModule -> [(SourcePos, String)]
+getSpecComments parsed = mapMaybe getSpecComment comments
+  where
+    comments = concat $ M.elems $ snd $ pm_annotations parsed
 
-exprDep :: CoreExpr -> [Id]
-exprDep = freeVars S.empty
+getSpecComment :: GHC.Located AnnotationComment -> Maybe (SourcePos, String)
+getSpecComment (GHC.L span (AnnBlockComment text))
+  | length text > 2 && isPrefixOf "{-@" text && isSuffixOf "@-}" text =
```
gridaphobe: Do we actually need the `length > 2` test?

ranjitjhala: Seems redundant given the other checks ...

gridaphobe: Indeed, e.g.

```
ghci> "{-@" `isPrefixOf` ""
False
```

spinda: Oops, right. That's left over from a previous version where that code used `take` and `drop`.

ranjitjhala:
Very exciting! thanks @spinda !!!

PR #700, opened 2016-04-29
ackermann

nikivazou:
More edits to enable pretty proofs of functional properties


gridaphobe:
I don't think we need a `--scrape-assumed` flag, it should just be done always.

nikivazou:
I deactivated the always: we just scrape the quals from assumed signatures when the flag is on. You think it is a bad idea?

gridaphobe:
No, I think we should always be scraping quals from assumed **and** asserted signatures. (Unless there are examples where it kills performance)

nikivazou:
I am auto-generating some assumed sigs and I really want to ignore the qualifiers from these auto generated signatures. 
The ignore-assumed was the easy solution. I will mark these signs as internals instead and ignore the internal sigs quals

gridaphobe:
Ah, I see. Ignoring generated sigs seems fine, but we definitely shouldn't ignore user-written sigs :)

PR #702, opened 2016-05-05
Patterns

ranjitjhala:
Add a `--pattern-inline` flag that special cases the handling of `e1 >>= \x -> e2` to avoid KVars. 

Would be nice to generalize / make this user extensible...


gridaphobe:
Indeed, user-extensible typing rules would be quite neat!

PR #705, opened 2016-05-06
Map fusion

nikivazou:
Proving map fusion


src/Language/Haskell/Liquid/Constraint/Generate.hs
```diff
@@ -1551,6 +1551,7 @@ funExpr _ _
 
 untick :: CoreExpr -> CoreExpr
 untick (Tick _ e)  = untick e 
+untick (App e (Type _)) = untick e
```
gridaphobe: This looks bogus, why are you removing the `Type` argument?

nikivazou: Let me rename: I want to get 

compose @a @b @c f g :: {v:_ | v == compose f g}

nikivazou:
Any idea how one can write the natural map function and get the following logic-friendly version?

```
map f xs 
  | nill xs   = N 
  | otherwise = C (f (hd xs)) (map f (tl xs)) 
```

nikivazou:
@ranjitjhala here are the summaries:

[develop.csv](https://github.com/ucsd-progsys/liquidhaskell/blob/08d9fa6923e66df9d1e01dcece0fb05875c0f876/tests/log/summary-develop.csv) VS [map_fusion.csv](https://github.com/ucsd-progsys/liquidhaskell/blob/08d9fa6923e66df9d1e01dcece0fb05875c0f876/tests/log/summary-map_fusion.csv)

There is a small slowdown, I assume because I am convertion logic in a curry form. Can I merge?

ranjitjhala:
Can't the conversion be guarded by the allowHO flag?

nikivazou:
It could, but better if the whole defunctionalization is guarded by the allowHO flag. 

Since the slowdown is not much, can I merge now and continue with the monadic laws and guard the defunctionalization step later?

ranjitjhala:
OK let me look at the tables and get back to you in like 20 minutes

PR #706, opened 2016-05-07
Fixes to ByteString specs

thoughtpolice:
The first one is obvious, but I'm particularly unsure about 31733db.

This was found in the process of using `liquidhaskell` on `binary-serialise-cbor`.

### DIFF
--- a/include/Data/ByteString.spec
+++ b/include/Data/ByteString.spec
@@ -34,10 +34,6 @@ append
 
 head :: { bs : Data.ByteString.ByteString | 1 <= bslen bs } -> Data.Word.Word8
 
-uncons
-    :: i : Data.ByteString.ByteString
-    -> Maybe (Data.Word.Word8, { o : Data.ByteString.ByteString | bslen o == bslen i - 1 })
-
 unsnoc
     :: i : Data.ByteString.ByteString
     -> Maybe ({ o : Data.ByteString.ByteString | bslen o == bslen i - 1 }, Data.Word.Word8)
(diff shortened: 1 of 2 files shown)


ranjitjhala:
Thanks! Both of these are good (and mirror what we verified of an older
version of bytestring)

+ [uncons](
https://github.com/ucsd-progsys/liquidhaskell/blob/master/benchmarks/bytestring-0.9.2.1/Data/ByteString.hs#L613
)

+ [unsafeTail](
https://github.com/ucsd-progsys/liquidhaskell/blob/master/benchmarks/bytestring-0.9.2.1/Data/ByteString/Unsafe.hs#L119
)

Looking at the above, we can strengthen the specification further to say
the
output is one smaller than the input...

{-@ unsafeTail :: b:{v:ByteString | bslen v > 0} -> {v:ByteString | bslen v
== bslen b - 1} @-}


One of these days, will port those signatures from benchmarks over to the
.spec ...

Did you want to update the PR with the stronger `unsafeTail` ? Will then
merge in?

Thanks a bunch!

thoughtpolice:
@ranjitjhala Done.

Can I ask what you think of about 31733db39f58e3bd88f7840cc2b45d907c8c2c94, which is completely inexplicable to me? If you merge it, then the specs will start working, but it's completely unclear what's going on.

In any case, it would be good to maybe at least boil a small reproducible test case out of it, if possible, or at least open another issue before closing this and forgetting about it, since it clearly seems to be some nasty bug...

gridaphobe:
Ugh, that is indeed a nasty parser bug (that I thought I had squashed..)

The error should also go away if you parenthesize the whole output type, i.e. `(Maybe ...)`.

Thanks for pointing it out!

gridaphobe:
The parsing issue is now tracked in #707. Thanks @thoughtpolice!

PR #713, opened 2016-05-16
allow refinements with out haskell types

nikivazou:


ranjitjhala:
No parse errors?! :) (I hope you didn't use `try` ...)

nikivazou:
I did, because backtracking is needed! But @gridaphobe said we used `try` correctly

gridaphobe:
Yes, we need `try` to distinguish

```
{ v:t | p }
```

from

```
{ p }
```

because `{ v` could be interpreted as a prefix of either.

PR #715, opened 2016-05-18
Liquid prover

nikivazou:


benchmarks/haskell16/pos/Overview.hs
```diff
@@ -64,8 +68,9 @@ fib_incr :: Int -> Proof
 {-@ fib_incr :: n:Nat -> {fib n <= fib (n+1)} @-}
 fib_incr n
    | n == 0
-   = toProof $
-      fib 0 <! fib 1
+   = fib 0 <! fib 1
+   ! QED
+
```
nikivazou: @ranjitjhala what do you think of this `QED` syntax?

ranjitjhala: nice! But too many `!` -- can you pick something else e.g. `*** QED` ?

nikivazou: Here https://github.com/ucsd-progsys/liquidhaskell/commit/b12adf7db481a4270bb1dc90dea1a342f1810c26#diff-c7b7719ee938950f3335b8bd0e6ef0deR69


I prefer the `!` because it is more uniform: most library operators have `!`

Both cases are better than toProof though

Issue #483, opened 2015-10-07
DsExpr crashes on Template Haskell


nikivazou:
If I run liquid on a file that contains some template haskell, for example in [TemplateHaskell.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/e9a2d9932da1efd11d9f6c1b0f2daeb17fa87732/tests/todo/TemplateHaskell.hs)
 it will crash with `dsExpr HsBracketOut`. 
Specifically, it will hit this panic: [DsExpr.lhs](https://github.com/ucsd-progsys/liquidhaskell/blob/master/src/Language/Haskell/Liquid/Desugar/DsExpr.lhs#L569).

Note that this panic message exists in the original desugarer code too.

gridaphobe:
This is a longstanding problem for us. You can try compiling liquid with `-DGHCI` to trigger the other branch, but IIRC that causes other things to crash instead.

PR #721, opened 2016-06-02
Fix dsBracket error with Template Haskell

spinda:
See https://github.com/ucsd-progsys/liquidhaskell/issues/483


nikivazou:
Thanks @spinda !

PR #722, opened 2016-06-02
Fix unboxed tuples

### LINKED ISSUE #718: Unboxed Tuples
Unboxed Tuples crash LH at the moment.

http://goto.ucsd.edu:8090/index.html#?demo=permalink%2F1464133143_12050.hs

```
LiquidHaskell Copyright 2009-15 Regents of the University of California. All Rights Reserved.

[1;94m
**** DONE:  Parsed All Specifications ******************************************
[0m[1;90m 
[0mliquid: Error: bytecode compiler can't handle unboxed tuples.
  Possibly due to foreign import/export decls in source.
  Workaround: use -fobject-code, or compile this module to .o separately.
```

This seems to be a GHC error that arises because we set `HscInterpreted` in

https://github.com/ucsd-progsys/liquidhaskell/blob/master/src/Language/Haskell/Liquid/GHC/Interface.hs#L123-L126

The FIXME says we need this in order to lookup certain unexported Names in modules.

1. Do we still need this in light of @spinda's work on `Bare`?
2. If not, can we use `-fobject-code` instead to support unboxed tuples?

spinda:
This completely fixes #718 by switching from `HscInterpreted`/`LinkInMemory` to `HscNothing`/`NoLink` when we go to call `loadModule`.

A previous attempt at a fix was to use `HscNothing`/`NoLink` everywhere, unless a module needed Template Haskell, in which case `HscInterpreted`/`LinkInMemory` were enabled. This meant that both Template Haskell *and* unboxed tuples could not be used in the same project; also, as it turns out, we depend on the behavior of GHC under `HscInterpreted`/`LinkInMemory` in a number of places. This new approach fixes the original issue and these additional problems.


gridaphobe:
@spinda can this be merged? (barring the conflict)

ranjitjhala:
(assuming @spinda busy with final -- I took a crack at it, but left the "conflicting" code in just in case)

ranjitjhala:
Doesn't look like this fixes #718 ?

ranjitjhala:
I take that back, seems to work now with `loadModule'` instead.

spinda:
Ah, sorry! Will go back and remove those conflicts. Thanks.

ranjitjhala:
I think I dealt with it and fixed it a while back... no worries!

Issue #743, opened 2016-06-21
Deriving Read instance causes unsoundness


nikivazou:
The [following example](https://github.com/ucsd-progsys/liquidhaskell/blob/8fe9dfb0f1e194f51809ee428113816dd9b83101/tests/todo/DerivingRead.hs) is SAFE while should be unsafe. 

```
{-@ checkNat :: Nat -> Int @-} 
checkNat :: Int -> Int 
checkNat x = x

unsound :: Int
unsound = checkNat (-1)


data TestBS = TestBS Int deriving (Read)
```
Current workarounds
 -  add `--trust-internals` flag to ignore the deriving instances
 -  delete the `deriving (Read)` instance

ranjitjhala:
This is a long standing problem, do you have any idea what the cause is? Does the read instance create some top-level literals?

ranjitjhala:
@nikivazou why don't we just make `--trust-internals` the default behavior and *not* an option ?

nikivazou:
I did not look more into it. Just minimized the code and filed the issue. It should be a top level binder. 
Yes lets make trust internals default! 

Also, suggesting to make the totality default!

ranjitjhala:
I looked at the output core, and basically there are *many* little top-level binders added which are not "called" and so probably get the type "False" (as they are not explicitly exported...)

ranjitjhala:
totality default...? hmm...seems brave... :)

nikivazou:
I am exporting everything, so even if they are not called they should be refined to true! 

I think totality is more common and easier to prove than termination. And totality and termination are nice default checks: if your code passes LiquidHaskell then it is total!

ranjitjhala:
ok, will do.

nikivazou:
Or we can do it after the deadline:)

ranjitjhala:
Hopefully fixed by https://github.com/ucsd-progsys/liquidhaskell/pull/749

ranjitjhala:
Unfortunately, the above doesn't work, as it breaks all our class measure stuff (all those *also* generate binders that look like "GHC internals" or are "derived" names...)

In the case of `Read` there are some odd recursive definitions, namely

`Bob.$fReadTestBS` depends on `$creadListPrec` which itself depends on `Bob.$fReadTestBS`,
yielding the `False` . I'm guessing _something_ here needs to be contra-variant?


```
 Bob.$fReadTestBS =
[...]

PR #749, opened 2016-06-24
Fix743

ranjitjhala:
Fix for issue #743.

We **cannot** make `--trust-internals` default because it means skipping all instance methods, which breaks the whole class measure thing.

Instead, this fix ensures that all derived/generated binders have "True" as their "top-level" refinement,
i.e. do not contaminate the top-level environment with some bogus inferred refinement. (This is needed
for things like, e.g. `Read` where there is some oddness involving the generated binders and dictionaries, and which, since they are auto-generated, cannot be annotated as `Lazy` by the programmer.)

### DIFF
--- a/src/Language/Haskell/Liquid/Constraint/Generate.hs
+++ b/src/Language/Haskell/Liquid/Constraint/Generate.hs
@@ -119,10 +119,10 @@ consAct :: GhcInfo -> CG ()
 consAct info = do
   γ'    <- initEnv      info
   sflag <- scheck   <$> get
-  -- tflag <- trustghc <$> get
+  tflag <- trustghc <$> get
   γ     <- if expandProofsMode then addCombine τProof γ' else return γ'
   cbs'  <- if expandProofsMode then mapM (expandProofs info (mkSigs γ)) $ cbs info else return $ cbs info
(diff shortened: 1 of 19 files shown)


ranjitjhala:
Tests pass locally, will merge once CI says ok.

src/Language/Haskell/Liquid/Constraint/Generate.hs
```diff
@@ -628,51 +622,49 @@ consCBLet :: CGEnv -> CoreBind -> CG CGEnv
...
-       let (γ',i) = removeInvariant γ cb
-       γ'' <- consCB (tflag && isStr) isStr γ' cb
+       let (γ', i) = removeInvariant γ cb                 --- DIFF
+       γ'' <- consCB (oldtcheck && isStr) isStr γ' cb
        modify $ \s -> s { tcheck = oldtcheck}
-       return $ restoreInvariant γ'' i
+       return $ restoreInvariant γ'' i                    --- DIFF
 
+-- HEREHEREHEREHERE
+-- why is consCBTop ALMOST equivalent to consCBLet ? except for this removeInvariant stuff?
```
nikivazou: remove invariant removes the measure invariant (post-condition) when the measure is checked

nikivazou:
Thanks Ranjit! Looks good!

PR #752, opened 2016-06-28
Bug in simplifyPatTuple

nikivazou:
tests/pos/SimplifyTupleFailure.hs crashed because the simplifyTuple rule changed the type of an expression. A simple solution: only apply the rule when the type is preserved. 

WHY changed the type? The type changes in the body of `isSrcTCP` because the first case becomes top level and `exprType` just returns the type argument of the case expression without checking the body of the expression.


nikivazou:
@ranjitjhala why didn't the `--no-pattern-inline` flag deactivate this transformation?

ranjitjhala:
Please don't merge this till I have looked at it; the pattern stuff is very important for eliminate...

ranjitjhala:
Not sure that this is a fix because the type is supposed to change… But let
me take a closer look

ranjitjhala:
Will take a look soon, but if you look at the `Rewrite.hs` you will see that:

1. There is some NOTE about types changing, so it _is_ supposed to handle that,
2. You can disable it with `--no-simplify-core`

ranjitjhala:
Also, I cannot fully understand your comment: "WHY changed the type?..." since you've worked this out, can you explain what the (GOOD) source expression is, and what the (BAD) rewritten target expression is? Thanks!

ranjitjhala:
Looks like the new test case still fails? Perhaps its missing a commit?

nikivazou:
Sorry, I pushed my extra check commit. Updating now.

nikivazou:
The problem is that an expressions 

```
E1 = let x1 = case x _ te of {...} in e 
``` 
is rewritten by making the case expression top level 
```
E2 = case x' _ te of {...}
```

now `CoreUtils.exprType` in case expressions works just by grapping the case type, thus
```
CoreUtils.exprType E2 ==  CoreUtils.exprType (case x' _ t2 of {....}) == t2 
                                      /= CoreUtils.exprType e == "realtype of E1"
```

Since we are using `CoreUtils.exprType` to create fresh true types during generation this inconsistency crates malformed subtyping constraints. 

The real fix should be to update the type of the case when rewriting `E1` to `E2` (does this make sense?), but not sure where this should happen.

How can the `simplifyPatTuple :: RewriteRule` not preserve types? It is implied by its type!

ranjitjhala:
Specifically, looks like

```haskell
let {
  ds_dBV :: (FOO.Port, FOO.Port)
  [LclId, Str=DmdType]
  ds_dBV =
    case FOO.idTCP x of _ [Occ=Dead] { FOO.PP ds_dBQ ds_dBR ->
    case ds_dBQ of _ [Occ=Dead] { FOO.PortId srcIP _ [Occ=Dead] ->
    case ds_dBR of _ [Occ=Dead] { FOO.PortId srcPort _ [Occ=Dead] ->
    (srcIP, srcPort)
    }
    }
    } } in
let {
  srcPort :: FOO.Port
  [LclId, Str=DmdType]
  srcPort =
    case ds_dBV of _ [Occ=Dead] { (_ [Occ=Dead], srcPort) ->
    srcPort
    } } in
let {
  srcIP :: FOO.Port
  [LclId, Str=DmdType]
  srcIP =
    case ds_dBV of _ [Occ=Dead] { (srcIP, _ [Occ=Dead]) -> srcIP } } in
(GHC.Classes.== @ (FOO.Port, FOO.Port) $dEq_aBu)
  (FOO.addrTE x, FOO.portTE x) (srcIP, srcPort)
```

is rewritten to

```
case FOO.idTCP x of _ [Occ=Dead] { FOO.PP ds_dBQ ds_dBR ->
case ds_dBQ of _ [Occ=Dead] { FOO.PortId srcIP _ [Occ=Dead] ->
case ds_dBR of _ [Occ=Dead] { FOO.PortId srcPort _ [Occ=Dead] ->
(GHC.Classes.== @ (FOO.Port, FOO.Port) $dEq_aBu)
  (FOO.addrTE x, FOO.portTE x) (srcIP, srcPort)
}
}
```

which to the naked eye looks fine, but you are saying that the type 
of the top-level `case-of`is a tuple instead of a `Bool` ?

src/Language/Haskell/Liquid/Transforms/Rewrite.hs
```diff
@@ -150,8 +150,8 @@ rewriteWith tx           = go
 
 -}
 
-_safeSimplifyPatTuple :: RewriteRule
-_safeSimplifyPatTuple e 
+safeSimplifyPatTuple :: RewriteRule
+safeSimplifyPatTuple e 
   = find ((CoreUtils.exprType e ==) . CoreUtils.exprType) (simplifyPatTuple e)
```
ranjitjhala: this line incomprehensible, i'm afraid.

src/Language/Haskell/Liquid/Transforms/Rewrite.hs
```diff
@@ -149,6 +150,10 @@ rewriteWith tx           = go
 
 -}
 
+safeSimplifyPatTuple :: RewriteRule
```
ranjitjhala: I think this is a lot clearer, if more verbose:

```haskell

safeSimplifyPatTuple :: RewriteRule
safeSimplifyPatTuple e
  | Just e' <- simplifyPatTuple e
  , CoreUtils.exprType e' == CoreUtils.exprType e
  = Just e'
  | otherwise
  = Nothing
```

src/Language/Haskell/Liquid/Transforms/Rewrite.hs
nikivazou: It is too haskellish, using find as a Foldable 

```
find :: (a -> Bool) -> Maybe a -> Maybe a
```

nikivazou:
You should update the type in the case expressions because exprType returns wrong type. 

https://downloads.haskell.org/~ghc/7.0.2/docs/html/libraries/ghc-7.0.2/src/CoreUtils.html#exprType

exprType (Case _ _ ty _)     = ty

ranjitjhala:
That sounds right...

ranjitjhala:
Ugh, occurs to me I should have just pulled from this branch or some such... oh well, sorry about that but thanks for digging this bug up!

PR #754, opened 2016-06-28
Fix p752

ranjitjhala:
Fixes the issue found and debugged by @nikivazou in https://github.com/ucsd-progsys/liquidhaskell/pull/752

See that for discussion and fix... Will merge if / when CI blesses it...


src/Language/Haskell/Liquid/Transforms/Rewrite.hs
```diff
@@ -151,8 +151,26 @@ rewriteWith tx           = go
 
 --------------------------------------------------------------------------------
 simplifyPatTuple :: RewriteRule
+simplifyPatTuple e =
+  case simplifyPatTuple' e of
+    Just e' -> if CoreUtils.exprType e == CoreUtils.exprType e'
+                 then Just e'
+                 else Just (tracePpr ("YIKES: RWR " ++ showPpr e) e')
```
nikivazou: Can you delete this tracing?

nikivazou:
Thanks @ranjitjhala !

PR #766, opened 2016-07-11
Check unbound variables in user specified signatures 

nikivazou:
The following used to be safe (not crashes).

```
data MI s  = Small { mi_input :: String  }
{-@ Small :: forall s. {v:String | s == v } -> MI s @-}
```

http://goto.ucsd.edu:8090/index.html#?demo=permalink%2F1468258362_14141.hs


nikivazou:
Hmmm, I thought the fix would be easy, but seems that it breaks all the abstract refinements test. 
Closing this PR and will create an issue instead.

PR #769, opened 2016-07-12
Reasoning about (Iavor's) type literals 

nikivazou:
`s::Symbol` is treated like a value `s:String` and 
`n::Nat` is treated like a value `n::Integer`. 

See [tests/neg/DependentTypes.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/d577423527694e14f253d6f18d0c5eca857ce7a2/tests/neg/DependentTypes.hs) and [tests/pos/DependentTypes.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/5e74cacab1755a4e2623bab2d669abccd037ff5e/tests/pos/DependentTypes.hs).


nikivazou:
Ready to merge if all tests pass

ranjitjhala:
let me merge it and use it as a test for the test-log-script.

PR #783, opened 2016-07-15
Cases to logic

nikivazou:


ranjitjhala:
Is this ready to merge?

nikivazou:
Yes, it is "minor" fixes on how we translate haskell to logic

PR #790, opened 2016-07-21
More vector assumptions

andrewthad:
This time against develop instead of master. Still WIP.


nikivazou:
@andrewthad  can you close this or rebuild it against the current develop? 
I don't see any reason why not merge.

PR #800, opened 2016-07-27
Refactor L.H.L.Desugar into a separate library

christetreault:
This PR will update Liquidhaskell to use a new liquiddesugar library which is currently located at https://github.com/christetreault/liquiddesugar.git

As it stands, the git submoduleing is all set up to use the head of the liquiddesugar repository, and this works as it did with GHC 7.10. In the future, this will allow us to maintain multiple liquiddesugar versions, where liquiddesugar A.B is compatible with GHC A.B. This will greatly reduce the level of effort of supporting new GHC versions as:

 - The liquidhaskell port itself will be greatly simplified as no part of liquidhaskell is essentially a fork of GHC any more
 - We no longer need to duplicate the desugar module for GHC versions or have thousands of lines of `#ifdef` guards. We can instead tear out the old desugar, and put in a new one and let stack/cabal sort it out.

For this all to work, we'll need to get liquiddesugar released to hackage before liquiddesugar 8.0 can happen.


ranjitjhala:
Sweet! Can you make this happen (assuming tests pass etc) undoubtedly there
will be a few bumps as with any submodule twiddling but the sooner we do it
the better!

Issue #806, opened 2016-08-06
Unable to build on Windows due to 'unix' package dependency


alclub:
> setup.exe: Package unix-2.7.1.0 can't be built on this system.

I get a different but similar message on Cygwin. Is there any way to get liquidhaskell running on Windows, or is Windows no longer supported?

gridaphobe:
Are you building from github? If so, can you try commenting out the `lhi` stanza in the cabal file?

https://github.com/ucsd-progsys/liquidhaskell/blob/master/liquidhaskell.cabal#L79-L104

The main `liquid` binary *should* build on windows, if it doesn't that's a bug!

gridaphobe:
@ranjitjhala do you absolutely need `unix` for `lhi`? I would think there's a way to do the daemonized server on windows too.

ranjitjhala:
Let's comment it out for now since its super dormant and will remain so for
several weeks at least...

ranjitjhala:
Eric can you do it? Am rather laptop challenged atm -- my ssd crashed :(

We should ensure develop builds on windows always...

alclub:
> Are you building from github? If so, can you try commenting out the lhi stanza in the cabal file?

Thanks for your assistance. Yes, I am building from github (using stack). I removed the 'lhi' block, now it quits with the error:

`<command line>: can't load .so/.DLL for: stdc++.dll (addDLL: could not load DLL)
    ghc.exe: stdc++: The specified module could not be found.`

There doesn't appear to be much information online about this issue except that references to "double-conversion" are causing some packages to fail to build on windows.

I've attempted [the fixes given here](https://github.com/commercialhaskell/stack/issues/754#issuecomment-165457815): 

but now I get the error:

` [ 38 of 105] Compiling Language.Haskell.Liquid.Types ( src\Language\Haskell\Liquid\Types.hs, .stack-work\dist\2672c1f3\build\Language\Haskell\Liquid\Types.o )
    ghc.exe: unable to load package 'double-conversion-2.0.1.0'
    ghc.exe: C:\x\xx\liquidhaskell\.stack-work\install\12018b8a\lib\x86_64-windows-ghc-7.10.3\double-conversion-2.0.1.0-9d4SFE0Uwt89cnl39kvRqX\HSdouble-conversion-2.0.1.0-9d4SFE0Uwt89cnl39kvRqX.o: unknown symbol '_Unwind_Resume'`

and also

[...]

PR #807, opened 2016-08-06
disable lhi for now

gridaphobe:
it's incomplete and breaks windows support by using `unix`.

partial fix for #806


nikivazou:
@gridaphobe I defined `mapFst` and `mapSnd` in fixpoint too, thus your PR fails. Can you merge/update? Sorry :(

gridaphobe:
@nikivazou I don't understand, this patch has nothing to do with `mapFst`.

gridaphobe:
ugh, maybe it's because i opened the PR against master..

PR #809, opened 2016-08-09
Fix typos in README

stefkin:


nikivazou:
Thanks so much @stefkin! Appreciate it! 
Can you please update your target branch to be `ucsd-progsys:develop` because for stability reasons we have made merging to master difficult:)

ranjitjhala:
Thanks a lot for this!

Btw @nikivazou or @gridaphobe is there a way to make develop the default
for PR? Or at least maybe we should create a small CONTRIBUTING.md that
says as much?

nikivazou:
Good point @ranjitjhala! I made develop the default branch

stefkin:
re-created on develop branch in #811

PR #811, opened 2016-08-10
Fix typos in README and add contributing guide

stefkin:
I also added a small contributing guide as @ranjitjhala suggested in #809.


nikivazou:
Thanks @stefkin!

PR #812, opened 2016-08-10
Start monoid laws

nikivazou:


ranjitjhala:
Wow that looks like some pretty fancy stuff... Nice!

nikivazou:
Thanks @ranjitjhala! It turned out more sophisticated than initially expected

Issue #832, opened 2016-09-08
Untrusted-code checking: needs `--termination-check` command line argument


rrnewton:
We're thinking about untrusted code execution a bit.  Currently, if we, the server, get a single file from the untrusted user, it may contain `{-@ LIQUID @-}` pragmas, including `--no-termination-check`.  

We could try to *strip* these ourselves before compiling the file.  But a more straightforward approach is to just make sure that command line arguments override in-file pragmas.  We already have a `--no-termination-check` command line flag, but we need its opposite, `--termination-check`.

Finally, note that currently a server needs to check every file separately, as checking a file with module imports does not imply that its imports have been been checked for nontermination (though it does scrape and import the specs).

rrnewton:
Also, note that this is only for preventing nontermination-cheats within the submitted code.  To block cheats from base itself depends on #815.

At the same time, we also depend on Safe Haskell plus a suitable trust list to block other problems.

nikivazou:
So, how about adding a `--safeLiquidHaskell` argument that for now just enforces termination check?

rrnewton:
That would be great.  Would that be an alias for multiple flags?  If not, how about just calling it `--termination-check` for symmetry with `--no-termination-check`?

nikivazou:
@rrnewton does #835 address what you had in mind?

rrnewton:
Yep, sure, as long as `--totality` overrides a `{-@ LIQUID --no-termination-check @-}` in the file, it suffices to address this issue. I'll close this one in favor of that.

PR #835, opened 2016-09-26
Total haskell flag

nikivazou:
Adding "--total-Haskell" flag that checks termination and totality (overriding potential "no-termination" flag). 

For issue #832. 

See [tests/neg/TotalHaskell.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/a10ed5beef30f0151fc56a97b14532d10875b55c/tests/neg/TotalHaskell.hs)


gridaphobe:
Let's not add yet another option for existing functionality. Rather, why not have `--totality` imply `--termination` (i.e. override `--no-termination`)? A total function ought to be terminating after all :)

nikivazou:
This is an interesting observation, but I am not entirely sure what is total in Haskell's semantics. Remember that our termination is under eager evaluation. I bet `repeat x = x : repeat x` is a total function....

gridaphobe:
True, but I find your argument a bit unconvincing as one might reasonably believe that `--termination` should accept `repeat`; it always produces something in WHNF.

I think that having `--totality` use eager semantics is no worse than having `--termination` do so. It just needs to be documented clearly (e.g. `--totality == --termination + no pattern errors etc.`).

Issue #819, opened 2016-08-17
Infix operators in type signatures


vikraman:
While it is possible to axiomatize an infix function, using it in a type signature results in a parse error.

``` haskell
{-@ axiomatize .+. @-}
(.+.) :: N -> N -> N
Zero .+. n = n
(Suc m) .+. n = Suc (m .+. n)

{-@ lemma :: { v:() | Zero .+. Zero == Zero } @-}
lemma :: ()
lemma = Zero .+. Zero ==. Zero *** QED
```

```
    84  28 error           Error: Cannot parse specification:

 84 | {-@ lemma :: { v:() | Zero .+. Zero == Zero } @-}
                                 ^

     unexpected "."
     expecting operator, "==", "=", "~~", "!=", "/=", "!~", "<", "<=", ">", ">=" or "(" (haskell-liquid)

```

nikivazou:
Parsing with infix operators is tricky and we currently do not support it. 

I acknowledge that this issue is annoying, but currently now high in my TODO list, sorry. 
@spinda was willing to fix it, but @vikraman if you are willing to give it a try here is the refinement parser 

https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/src/Language/Fixpoint/Parse.hs#L276

vikraman:
Ouch, this also causes tuples to not work.

nikivazou:
Yeap, in logic we only support prefix functions. 

Lists, tuples, or anything infix is not properly parsed into logic.

ranjitjhala:
Also: #819, #843, #902 

@nikivazou this is a really annoying limitation, because you cannot
even  reflect any operator. For example in the below, LH _never_ has 
to actually parse an infix operator, so it _should_ be easy to replace `mappend`
with `++` and yet we cannot (I get some weird error.) At the very least stuff like 
the below where the LH parser doesn't change should be easy to fix?

```haskell
data List a = Emp | a ::: List a  deriving (Eq)

{-@ reflect append @-}
append Emp        ys = ys
append (x ::: xs) ys = x ::: (xs `append` ys)

{-@ inline assocThm @-}
assocThm xs ys zs
  = (xs `append` ys) `append` zs == xs `append` (ys `append` zs)

{-@ assocPf :: xs:_ -> ys:_ -> zs:_ -> { assocThm xs ys zs } @-}
assocPf Emp ys zs
  =   (Emp `append` ys) `append` zs
[...]

PR #843, opened 2016-10-05
Infix Operators in the logic

nikivazou:
Infix operators can be defined in the logic using a Haskell like notation

     {-@ infixl 9 <> @-}

Addresses issue #819

### DIFF
--- a/README.md
+++ b/README.md
@@ -800,6 +800,17 @@ see [tests/pos/Map.hs](tests/pos/Map.hs)
 2. Value parameters are specified in **upper**case: `X`, `Y`, `Z` etc.
 
 
+Infix Logic Operators
+---------------------
+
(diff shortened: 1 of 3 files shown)


ranjitjhala:
I don't believe this :) (bravo!!!)

nikivazou:
And it does not break things:) 

May I merge?

ranjitjhala:
(Assuming all tests pass) yes

gridaphobe:
Is it possible to seed the fixity table with the fixities that GHC already knows about?

nikivazou:
I used this 
https://www.haskell.org/onlinereport/decls.html#fixity

Other than that, by "buld-in" Liquid Haskell does not defined any other infix ops.

PR #867, opened 2016-11-10
add local defines

nikivazou:
With define we can replace Haskell variables with symbols during constrain. generation 

eg, in the following example

https://github.com/ucsd-progsys/liquidhaskell/pull/867/commits/0123b4081cad10df96ee23f6cc621766b32c8cdc#diff-6e20e8531df8072c56366fdfad9b6c4fR48

```
{-@ define $ceq = eqN @-}
```

replaces the internal class method `$ceq` with `eqN`. 

I am not supplying documentation yet, as this is a temporal solution.


ranjitjhala:
What does this PR do? Whats a test that did (or did not) work earlier that
does not (or does) now?

Issue #866, opened 2016-11-10
Crash on irrefutable patterns


nikivazou:
The following code crashes Liquid Haskell

```
data Body = Body Int Int 

genBody :: Int -> Body
genBody s = Body x y 
  where 
    (x:y:_) = take 2 $ repeat s 
```

Error: 

```
 /Users/niki/liquidtypes/liquidhaskell/verified-instances/examples/nbody/allpairs.hs:4:1-7: Error: Uh oh.

 4 | genBody s = Body x y
     ^^^^^^^

     addC: malformed constraint:
{lq_tmp$x##263 : ({lq_tmp$x##259 : Int^"lq_tmp$x##248" | $k_##247[lq_tmp$x##245:=lq_anf$##1677723166##dpg][VV##246:=lq_tmp$x##259][VV##256:=lq_tmp$x##263]}, {lq_tmp$x##260 : Int^"lq_tmp$x##251" | $k_##250[VV##249:=lq_tmp$x##260][lq_tmp$x##245:=lq_anf$##1677723166##dpg][VV##256:=lq_tmp$x##263]})^"lq_tmp$x##258" "lq_tmp$x##2" | $k_##257[lq_tmp$x##245:=lq_anf$##1677723166##dpg][VV##256:=lq_tmp$x##263]}
 <:
Body((GHC.Types.Int, GHC.Types.Int), Main.Body)
```

ranjitjhala:
This may be due to the simplifications done by eliminate .. can you run
with no simplify core?

ranjitjhala:
fixed by #876

Issue #862, opened 2016-10-25
Huge slowdowns when using `undefined` in proofs


nikivazou:
For example, the time to check the Ackermann proofs goes from `14s` to `473s` with the following `undefined` insertion 

https://github.com/ucsd-progsys/liquidhaskell/commit/5d6c0f3995a85bb566ea22b360c77ccf17313d28

nikivazou:
@ranjitjhala you said you were interested in taking a look at that

ranjitjhala:
wow! Thanks for finding this!

Will look at this post-PLDI!

nikivazou:
Np! My current workaround is to use `trivial` and actually get the failing proof instead of `undefined`. 

Thanks @vikraman for forcing me to report this:)

gridaphobe:
Does it also happen if you replace `undefined` with something like `let x = x in x`? It sounds very similar to an issue we had a few years ago in bytestring/text where having `false` in the environment would cause an explosion in the size of inferred types and a corresponding slowdown.

ranjitjhala:
looks like `tests/pos/mapreduce.hs` also has this problem...

ranjitjhala:
closed by #879

PR #876, opened 2016-12-03
Fix for issue #866

ranjitjhala:
Fix for issue #866 

- also, moving `tests/pos/mapreduce.hs` into `tests/todo/` till the undefined issue #862  is fixed.


ranjitjhala:
@gridaphobe we're still doing the CI right?!

PR #877, opened 2016-12-03
Add cabal install instruction to check GHC version

shhyou:
This pull request updates installation guide when using `cabal` to ensure that the correct GHC version is in the environment. Without this, `cabal install liquidhaskell` with GHC 8 will try to install liquidhaskell-0.4.x.x which does not has an upper bound on GHC's version (and will fail to compile).
The user might not realize that `cabal install` is trying an old version due to dependency problem.

```
$ ghc -V
The Glorious Glasgow Haskell Compilation System, version 8.0.1
$ cabal install liquidhaskell --dry-run
Resolving dependencies...
In order, the following would be installed (use -v for more details):
... [skipped] ...
liquid-fixpoint-0.3.0.1 (latest: 0.5.0.1)
liquidhaskell-0.4.1.1 (latest: 0.6.0.0)
```

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -23,6 +23,13 @@ Simply do:
 
 We are working to put `liquid` on `stackage`.
 
+Note: Currently, LiquidHaskell on hackage does not compile against GHC 8.
+Please make sure you are installing with GHC version less than 8. You can
+designate a specific version of LiquidHaskell to ensure that the correct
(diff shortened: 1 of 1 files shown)


ranjitjhala:
Thanks very much for this! [hopefully we'll get to GHC 8 soon ...]

PR #879, opened 2016-12-05
Eliminate all K-Vars in `higher order` mode

ranjitjhala:
otherwise, instantiation can be dog slow.

This PR makes prover benchmarks *much* faster.

Matching PR for 

https://github.com/ucsd-progsys/liquid-fixpoint/pull/271

**NOTE** running with "--higherorder" now *by default* switches off all qualifier usage; i.e. all inference is via eliminate, if you have recursion, you have to specify the types.


ranjitjhala:
fixes #862

PR #899, opened 2016-12-23
Manually allow verified type classes 

nikivazou:
This branch provides the first step to allow verified type classes. 

I) It allows assumed types in the refined instances (see [here](https://github.com/ucsd-progsys/liquidhaskell/blob/9f43d8b566682589520f8eea576cb2e1341a565b/tests/pos/VerifiedMonoid.hs#L86)).

II) It allows to refine the instances of refined type classes (simply by deactivating the relevant [check](https://github.com/ucsd-progsys/liquidhaskell/blob/9f43d8b566682589520f8eea576cb2e1341a565b/src/Language/Haskell/Liquid/Bare/Check.hs#L84)). Note this is unsound see below. 

With these two features enables, the user can follow three steps of annotations to allow Verified Type Classes. These three steps are described by example [here](https://github.com/ucsd-progsys/liquidhaskell/blob/9f43d8b566682589520f8eea576cb2e1341a565b/tests/pos/VerifiedMonoid.hs#L86) and can be automated by Liquid Haskell, if the user specifies that class methods should be reflected. Eg
```
{-@ class VerifiedMonoid a where
    reflect mempty  :: a 
    reflect mappend :: a -> a -> a  
    leftId  :: x:a -> {v:Proof | mappend mempty x = x }
    rightId :: x:a -> {v:Proof | mappend x mempty = x }
    assoc   :: x:a -> y:a -> z:a -> {v:Proof | mappend x (mappend y z) = mappend (mappend x y) z}
  @-}
```
should automatically create (this)[https://github.com/ucsd-progsys/liquidhaskell/blob/9f43d8b566682589520f8eea576cb2e1341a565b/tests/pos/VerifiedMonoid.hs#L67-L92]
 

Note that II) can lead to unsoundness as described in (crash/RClass.hs)[https://github.com/ucsd-progsys/liquidhaskell/blob/9f43d8b566682589520f8eea576cb2e1341a565b/tests/crash/RClass.hs]. The following three suggestions deal wih this unsoundness.
1. Leave the check (that instances of refined classes are not refined) but change error to a warning. 
2. Do not allow the user to refine instances of refined type classes but allow Liquid Haskell to automatically generated refined instances when reflection occurs. 
3. Check that the two refinements are consistent by firing proper subtyping queries. Eg. for 

```
class A b where 
  m1 :: x:b -> {v:b | v <= x }
  mi :: ti
```

```
instance A Int where 
  m1 :: x:{Int | 0 < x} -> {v:Int | v == x } 
  mi :: ti'
```

Check that `ti' <: ti`. This is the propert and most complicated (implementation wise) solution.


ranjitjhala:
Can you submit this against no-prop?

nikivazou:
@ranjitjhala there is only one minor conflict. 

Can I merge this? 

Seems that github has now web interface to resolve conflicts!

ranjitjhala:
What did you do? Ie minor conflicts between what? Are you pulling stuff
from noprop?

nikivazou:
@ranjitjhala I created a different pull requiest to merge this edits to your branch

ranjitjhala:
Ok! So I should just keep working on my branch right? Or something else?

I think I am "almost" done... which means:

0 - still need to run text and bytestring
1 - a couple of inline tests fail
2 - all the fancy higher order tests fail (as I commented out defunc
temporarily)

nikivazou:
Yes, just keep working at your branch. I will merge mine on develop since
there are no conflicts

PR #906, opened 2017-01-05
Use Bool instead of Prop

ranjitjhala:
Matching PR for: https://github.com/ucsd-progsys/liquid-fixpoint/pull/275. 

From `CHANGES.md` 

- **breaking change** Remove the `Bool` vs. `Prop` distinction. This means that: 

    * signatures that use(d) `Prop` as a type, e.g. 
      `foo :: Int -> Prop` should just be `foo :: Int -> Bool`.

    * refinements that use(d) `Prop v` e.g. 
      `isNull :: xs:[a] -> {v:Bool | Prop v <=> len xs > 0}`
      should just be `isNull :: xs:[a] -> {v:Bool | v <=> len xs > 0}`.

- Add `--eliminate={none, some, all}`. Here
  * `none` means don't use eliminate at all, use qualifiers everywhere (old-style)
  * `some` which is the **DEFAULT**  -- means eliminate all the **non-cut** variables
  * `all`  means eliminate where you can, and solve *cut* variables to `True`.

- Change `--higherorder` so that it uses *only* the qualifiers obtained from
  type aliases (e.g. `type Nat = {v:Int | ... }`) and nothing else. This
  requires `eliminate=some`.


ranjitjhala:
resolved conflicts. waiting for @gridaphobe to figure out the circle issues...

nikivazou:
@ranjitjhala @gridaphobe circle fails because of Haddock failure.... Let me update

nikivazou:
Merged! Thanks so much @ranjitjhala !!!

PR #907, opened 2017-01-05
Fix a broken link

y-taka-23:
I'm afraid that the original target `docs/tutorial` no longer exists.


ranjitjhala:
Oops, many thanks for catching that!!

PR #912, opened 2017-01-07
Smt axiomatization

nikivazou:
See tests/pos|neg/Automate.hs for axiomatization of the fib function. 

Surprisingly, SMT proving `fibA 30 == 832040` is faster than running `fibA 30` in Haskell! 

I suggest 
  - rename all current axiomatize to reflect 
  - use the current reflect to build further fuel strategies etc?

### DIFF
--- a/benchmarks/pldi17/neg/Ackermann.hs
+++ b/benchmarks/pldi17/neg/Ackermann.hs
@@ -13,7 +13,7 @@ import Helper
 
 -- | First ackermann definition
 
-{-@ axiomatize ack @-}
+{-@ reflect ack @-}
 {-@ ack :: n:Nat -> x:Nat -> Nat / [n, x] @-}
 ack :: Int -> Int -> Int
 ack n x
@@ -26,7 +26,7 @@ ack n x
 
 -- | Second ackermann definition
 
-{-@ axiomatize iack @-}
+{-@ reflect iack @-}
 {-@ iack :: Nat -> Nat -> Nat -> Nat @-}
 
 iack :: Int -> Int -> Int -> Int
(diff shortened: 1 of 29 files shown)


atondwal:
Does this only interface with Fixpoint directly? AFAIK the `Inputable` instance for `Expr` doesn't parse `forall`s 

```
*Main [--cut--]> rr' "" "bind 0 ax : {v: bool | (f == g (a+3))}" :: FInfo ()
FI {cm = fromList [], ws = fromList [], bs = BE {_beSize = 1, beBinds = fromList [(0,("ax",{v : bool | [(f = (g (a + 3)))]}))]}, gLits = [], dLits = [], kuts = KS {ksVars = fromList []}, quals = [], bindInfo = fromList [], hoInfo = HOI {hoBinds = False, hoQuals = False}, asserts = []}
*Main [--cut--]> rr' "" "bind 0 ax : {v: bool | forall a. (f == g (a+3))}" :: FInfo ()
*** Exception: Error [Error1 {errLoc = SS {sp_start = (line 1, column 37), sp_stop = (line 1, column 37)}, errMsg = parseError (line 1, column 37):
unexpected "="
expecting operator, ")", "::" or ":"
when parsing from }]
```

ranjitjhala:
may have to extend the parser to handle `forall`

nikivazou:
Interesting observation: axiomatization of `compose f g x = f (g x)` causes divergence even though `compose` is not revursive! 

But note that after defunctionalization all `compose`, `f`, `g` are ints and the SMT axiom is 

`forall (x:Int, g:Int, f:Int) (apply (apply (apply (compose f)) g) x) == apply f (apply g x)`

PR #913, opened 2017-01-08
Fix a broken link

y-taka-23:
Sorry for bothering you again, but the hierarchy looks incorrect.


ranjitjhala:
Thanks for catching that!!

PR #922, opened 2017-01-15
Enable Hackage-friendly stack.yaml setting

alanz:
This will transparently add upper and lower bounds to all package dependencies not currently having them, based on the snapshot in use, when the package is uploaded via either stack upload or stack sdist.


gridaphobe:
Ooh, this sounds wonderful, thanks!

PR #923, opened 2017-01-16
Instantiation of axioms within Liquid Haskell

nikivazou:
Using the flag 

- `{-@ LIQUID "--automatic-instances=smtinstances" @-}` for SMT (unpredictable) instantiation 
(e.g, [tests/pos/Automate.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/d16ee75b39ff5be9a855e20e912944bc0fe74bde/tests/pos/Automate.hs))
-  `{-@ LIQUID "--automatic-instances=liquidinstances" @-}` for instantiations withing Liquid Haskell. Example currently passing 
    -  [tests/pos/MapFusion.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/d16ee75b39ff5be9a855e20e912944bc0fe74bde/tests/pos/MapFusion.hs)
    - [pos/FoldrUniversal.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/d16ee75b39ff5be9a855e20e912944bc0fe74bde/benchmarks/proofautomation/pos/FoldrUniversal.hs)
    - [proofautomation/pos/Compose.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/d16ee75b39ff5be9a855e20e912944bc0fe74bde/benchmarks/proofautomation/pos/Compose.hs)
    - [tests/pos/LiquidAutomate.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/d16ee75b39ff5be9a855e20e912944bc0fe74bde/tests/pos/LiquidAutomate.hs)

__Currently__ instantiation happens according to a user specified (default to 2) user parameter that applies to all axioms. Eg. `{-@ LIQUID "--fuel=3" @-}`. You can see how the fuel parameter affects instantiations in the classic fib example: [tests/pos/LiquidAutomate.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/d16ee75b39ff5be9a855e20e912944bc0fe74bde/tests/pos/LiquidAutomate.hs)

Note: This is totally beta so I am not going to update the readme.md yet. But should not break anything since everything is flagged, so let's merge!


ranjitjhala:
can you make that 

```
{-@ LIQUID "--automatic-instances=smt" @-}
{-@ LIQUID "--automatic-instances=liquid" @-}
```

?

ranjitjhala:
(and then lets merge...)

nikivazou:
We already have an SMT data constructor and I do not really know how to allow the user write a parameter different than the name of the data constructor ... (Though shouldn't be that difficult...)

gridaphobe:
I have a small objection to the name of the flag, it doesn't really explain what it does (and `instances` makes me think immediately of type classes). Perhaps something more clear like `--unroll-functions=smt/liquid`?

nikivazou:
Well, since this branch is making a lot of edits, I will merge and opened an issue to discuss the flag names.

nikivazou:
@gridaphobe  Hope the merge in the fixpoint does not break this too....

gridaphobe:
Well, there's a reason we're using a submodule to include fixpoint in LH, so we can avoid including breaking changes until we're ready for them.

nikivazou:
Yes, but now the breaking and working changes are all together merged in develop, so not sure how git reacts to that

ranjitjhala:
Ok given @nikivazou explanation. I am fine with sticking with her original
thing. Maybe we can put @gridaphobe suggestion in the help text for that
flag.

PR #930, opened 2017-01-26
Import reflects

ranjitjhala:
This PR adds (preliminary!) support for importing "lifted specifications".  

Matching PR for https://github.com/ucsd-progsys/liquid-fixpoint/pull/285

1. When you *define* a Haskell function as a `measure`, 
or `inline` or `reflect` in a *Library Module*,
2. You can then *use* the corresponding definition when 
you import the library in *Client Modules*.

For example, see `tests/pos/Reflect{Lib, Client}{0, 1, 2}.hs` 

This works by *serializing* the relevant definitions from the `Library`
in a file `.liquid/Library.hs.bspec`, and then loading those in when 
the `Library` is imported.

**NOTE:** 

1. Only `BareSpec` specifications can be serialized -- so 
as we go forward, make sure that all definitions, at some point, live
as a `BareSpec`.

2. Currently we require that the `Library.hs` has been processed 
before the client. (The cabal/stack integration should take care of this.)

3. It would be nice to figure out how to tuck the information 
in the `.bspec` directly inside the `.hi` files (perhaps as GHC annotations.)

4. This PR doesn't at all deal with the terrible hack that all specifications 
have "global" names; that still needs to be fixed.


ranjitjhala:
ok, before merging I will:

1. clean up comments and general debugging noise.
2. add more explanation to the above and README.

ranjitjhala:
@gridaphobe :  looks like there is some weird cabal / dependency hassle 
due to the new fixpoint 0.6.0.1 which uses a new `pretty` which perhaps 
has to do with:
 
https://github.com/ucsd-progsys/liquid-fixpoint/pull/284 

? Can you figure it out ? (It builds locally for me with stack, but *after* there are some rather dire complaints from stack...)

gridaphobe:
Ack, sorry, I forgot to sort that out. Can you try pulling in the latest fixpoint develop?

ranjitjhala:
ok lets see what happens! (what did you do btw?)

ranjitjhala:
ah some black magic with cabal... oh well lets see what ci says!

gridaphobe:
CPP nonsense to allow fixpoint to build against the old *and* new `pretty`. 

1.1.3 changed the implementation of `Doc` such that our `Serializable Doc` instance requires extra instances, for types that do not exist pre-1.1.3. Sadly, `pretty-1.1.2` is depended on by some combination of TH and GHC 7.10, so we can’t just say use 1.1.3 or go home.

ranjitjhala:
Yikes, well lets see what happens. Locally `stack` says the rather ominous sounding:

```
    Configuring liquidhaskell-0.7.0.0...
    Warning: This package indirectly depends on multiple versions of the same
    package. This is highly likely to cause a compile failure.
    package template-haskell-2.10.0.0 requires pretty-1.1.2.0
    package liquidhaskell-0.7.0.0 requires pretty-1.1.3.1
    package liquid-fixpoint-0.6.0.1 requires pretty-1.1.3.1
    package Diff-0.3.2 requires pretty-1.1.3.1
    package Cabal-1.22.8.0 requires pretty-1.1.3.1
    Preprocessing library liquidhaskell-0.7.0.0...
```

ranjitjhala:
ooh ... is it because I added the following to stack.yaml

```
extra-deps:
- pretty-1.1.3.1
```

?

gridaphobe:
Yes, the pretty-1.1.3.1 constraint is not needed, better to let stack pick the one in stackage.

ranjitjhala:
seems to have built on CI ...

PR #936, opened 2017-01-30
Eval proof strategy

nikivazou:
Proofs are now done by (in place) evaluation. 

I will formalize and describe the evaluation strategy once all pldi tests pass.

### DIFF
--- a/TODO.markdown
+++ b/TODO.markdown
@@ -1,3 +1,12 @@
+- Numerics 
+  - Overview.hs
+  - Ack
+  - Fib
+
+- Unification -- HERE 
+
+
+
 TODO
 ====
 
--- a/TODO.markdown
+++ b/TODO.markdown
@@ -3,8 +3,6 @@
   - Ack
   - Fib
 
-- Unification -- HERE 
-
 
 
 TODO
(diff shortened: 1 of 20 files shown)


ranjitjhala:
Cool! Does this (automatically) prove the following equality:


```haskell
((x:xs) ++ ys) ++ zs == xs : ((xs ++ ys) ++ zs)
```

nikivazou:
Yes, all these are proved 

https://github.com/ucsd-progsys/liquidhaskell/blob/a43a6659e72b18a918f9929a6583007701332275/benchmarks/proofautomation/pos/Lists.hs

(But I have added few hacks...)

The only things that are not proved are the numeric fib operators, for which we go back to my initial (no in-place) generation of equations. 


Moreover, verification time for `Unification` (that you complained) has actually dropped.

ranjitjhala:
cool!

> (But I have added few hacks...)
hmm ... :)

ok will have to hear what you did exactly... I thought this was quite
non-trivial to implement.

What has "Unification" dropped to?

nikivazou:
I definitely need to formalize the evaluation strategy...
I mean I have few hacks so that all our pldi bench are automated. 

Currently mostly the Reader examples fail. Do you want to speak sometime tomorrow?

PR #938, opened 2017-02-06
Eval with smt

nikivazou:


ranjitjhala:
This is absolutely not the place for this code; in particular constraint generation should not be cluttered in this way.

ranjitjhala:
This is absolutely not the place for this code; in particular constraint generation should not be cluttered in this way.

Also there are no measurements of how the symbolic eval is affecting runtime, and afaict it is NOT being guarded by a flag. 

so **please undo the merge** into develop until either you have performance numbers or you have a flag.

PR #939, opened 2017-03-01
Add link to the main web site

alanz:
There should perhaps be a link to the tutorial from here too.


ranjitjhala:
Ah, yes excellent idea -- thanks @alanz !

PR #946, opened 2017-03-08
T289

ranjitjhala:
Fix for liquid-fixpoint/#289


ranjitjhala:
Will merge after CI says ok...

PR #954, opened 2017-03-14
Add trivial parser test suite

chrisdone:
I anticipate this is an easy to merge PR. This adds a simple test suite that you can run like this:

```
bash-3.2$ stack test :liquidhaskell-parser
liquidhaskell-0.7.0.0: test (suite: liquidhaskell-parser)


Should succeed
  Int
  k:Int -> Int
Should fail
  Maybe k:Int -> Int

Finished in 0.0048 seconds
3 examples, 0 failures
```

I spent most of my energy on trying to get Eq or Show instances for the `Pspec` and related types, but gave up. So I just put in three cases. The Show instance for Psec is now derived. If you want the debugging one, it's still there commented out. 

When you discover a new case (that should or shouldn't work), you just run this in the REPL

```haskell
> parseSingleSpec "a :: {x:Int|x>0} -> ()"
Right "Asrts ([\"a\" defined from: \"<test input>\" (line 1, column 1) to: \"<test input>\" (line 1, column 2)],(lq_tmp$db##0:{x##1 : Int | x##1 > 0} -> () defined from: \"<test input>\" (line 1, column 6) to: \"<test input>\" (line 1, column 23),Nothing))"
```

and then you can just copy/paste that into the test suite as a fail/succeed case. 

As I discover odd parse errors or messages that could be improved, I can add them here.

### DIFF
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -281,3 +281,19 @@ test-suite test
                ,     tasty-hunit >= 0.9
                ,     tasty-rerun >= 1.1
                ,     transformers >= 0.3
+
+test-suite liquidhaskell-parser
+  default-language: Haskell2010
+  type: exitcode-stdio-1.0
+  hs-source-dirs: tests
+  ghc-options: -W
+  main-is: Parser.hs
+  build-depends:
+    base  >= 4 && < 5,
+    liquidhaskell,
+    HUnit,
+    hspec,
+    parsec,
+    bifunctors,
+    text,
+    mtl
--- a/src/Language/Haskell/Liquid/Parse.hs
+++ b/src/Language/Haskell/Liquid/Parse.hs
@@ -701,10 +701,10 @@ data Pspec ty ctor
   | Varia   (LocSymbol, [Variance])
   | BFix    ()
   | Define  (LocSymbol, Symbol)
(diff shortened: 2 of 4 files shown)


ranjitjhala:
SUPER!! Thanks @chrisdone!! I'll try to create
the `Eq` and `Show` instances, and add more tests
from our issues.

chrisdone:
@ranjitjhala Cool. I don't mind adding the Eq instance -- I did start it, but the `Predicate` type has an Eq instance which is defined further up the module hierarchy (because it uses functions from that module), yet it's defined equal or lower than the Pspec type, so we run into a cyclic issue. I'm sure there's a way to resolve that (moving things around), but it seemed too intrusive for an initial PR.

chrisdone:
Specifically: 

defined here: https://github.com/ucsd-progsys/liquidhaskell/blob/8fe140f10e9fb72292945bdc35673f58b5bec663/src/Language/Haskell/Liquid/Types.hs#L477

Eq instance here: https://github.com/ucsd-progsys/liquidhaskell/blob/8fe140f10e9fb72292945bdc35673f58b5bec663/src/Language/Haskell/Liquid/Types/RefType.hs#L329

Hope that helps.

PR #963, opened 2017-03-16
Fix travis build process

### LINKED ISSUE #960: travis does not install z3 lib
Travis runs ubuntu precise (12.04).

Z3 is not available via apt.

It seems there are 2 options

1. Install z3 from source
2. Run via a docker image, with a later distribution

Are there any preferences?

alanz:
The z3 library was no longer available in ubuntu 12.04 (precise) as used
by travis.  Install it from the hvr/z3 PPA.

Thanks @hvr

Note: The tests currently fail, and I think the complex script based
stack build can be simplified with the more modern stack (1.4 now)

Closes #960

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,22 @@ branches:
   only:
   - master
   - develop
+  - parser-az
+  - travis
+
+# addons:
+#   apt:
+#     packages:
+#     - libgmp-dev
 
 addons:
   apt:
     packages:
-    - libgmp-dev
-    - z3
+      - libgmp-dev
+      - libz3-dev
+    sources:
+      - sourceline: ppa:hvr/z3
+
 
 # Caching so the next build will be fast too.
 cache:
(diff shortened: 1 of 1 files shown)


ranjitjhala:
Awesome! @gridaphobe shall we switch to this?

hvr:
If there's interest, I can try to set you up with a `cabal new-build` based Travis CI script. This project seems well suited to benefit from cabal's new [nix-style local build](http://cabal.readthedocs.io/en/latest/nix-local-build-overview.html) features and it would provide me with the opportunity to figure out if something's missing in cabal for real-world use-cases.

gridaphobe:
We haven't used TravisCI in a while, we switched to CircleCI since it doesn't have the 50min hard timeout, so we can run a single test build. It ends up being much faster than waiting for the 5 or 6 Travis builds to complete.

I would probably favor simply removing the travis scripts.

alanz:
How does CircleCI work? Are the builds publicly visible?  If I want to run the tests against my fork, is there an easy way to do that, via a free account or some such?

alanz:
Ok, I found the comments in #964, will continue there

nikivazou:
So, should we merge this even though we use Circle instead of Travis?

alanz:
I would say it must either be fixed to pass properly, or delete.

And the README should say that Circle is used for  CI

ranjitjhala:
hey whats going on with this? maybe @gridaphobe can take a look post the OOPSLA deadline?

alanz:
In my view we should kill the travis file

ranjitjhala:
Ping @gridaphobe?

gridaphobe:
Oops, sorry, I'm going to close this PR and kill the travis file as suggested..

PR #965, opened 2017-03-16
Add cabal.project file

hvr:
This provides cabal project meta-data for cabal to make it easier to build from Git

For more information see http://cabal.readthedocs.io/en/latest/nix-local-build-overview.html

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,0 +1,12 @@
+-- For more information see
+--  http://cabal.readthedocs.io/en/latest/nix-local-build-overview.html
+
+packages: .
+          ./liquid-fixpoint
+          ./liquiddesugar
+
+package liquid-fixpoint
+  flags: devel
+
+package liquidhaskell
+  flags: devel


hvr:
Btw, if there's interest, I can try to update `circle.yml` to make use of `cabal new-build` (which would also make use of `cabal.project`)

gridaphobe:
Cool, thanks!

Is there a benefit to using `new-build` for CI? It seemed like more of an improvement for the development workflow.

hvr:
@gridaphobe Well, one of the first things I did when `cabal new-build` became available was migrating CI jobs to it because  `cabal new-build` gives you control over many aspects and has quite a few benefits, especially regarding it's nix-style implicit sandboxing which allows for more intelligent caching (Travis supports caching data between builds, not sure if Circle CI does that too). Soon there's gonna be garbage collection which should help reducing the caching footprint to the minimum possible and require less maintainance.

Here's a simple example for a simple multi-package CI job: https://github.com/hvr/uuid/blob/master/.travis.yml

PR #975, opened 2017-03-23
Clean up the parser

alanz:
This PR is a companion to https://github.com/ucsd-progsys/liquid-fixpoint/pull/301

The parser is now based on `reserved` and `reservedOp`, simplifying token detection.

Also, the top level `specP` no longer has any `try` in it, so error reporting should be better localised, as it will not back up to the very top of the failing item.


alanz:
Note: the circle.yml file will need attention when it is merged, this one explictly checks out and builds the https://github.com/alanz/liquid-fixpoint repo `parser-az` branch.

See https://circleci.com/gh/alanz/liquidhaskell/51#config/containers/0

ranjitjhala:
👍 👍 👍 outstanding! I am going to merge this in, and _try_ to update the circle.yml thing too, but may need help from @gridaphobe if the latter step falls to pieces... :)

alanz:
You just need to revert to what it was before my patch

alanz:
I think the current problem is with caching., where the sandbox is regarded as part of the source.

add 

```
- rm -fr .cabal-sandbox
- rm cabal.sandbox.config
```

to the bit just before setting up the sandbox.

And I would suggest putting the sandbox out of the source tree (in a cached directory) so it can be managed by the circle caching process

alanz:
See https://github.com/ucsd-progsys/liquidhaskell/pull/976 for a working `circle.yml`

PR #976, opened 2017-03-23
Blow away stale sandbox

alanz:


gridaphobe:
@alanz looks like the `~` in `~/cabal-sandbox` isn't being expanded correctly.

From the build log:
```
$ cabal sandbox --sandbox=~/cabal-sandbox init
Writing a default package environment file to
/home/ubuntu/liquidhaskell/cabal.sandbox.config
Creating a new sandbox at /home/ubuntu/liquidhaskell/~/cabal-sandbox
```

```
Creating dependency cache from the following paths:
...
  ✗ /home/ubuntu/cabal-sandbox
...
Creating source cache from the following paths:
  ✓ /home/ubuntu/liquidhaskell/.git
```

Thanks for all the work you're putting in to this!

alanz:
ok, so the build works because it is a fresh sandbox.

ranjitjhala:
yes, many MANY thanks -- we've been quite foxed by this business!!!

gridaphobe:
Looks like circle is happy!

PR #977, opened 2017-03-24
[WIP don't merge] Improve error messages

alanz:
Finally, use the parsec ability to generate reaonable error messages.

To start with I just put the name of the production in the `<?>` tags, these can be changed to something more useful over time.

Also, we now get the actual error message, rather than just a location and "leftover when parsing".


ranjitjhala:
Oh wow!!!! Thanks @alanz!

PR #985, opened 2017-04-09
Harmonise reserved words to be lowercase

### LINKED ISSUE #971: parser keyword case consistency
Most of the top level declarations in liquidhaskell are lower case, except for

```
Local
Decrease
LAZYVAR
Strict
Lazy
LIQUID
```

Is there a specific reason these are not all lower case?

alanz:
Convert the following to be all lowercase

```
Local
Decrease
LAZYVAR
```

Closes #971


ranjitjhala:
fantastic!!! 🎉

PR #986, opened 2017-04-10
Remove "defined" keyword, it is a synonym for "measure"

### LINKED ISSUE #972: "defined" vs "measure"
It seems that `defined` is parsed as a synonym for `measure`.

Is this intentional? Are both needed?

alanz:
Closes #972


ranjitjhala:
🎉

PR #988, opened 2017-04-14
fixpoint bump for solveNative refactor

atondwal:
This is mostly to make sure that the tests don't break. For some reason, they
don't seem to run quite right on my machine, so I want to test it in CircleCI


atondwal:
grr, Circle doesn't pull in commits from fixpoint pull request while getting submodules :(

PR #991, opened 2017-04-20
First two chapters of SF.

notcome:
I have partially ported proofs (given or exercises) into Liquid Haskell. The “partial” is to avoid unnecessary repetition.

### DIFF
--- a/tests/pos/bar.hs
+++ b/tests/pos/bar.hs
@@ -1,5 +1,0 @@
-module Bar where
-
-import Foo
-
-foo = bar
--- a/tests/pos/sf/Basics.hs
+++ b/tests/pos/sf/Basics.hs
@@ -1,0 +1,153 @@
+module Basics (
+  Bool(..),
+  boolAnd, boolOr,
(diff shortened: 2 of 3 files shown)


ranjitjhala:
@notcome -- Thanks! It looks like `sf/Induction.hs` and `sf/Basics.hs` are failing on CI can you see whats going on?

notcome:
I suspect this is due to that I didn’t add an `--exact-data-con` option. A new version is updated.

ranjitjhala:
Hmm. but then how did it work for you locally?

​

notcome:
I just run liquid with `liquid --exact-data-con` every time. It is just an upper arrow key thing, so I can stay lazy. The new version contains the `{-@ LIQUID "--exact-data-con" @-}` at the top of each file.

ranjitjhala:
ah, ok :-)

ranjitjhala:
:Tara:

ranjitjhala:
:tada:

PR #1002, opened 2017-04-27
extend pbe-bug #997

ranjitjhala:


ranjitjhala:
Is this the same reason (e.g. not expanding under `&&`)?

PR #1006, opened 2017-04-28
Add liquidinstances in Readme.md

nikivazou:


ranjitjhala:
I'm afraid I don't really follow the explanations -- and I've been _using_ the instances. Don't even quite follow the difference between 'global' and local' ? But we can always iterate...

PR #1016, opened 2017-05-02
Liquid desugar ghc 8

nikivazou:


gridaphobe:
Wow, that was quick!

nikivazou:
I had it while moving to ghc-8, but then just removed it because I thought we do not need it. 

Still Travis likes NOTHING.

gridaphobe:
ping @atondwal, seems like your hack ought to help?

atondwal:
Hmm git seems to be getting the object, but isn't realizing that it's a commit for some reason. Try adding a dummy commit to desugar and updating the submodule commit?

Issue #1013, opened 2017-05-02
Refinements using RankNTypes cause Liquid Haskell to say "Uh oh"


RyanGlScott:
I'm trying to prove properties of `Functor` instances. For example, the `fmap id` law:

```haskell
-- VF.hs
{-@ LIQUID "--higherorder"     @-}
{-@ LIQUID "--totality"        @-}
{-@ LIQUID "--exactdc"         @-}
{-# LANGUAGE RankNTypes #-}
module VF (VerifiedFunctor(..)) where

import Language.Haskell.Liquid.ProofCombinators

{-@ axiomatize _identity @-}
_identity :: a -> a
_identity x = x
{-# INLINE _identity #-}

{-@
data VerifiedFunctor m = VerifiedFunctor {
    fmap   :: forall a b. (a -> b) -> m a -> m b
  , fmapId :: forall a. x:m a -> { fmap _identity x == x }
  }
@-}
data VerifiedFunctor m = VerifiedFunctor {
    fmap   :: forall a b. (a -> b) -> m a -> m b
  , fmapId :: forall a. m a -> Proof
  }
```

I can do so for an extremely simple datatype:

```haskell
-- Generics.hs
{-@ LIQUID "--higherorder"    @-}
{-@ LIQUID "--totality"       @-}
{-@ LIQUID "--exactdc"        @-}
{-# LANGUAGE RankNTypes #-}
module Generics where

import VF
import Language.Haskell.Liquid.ProofCombinators

{-@ axiomatize _identity @-}
_identity :: a -> a
_identity x = x
{-# INLINE _identity #-}

{-@ data Par1 p = Par1 { unPar1 :: p } @-}
data Par1 p = Par1 { unPar1 :: p }

{-@ axiomatize fmapPar1 @-}
fmapPar1 :: (p -> q) -> Par1 p -> Par1 q
fmapPar1 f (Par1 p) = Par1 (f p)

{-@ fmapPar1Id :: x:Par1 p
               -> { fmapPar1 _identity x == x }
@-}
fmapPar1Id :: Par1 p -> Proof
fmapPar1Id par@(Par1 p)
  =   fmapPar1 _identity par
  ==. Par1 (_identity p)
  ==. Par1 p
  ==. par
  *** QED
```

Running `liquid` on this works fine. But if I try a slightly more complicated datatype:

```haskell
-- Generics2.hs
{-@ LIQUID "--higherorder"    @-}
[...]

PR #1022, opened 2017-05-05
Type eq with deep type substitution 

nikivazou:
Fix for #1013


ranjitjhala:
This is going to conflict with #1021 (there are changes in BARE).

Please do not merge till I am done with that. (Which should be

imminent, as I disabled haddock...)

ranjitjhala:
lets see what circle says...

PR #1026, opened 2017-05-07
Gradual

nikivazou:


nikivazou:
I removed them right? 

I am playing with interactive inference

ranjitjhala:
yes, they're removed. just wondering when/why they were added...

nikivazou:
@ranjitjhala did you see that?

https://github.com/ucsd-progsys/liquid-fixpoint/compare/bc80878ea33d79c20d40cedbad8e128a91022b66...72b1f4cd8664da5333c43a512941698cd2221c99#diff-47aa83d3fa74363a37e5bc741d890568R337

I did not check if it still exists after your today's edits, but I broke after your import edits.

PR #1040, opened 2017-06-04
dependent pairs

nikivazou:
Syntactic sugar for dependent pairs (without explicitely using Abstract Refinements)


ranjitjhala:
What is the motivation here, that the abs-ref syntax is a bit heavyweight for this common use case? Or something else? At any rate, please add documentation to README?

nikivazou:
Exactly! 

In my example `(x::Int, {v:Int | x <= v})` is more natural than its desugar `(Int, Int)<\x -> {v:Int | x <= v}>`

ranjitjhala:
Cool, does this work for, say, triples?

nikivazou:
@ranjitjhala it works for triples etc
 I did not special case pairs and again, it is just parsing syntactic sugar!

PR #1047, opened 2017-06-15
don't add the directory containing the file to the search path

gridaphobe:
See NOTE [searchpath] in L.H.L.UX.CmdLine


ranjitjhala:
Rather surprised that the tests pass with this change...

gridaphobe:
haha, me too..

PR #1050, opened 2017-06-15
reorganize includes to support verifying `base`

gridaphobe:
our includes are not very well structured and cause spurious
imports. this hasn't been a problem so far, but is when we start trying
to verify base, which has a complicated, staged hierarchy.

in particular, we have to be careful that our optional flags like
--totality don't force an import of Prelude.


ranjitjhala:
whats going on with this PR?

src/Language/Haskell/Liquid/GHC/Interface.hs
```diff
@@ -143,8 +143,10 @@ configureDynFlags cfg tmp = do
                                                 (ModRenaming True [])
                                 : packageFlags df'
                  -- , profAuto     = ProfAutoCalls
-                 , ghcLink      = LinkInMemory
-                 , hscTarget    = HscInterpreted
+                 -- , ghcLink      = LinkInMemory
+                 -- , hscTarget    = HscInterpreted
+                 , ghcLink      = NoLink
+                 , hscTarget    = HscNothing
```
gridaphobe: @ranjitjhala it's not quite ready because of this change. Another issue Sean has run into is that base uses unboxed tuples, which ghci (and thus LH via `hscTarget=HscInterpreted` does not support).

Using `HscInterpreted` has always felt wrong to me since we don't actually need to do code generation, but switching to `HscNothing` causes issues for name resolution as it affects which / how many names GHC keeps in its environment after type checking...

I'm trying to fix the name resolution to work with `HscNothing`, but got sidetracked by drafting the dissertation. I'll take another look at it this afternoon.

ranjitjhala:
ok, thanks!

gridaphobe:
@ranjitjhala @nikivazou made some progress, I think there are three things left holding this back:

1. My patch currently prevents one from writing specs for imported functions, ie specs must go in the defining module or a `.spec` file. This might be a bit tedious to deal with (it does break a few of our benchmarks, notably vector-algos due to the assumed specs for vector), but I think this is a saner strategy than what we do now, which is to allow any module to refine any function and propagate those specs across imports. What do you think?

2. `tests/pos/VerifiedMonoid` is broken because the `HscNothing` change has caused GHC to not produce a core binder for `$cmempty`, and this breaks our name resolution logic for the class/instance signatures. I'm still trying to figure out why, it seems that our fork of the desugarer is no longer inserting ticks everywhere like it should...

3. A bunch of the reflection benchmarks are broken, with errors like

```
22:16: Error: Illegal type specification for `MonadMaybe.return`

 22 | {-@ axiomatize return @-}
                     ^
     MonadMaybe.return :: lq1:a -> {VV : (Maybe a) | VV == MonadMaybe.return lq1
                                                     && VV == MonadMaybe.Just}
     Sort Error in Refinement: {VV : (MonadMaybe.Maybe a_a27R) | (VV == MonadMaybe.return lq1
                                                                  && VV == MonadMaybe.Just)}
     Cannot unify (MonadMaybe.Maybe a_a27R) with func(0, [@(44); (MonadMaybe.Maybe @(44))]) in expression: VV == MonadMaybe.Just
  because
Invalid Relation VV == MonadMaybe.Just with operand types (MonadMaybe.Maybe a_a27R) and func(1, [@(0); (MonadMaybe.Maybe @(0))])
```

Does this look familiar to you @nikivazou ?

nikivazou:
@gridaphobe should we close this?

gridaphobe:
I think reorganizing the includes to better match the structure of `base` is still worthwhile, but probably not high priority if nobody is actively working on verifying `base` (or using low-level functions from `base` that are not exposed correctly in our includes).

It's been a while though, and this PR has clearly bitrotted somewhat, so it might make sense to close this PR and open an issue instead.

Issue #1061, opened 2017-07-19
Support hexadecimal constants in specifications


sdleffler:
It would be nice to be able to use hex constants in specifications, e.g.

```
{-@ assume chr# :: {n:Int# | 0 <= n && n <= 0x10FFFF} -> Char# @-}
```

instead of

```
{-@ assume chr# :: {n:Int# | 0 <= n && n <= 1114111} -> Char# @-}
```

ranjitjhala:
Good point! Should be quite easy to do IMO [should just have to tweak the parser a touch.]

ranjitjhala:
should be fixed by above unless i've messed up the submodules 😟

PR #1062, opened 2017-07-19
Hex

ranjitjhala:
Allow numeric literals to be in hex (fix for #1061)


atondwal:
should we add an lh test as well as the lf ones?

ranjitjhala:
Merging as the two tests appear to run locally...

PR #1066, opened 2017-07-20
Add CHANGES.md and README.md to extra-source-files

RyanGlScott:
Doing so makes Hackage link directly to the changelog, and renders the README in the Hackage description of the library.


gridaphobe:
Thanks!

PR #1067, opened 2017-07-25
Don't select binders introduced by transform-rec for the decreasing parameter

### LINKED ISSUE #1065: TransformRec makes it impossible to prove termination!
Here's a function @sdleffler is working on in GHC.Exts. A combination of the transform-rec pass and some ugly bugs in LH make it apparently impossible to prove this thing terminating...

```
{-# LANGUAGE TypeFamilies #-}
module GroupBy where

groupByFB :: ([a] -> lst -> lst) -> lst -> (a -> a -> Bool) -> [a] -> lst
groupByFB c n eq xs0 = groupByFBCore xs0
        {- groupByFBCore :: zoo:_ -> _ / [len zoo] @-}
        {- groupByFBCore :: c:_ -> n:_ -> eq:_ -> xs0:_ -> zoo:_ -> _ / [len zoo] @-}
        {- Decrease groupByFBCore 5 @-}
  where groupByFBCore [] = n
        groupByFBCore (x:xs) = c (x:ys) (groupByFBCore zs)
            where (ys, zs) = span (eq x) xs
```

With no annotations, we get a surprising type error.

```
 /Users/gridaphobe/Source/liquid/haskell/GroupBy.hs:10:42-54: Error: Liquid Type Mismatch

 10 |         groupByFBCore (x:xs) = c (x:ys) (groupByFBCore zs)
                                               ^^^^^^^^^^^^^
   Inferred type
     VV : {v : [a] | v == xs0}

     VV : {VV : [a] | VV == xs0 && len VV < len xs0}
```

This is no good, the required type is obviously impossible. Furthermore, what a reference to `xs0` doing inside `groupByFBCore`? It's a binder from the outer `groupByFB`! 

Looking at the core, we see that transform-rec has copied all of the parameters from the outer `groupByFB` to `groupByFBCore`

BEFORE
```
       groupByFBCore =
         \ (ds_dAy :: [a]) ->
```
AFTER
```
       groupByFBCore [Occ=LoopBreaker]
         :: forall a lst ->
            ([a] -> lst -> lst)
            -> lst -> (a -> a -> GHC.Types.Bool) -> [a] -> [a] -> lst
       [LclId, Str=DmdType]
       groupByFBCore =
         \ (@ a)
           (@ lst)
           (c :: [a] -> lst -> lst)
           (n :: lst)
           (eq :: a -> a -> GHC.Types.Bool)
           (xs0 :: [a])
           (ds_dAy :: [a]) ->
...
                 ((groupByFBCore @ a @ lst c n eq xs0) zs)
```

So transform-rec is passing these extra parameters through each recursive call unchanged. I forget what the purpose of this pass is, but fine. 

Now `xs0` is the first parameter with a termination metric, so we try and fail to prove it decreases. Mystery solved! Let's just write an explicit metric

```
        {-@ groupByFBCore :: c:_ -> n:_ -> eq:_ -> xs0:_ -> zoo:_ -> _ / [len zoo] @-}
```

```
 6 |         {-@ groupByFBCore :: c:_ -> n:_ -> eq:_ -> xs0:_ -> zoo:_ -> _ / [len zoo] @-}
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

 The Liquid type

     _ -> _ -> _ -> _ -> _ -> _

 is inconsistent with the Haskell type

     [a] -> lst
```

Hmm, it appears LH is doing the consistency check before the transform-rec pass. Fine, we'll change the signature to

```
        {-@ groupByFBCore :: zoo:_ -> _ / [len zoo] @-}
```

```
 /Users/gridaphobe/Source/liquid/haskell/GroupBy.hs:7:30: Error: Illegal termination specification for `groupByFBCore`

 7 |         {-@ groupByFBCore :: zoo:_ -> _ / [len zoo] @-}
                                  ^
     Termination metric len zoo is ill-sorted
     Cannot unify (@(43)  @(42)) with func(0, [[a_azU]; lst_azV; lst_azV]) in expression: len zoo
```

But now when we try to generate constraints for the termination check, we're suddenly using the post-transform-rec type! Yikes!

My last thought was to try the old `Decrease` annotation.

```
        {-@ Decrease groupByFBCore 5 @-}
```

```
 /Users/gridaphobe/Source/liquid/haskell/GroupBy.hs:8:22: Error: Cannot parse specification:

 8 |         {-@ Decrease groupByFBCore 5 @-}
                          ^
     unexpected "g"
     expecting space or "::"
```

But perhaps this annotation isn't even supported anymore?

gridaphobe:
This is a gross hack, but I think it just might work! See the comment on the new `setIdTRecBound`.

fixes #1065

### DIFF
--- a/src/Language/Haskell/Liquid/Constraint/Generate.hs
+++ b/src/Language/Haskell/Liquid/Constraint/Generate.hs
@@ -74,6 +74,7 @@ import           Language.Haskell.Liquid.Types.Literals
 -- NOPROVER import           Language.Haskell.Liquid.Constraint.Axioms
 import           Language.Haskell.Liquid.Constraint.Types
 import           Language.Haskell.Liquid.Constraint.Constraint
+import           Language.Haskell.Liquid.Transforms.Rec
 
 -- import Debug.Trace (trace)
 
@@ -114,21 +115,21 @@ consAct cfg info = do
 --------------------------------------------------------------------------------
 -- | TERMINATION TYPE ----------------------------------------------------------
 --------------------------------------------------------------------------------
-makeDecrIndex :: (Var, Template SpecType)-> CG [Int]
-makeDecrIndex (x, Assumed t)
-  = do dindex <- makeDecrIndexTy x t
+makeDecrIndex :: (Var, Template SpecType, [Var]) -> CG [Int]
+makeDecrIndex (x, Assumed t, args)
+  = do dindex <- makeDecrIndexTy x t args
        case dindex of
          Left _  -> return []
          Right i -> return i
-makeDecrIndex (x, Asserted t)
-  = do dindex <- makeDecrIndexTy x t
+makeDecrIndex (x, Asserted t, args)
+  = do dindex <- makeDecrIndexTy x t args
        case dindex of
          Left msg -> addWarning msg >> return []
          Right i  -> return i
 makeDecrIndex _ = return []
 
-makeDecrIndexTy :: Var -> SpecType -> CG (Either (TError t) [Int])
-makeDecrIndexTy x t
+makeDecrIndexTy :: Var -> SpecType -> [Var] -> CG (Either (TError t) [Int])
+makeDecrIndexTy x t args
   = do spDecr <- specDecr <$> get
        autosz <- autoSize <$> get
        hint   <- checkHint' autosz (L.lookup x $ spDecr)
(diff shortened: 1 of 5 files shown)


ranjitjhala:
Weird. but seems to work! 

Is there some short example that explains the fix? e.g.

1. Source program
2. Here's why the old thing crashed
3. Here's what the new thing does to avoid the crash?

gridaphobe:
@ranjitjhala #1065 has a fairly minimal example with a long-form explanation of what goes wrong. I've copied that example into the new `tests/pos/T1065.hs` and added a note with a brief explanation to `LHL.Transforms.Rec`.

PR #1068, opened 2017-07-26
ADT

ranjitjhala:
Shuffling / refactoring more stuff to allow for ADTs in the SMT logic.

- There is a single function to make `SMTContext` and declare binders in it,
- This function requires that we have already created an `SInfo` (uniqify, elaborated, etc.)
- `instantiate` (PLE) now runs _after_ `elaborate` which is a hassle because `instantiate` 
   expects refinements are _not_ elaborated.

Merging into `develop` now (assuming circle is ok), as this should be the end 
of the cross-cutting changes, and so @atondwal can fix the weird issues with PLE.


ranjitjhala:
[moving note here from slack]

@nikivazou there are _no tests_ for the “gradual” stuff in fixpoint;
the only failing tests are in LH. It looks like my recent refactoring
has broken some gradual tests (in LH). I am going to disable those
tests to merge ADT into `develop` (once the other stuff passes)
for now.

You can try to restore the gradual thing — its _most likely_
something quite trivial but I don’t know the gradual code so am not sure.

If you add some _small_ .fq tests that work with the _current_
`develop` but break with the _new_ `adt` branch then I can take
a look. (edited)

ranjitjhala:
Matching or https://github.com/ucsd-progsys/liquid-fixpoint/pull/316

PR #1069, opened 2017-07-27
Safer liquid haskell

nikivazou:
Turn on the totality check by default. 

Give false precondition to Haskell's `error` function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,10 +298,7 @@ To use unqualified names, much easier to read, use:
 Totality Check
 --------------
 
-LiquidHaskell can prove the absence of pattern match failures.
-Use the `totality` flag to prove that all defined functions are total.
-
-    liquid --totality test.hs
+LiquidHaskell proves the absence of pattern match failures.
 
 For example, the definition
 
@@ -314,6 +311,10 @@ If we exclude `Nothing` from its domain, for example using the following specifi
     {-@ fromJust :: {v:Maybe a | (isJust v)} -> a @-}
 
 `fromJust` will be safe.
+
+Use the `no-totality` flag to disable totality checking.
+
+    liquid --no-totality test.hs
 
 Termination Check
 -----------------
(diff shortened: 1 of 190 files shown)


gridaphobe:
I support this in general, but before turning on --totality by default we really need to improve the error messages.

```haskell
module Head where
{-@ LIQUID "--totality" @-}

head z = case z of
           (x:xs) -> x
```
```
 /Users/gridaphobe/Source/liquid/haskell/Head.hs:(4,1)-(5,22): Error: Liquid Type Mismatch
   Inferred type
     VV : {v : Addr# | v == "/Users/gridaphobe/Source/liquid/haskell/Head.hs:(4,10)-(5,22)|case"
                       && v == ?a}

     VV : {VV : Addr# | false}

   In Context
     ?a : {?a : Addr# | ?a == "/Users/gridaphobe/Source/liquid/haskell/Head.hs:(4,10)-(5,22)|case"}
```

Notice how

1. The error blames the entire definition rather than the case expression.
2. The error message talks about this weird `Addr#` thing that I didn't write.

Ideally we could give some description of what kinds of values are not covered by the patterns (can we synthesize a refinement type that describes them?), but we could at least make the error message something more clear like "this pattern is not exhaustive".

nikivazou:
Sure, but this is not easy (since LH does not know at all that it is proving totality). 

I think we should first make totality default (esp since there are modifications all over the place) and improve errors after.

ranjitjhala:
I think I agree with @nikvazou on this.

ranjitjhala:
@nikivazou can you add @gridaphobe s example as the kind of totality error
in the README?

Not as good as a proper message but a lot better than nothing.

nikivazou:
Let me create an issue instead.

gridaphobe:
Can you do both? :) The reason to add it to the readme is that users are more likely to look at that than the issues list..

Issue #1078, opened 2017-08-27
Refinements in `include/Data/ByteString/Char8.spec` not working


Gabriel439:
None of the refinements from `include/Data/ByteString/Char8.spec` appear to work.  For example, this program passes the `liquid` type checker when it shouldn't:

```haskell
import qualified Data.ByteString.Char8

example :: Char
example = Data.ByteString.Char8.head Data.ByteString.Char8.empty
```

I suspect the reason is that the `bslen` measure is not in scope in the  `include/Data/ByteString/Char8.spec`, so the type-checker silently ignores all of the refinements in that file.  The `bslen` measure is actually defined in `include/Data/ByteString.spec`

Assuming that my guess is correct, it's not clear how to resolve this:

Should I duplicate the `bslen` measure definition from `include/Data/ByteString/Char8.spec` to `include/Data/ByteString.spec`?  I don't know if that would cause problems if you were to mix functions from both modules

Should I move all the refinements from `include/Data/ByteString/Char8.spec` to `include/Data/ByteString.spec` so that they can reuse the same `bslen` measure?  I'm not sure if it matters what file you define Liquid Haskell `.spec` includes in

Or is there some way to import the `bslen` measure from `include/Data/ByteString.spec` to `include/Data/ByteString/Char8.spec`?

ranjitjhala:
Oh no, this is terrible.

It shouldn't be what you're saying because those
specs should be rejected as ill-sorted
(as `bslen` is not in scope).

Let me take a look! (And thanks for pointing this out!)

ranjitjhala:
Should be fixed by #1079

Wow, the bug was in ... the `liquidhaskell.cabal` file! For some reason, the `data-files` field was missing: `, include/Data/ByteString/*.spec` and so that was not even copied over to the relevant `dist` directory and hence was just being ignored...

I have added the above line to the .cabal file. 

Once you do that, on your original source you get the error:

```
 /Users/rjhala/research/stack/liquidhaskell/.stack-work/install/x86_64-osx/lts-8.9/8.0.2/share/x86_64-osx-ghc-8.0.2/liquidhaskell-0.8.0.0/include/Data/ByteString/Char8.spec:(3,17)-(5,0): Error: Illegal type specification for `assumed type Data.ByteString.empty`
     assumed type Data.ByteString.empty :: {bs : ByteString | bslen bs == 0}
[...]

PR #1079, opened 2017-08-27
T1078

ranjitjhala:
Fix for #1078

### DIFF
--- a/include/Data/ByteString/Char8.spec
+++ b/include/Data/ByteString/Char8.spec
@@ -1,5 +1,7 @@
 module spec Data.ByteString.Char8 where
 
+import Data.ByteString 
+
 assume empty :: { bs : Data.ByteString.ByteString | bslen bs == 0 }
 
 assume singleton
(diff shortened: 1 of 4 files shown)


include/Data/ByteString/Char8.spec
```diff
@@ -398,3 +391,10 @@ assume hGetNonBlocking
     :: System.IO.Handle
     -> n : { n : Int | 0 <= n }
     -> IO { bs : Data.ByteString.ByteString | bslen bs <= n }
+
+// assume partition
```
ranjitjhala: Apparently `partition` is no longer exported by `D.B.Char8`?

ranjitjhala:
Thanks! Btw I should have updated the comment, I rigged it so that the Char8.spec "imports" bytestring.spec so your original example should work as is i.e. Without the extra import e.g. 

See https://github.com/ucsd-progsys/liquidhaskell/blob/d133c6fad7279d69d6a6f88718c3d8648583a1d8/tests/neg/T1078.hs

PR #1081, opened 2017-09-03
Add support for `Data.String.fromString`

### LINKED ISSUE #822: Reasoning about length of concrete bytestrings
For example the [following](http://goto.ucsd.edu:8090/index.html#?demo=permalink%2F1472583424_14895.hs) fails 

```
{-@ bar :: {v:ByteString | bslen v == 1 } @-}
bar = " "
```

With intermediate code 

```
Main.bar =
   (Data.String.fromString
      @ Data.ByteString.Internal.ByteString
      Data.ByteString.Internal.$fIsStringByteString)
     (GHC.CString.unpackCString# " "#)]
```

Gabriel439:
Fixes #822

Based on @nikivazou's suggestion

This refines the output of `fromString` to add a new `stringlen` measure
and then adds invariants to the various `ByteString` types to
synchronize this measure with their existing length measures

Testing this also identified some refined types for functions which no
longer exist in more recent versions of the `bytestring` package, so I
removed from the specification files

### DIFF
--- a/include/Data/ByteString.spec
+++ b/include/Data/ByteString.spec
@@ -1,8 +1,12 @@
 module spec Data.ByteString where
 
+import Data.String
+
 measure bslen :: Data.ByteString.ByteString -> { n : Int | 0 <= n }
 
 invariant { bs : Data.ByteString.ByteString  | 0 <= bslen bs }
+
+invariant { bs : Data.ByteString.ByteString | bslen bs == stringlen bs }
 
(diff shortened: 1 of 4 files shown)


ranjitjhala:
Awesome, thanks! I will fix whatever fiddly thing is happening with our
regression test and merge this afternoon ...

Gabriel439:
You're welcome! :)

Issue #1049, opened 2017-06-15
Annotations are ignored when comment is not closed with @-}


nikivazou:
The following is SAFE 

```
{-@ incr :: x:Int -> {v:Int | x < v } -}
incr :: Int -> Int 
incr x = x - 1
```
Because the annotation is not closing with `@`. 
Maybe we need to create a warning in this case?

ranjitjhala:
Whoa! Are the annotations even parsed? Actually I think not, because the parser literally sucks out text **between** `{-@ ... @-}` before doing any parsing, so if you don't close, then ... that text is just empty... :-/

nikivazou:
Exactly, the annotations are not parsed. 

I just got an email from a user wondering what is wrong with an annotation like this and took me a log time to spot the typo. So, maybe a warning could be helpful.

ranjitjhala:
Hmm. Maybe tweak the "between" parser to warn if there is an "open" `{-@` ? (should be easy...)

gridaphobe:
GHC throws an error if you have an improperly closed pragma, e.g.

```
{-# LANGUAGE Blah -}
```

```
    /Users/gridaphobe/Source/liquid/haskell/src/Language/Haskell/Liquid/GHC/Interface.hs:1:14: error:
        Cannot parse LANGUAGE pragma
        Expecting comma-separated list of language options,
        each starting with a capital letter
          E.g. {-# LANGUAGE TemplateHaskell, GADTs #-}
```

I think this is a good idea (but we can do better with the error message)!

PR #1082, opened 2017-09-05
Catch unclosed annotation

Abhiroop:
Trying to fix https://github.com/ucsd-progsys/liquidhaskell/issues/1049

I initially thought of changing the `extractSpecComment` function of https://github.com/ucsd-progsys/liquidhaskell/blob/develop/src/Language/Haskell/Liquid/GHC/Interface.hs#L453 to 

```haskell
extractSpecComment :: GHC.Located AnnotationComment -> Either String (SourcePos, String)
extractSpecComment (GHC.L span (AnnBlockComment text)) =
  case (length text > 2,isPrefixOf "{-@" text,isSuffixOf "@-}" text) of
        (True,True,True)  -> Right (offsetPos, take (length text - 6) $ drop 3 text)
        (True,False,True) -> Left "Missed the prefix {-@"
        (True,True,False) -> Left "Missed the suffix @-}"
        (_,_,_) -> Left ""
  where offsetPos = incSourceColumn (srcSpanSourcePos span) 3
extractSpecComment _ = Left ""
```

and extracting that error info, but that would involve changing some major type signatures. So I tried something simple like this. This works but I am not able to figure out the line no info currently. If someone can guide me on that, I can try adding the line no info (and perhaps a more helpful message) to this PR. Thanks.


ranjitjhala:
Thanks very much @Abhiroop, much appreciated!  

Looks like there are some failing tests, let me take 
a look and see why and if I can also get the source 
locations working.

ranjitjhala:
Hi @Abhiroop, thanks again! I found that it was easiest to 

1. get the `SourceSpan` of the bad comment, and
2. simultaneously check whether a comment was bad

at the point where `hsSpecificationP` is called, or rather,
at the function which extracts the `[String]` that are then 
_sent to_ `hsSpecificationP`, namely in `extractSpecComments`.

That addressed both the missing location, and also ensures 
that the other tests work.

Thanks a bunch for this!!!

Abhiroop:
I was already keeping an eye on your changes and noticed the changes you made to `extractSpecComment`. I wanted something similar in the design. Glad to help. :)

PR #1087, opened 2017-09-14
T326

### LINKED ISSUE #1086: Reflection fails with type-aliases in lambdas
The following program

```haskell
{-# LANGUAGE ScopedTypeVariables #-}

module Bob where

type Val = Int

{-@ reflect poo @-}
poo :: Int -> Int
poo y = (\(x :: Val) -> x + 1) y
```

fails with

```

 /Users/rjhala/research/stack/liquidhaskell/tests/todo/alias.hs:8:13: Error: Illegal type specification for `replaceLocalBindsOne Bob.poo`
  
 8 | {-@ reflect poo @-}
                 ^
  
     replaceLocalBindsOne Bob.poo :: lq1:Int -> {VV : Int | VV == Bob.poo lq1
                                                            && VV == lam ds_dHq : Bob.Val . ds_dHq + 1 lq1}
     Sort Error in Refinement: {VV : int | (VV == Bob.poo lq1
                                            && VV == lam ds_dHq : Bob.Val . ds_dHq + 1 lq1)}
     Operands have different types Bob.Val and int in ds_dHq + 1
```

because the reflected signature binds the lambda to something of type `Val` instead of expanding the alias and treating it as `Int`...

ranjitjhala:
Fix #1086 

Looks like we'd forgotten to `expandTypeSynonyms`

### DIFF
--- a/src/Language/Haskell/Liquid/Bare.hs
+++ b/src/Language/Haskell/Liquid/Bare.hs
@@ -836,7 +836,7 @@ replaceLocalBindsOne allowHO v
                              env' (zip ty_binds ty_args)
            let res  = substa (f env) ty_res
            let t'   = fromRTypeRep $ t { ty_args = args, ty_res = res }
-           let msg  = ErrTySpec (GM.sourcePosSrcSpan l) ({- text "replaceLocalBindsOne" <+> -} pprint v) t'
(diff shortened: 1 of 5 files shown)


ranjitjhala:
Also fixes https://github.com/ucsd-progsys/liquid-fixpoint/issues/326

PR #1088, opened 2017-09-15
Remove tests that depend on proofmethod=arithmetic

atondwal:
Preparing to remove that proofmethod from fixpoint


atondwal:
updated to match fixpoint#328

ranjitjhala:
@atondwal looks like you need to update LF, this build is broken.

atondwal:
Okay, that should've fixed it

PR #1106, opened 2017-09-30
imported reflected HO functions

nikivazou:
Before this edit we could not prove the below theorem (see [here](https://github.com/ucsd-progsys/liquidhaskell/blob/42fa24a8a2ace7b6464007b8b13da5f77769a046/tests/theorem_proving/T1106.hs)) when `foo` is imported. 

```
mapFoo (foo c) (Foo b) = Foo (foo c b)
```

The reason? The type of `foo c` is true while it should be `{v: a -> b | v == foo c}`

When higher-order flag is on two kinds of functions were given singleton/functional types (like `v == foo c`)

  - local fuctional variable (as stored in `fargs γ`)
  - reflected functions (as stored in `aenv γ`)

The problem: imported reflected functions are not added in `aenv`

This branch solves this problem by giving singleton functional types to all function applications, when the HO flag is on. 
But, travis rejects this solution since it can give internal unsorted refinements. 

Another solution would be to add imported reflected functions to `aenv` . 
@ranjitjhala where can I find the names for all the imported reflected functions??? 

@jprider63 this PR will solve your problem once merged.

### DIFF
--- a/src/Language/Haskell/Liquid/Constraint/Generate.hs
+++ b/src/Language/Haskell/Liquid/Constraint/Generate.hs
@@ -1289,8 +1289,25 @@ makeSingleton γ e t
   = t
 
 funExpr :: CGEnv -> CoreExpr -> Maybe F.Expr
+
+{- OLD: only sinlgetons for 
+-- reflected functions as appear in aenv
+-- this breaks now as imported relfected functions
+-- do no live in aenv
+-- NV to RJ: to restore this code add imported reflected
+-- functions to aenv 
 funExpr γ (Var v) | M.member v $ aenv γ
   = F.EVar <$> (M.lookup v $ aenv γ)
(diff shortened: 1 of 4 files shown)


ranjitjhala:
I couldn't tell from the patch, does this affect all the code or only when
the higher order flags on ?

ranjitjhala:
Either way it's fine!

nikivazou:
Only when the HO flag is on, but it is not fine... 

I have some failing tests because it inserts in the refinements functions that are not encoded in the logic, leading to sort check error. 

So, is there an easy way to add all the imported reflected functions in `aenv`?

ranjitjhala:
Sure ... just find the imported functions (look for gsReflects in GhcSpec)
and add them to the aenv?

PR #1107, opened 2017-10-03
Wide Help 

themattchan:
So that the descriptions don't get wrapped into teeny narrow columns.

PS: sorry for the this being so hacky, the `CmdArgs` library doesn't allow tweaking the format of its generated help menu...

Before:
screen shot 2017-10-02 at 6 34 54 pm

After:
screen shot 2017-10-02 at 6 34 46 pm


ranjitjhala:
Wow this is awesome, thanks @themattchan ! 

Is it possible to make the wide version the "default" ? (seems desirable?)

themattchan:
@ranjitjhala this change makes it the default!

ranjitjhala:
Oh, sweet! thanks!

PR #1129, opened 2017-10-13
Bump aeson version

phadej:
Note: `AnnErrors` seems to be dead code, should it be removed?

### DIFF
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -121,7 +121,7 @@ Library
                 , unordered-containers >= 0.2
                 , liquid-fixpoint >= 0.7.0.5
                 , located-base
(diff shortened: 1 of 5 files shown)


liquidhaskell.cabal
```diff
@@ -309,7 +309,7 @@ test-suite liquidhaskell-parser
   if flag(devel)
     hs-source-dirs:   tests src
     build-depends:
-                  aeson >= 0.10 && < 1.0
```
phadej: We inherit bounds from the `library`.

ranjitjhala:
This is terrific, thanks @phadej ! (Yes, thats odd that its still there... I can remove it the next time I'm there, thanks for pointing it out!

Issue #1122, opened 2017-10-09
Desugarer breaks (again) with GHC 8.2.1


nikivazou:
https://twitter.com/LucaCiciriello/status/917299709013880837

ranjitjhala:
Grr.

PR #1131, opened 2017-10-14
WIP: ghc-8.2

phadej:
Inspired by @Gabriel439 I made a bulk of porting liquidhaskell to GHC-8.2. At the current state it compiles, but I had to sneak few `undefined`s here and there, so this doesn't really work.

It seems that desugarer is copy-pasted from GHC (why? are there some changes). What I did:
- made a diff between `liquidhaskell` and GHC's `ghc-8.0.2-release
- copied over `ghc-8.2.1-release` `deSugar` files
- applied the patch
- went over thru conflicts

The rest of the patch is related mostly with these three commits in GHC:
- Re-add FunTy (big patch)
  https://github.com/ghc/ghc/commit/77bb09270c70455bbd547470c4e995707d19f37d
- Major patch to introduce TyConBinder
  https://github.com/ghc/ghc/commit/e368f3265b80aeb337fbac3f6a70ee54ab14edfd
- s/Invisible/Inferred/g s/Visible/Required/g
  https://github.com/ghc/ghc/commit/5fdb854cbad734ed8113ea23485d834156b49df1

I'm not sure if the changes are sound (I only type-checked the code, didn't run it).

If someone wants to advise me to complete the journey, or even better pick up from here that would be super super great.

### DIFF
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -263,8 +263,8 @@ test-suite test
   type:              exitcode-stdio-1.0
   hs-source-dirs:    tests
   ghc-options:       -W -threaded
-  if flag(devel)
-    ghc-options: -Werror
+--  if flag(devel)
+--    ghc-options: -Werror
   main-is:           test.hs
   build-depends:     base >=4.8.1.0 && <5
                ,     containers >= 0.5
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -254,8 +254,8 @@ Library
    ghc-options: -W -fwarn-missing-signatures
    if flag(include)
      hs-source-dirs: devel
-   if flag(devel)
-     ghc-options: -Werror
+   --if flag(devel)
+   --  ghc-options: -Werror
    Default-Extensions: PatternGuards
 
 test-suite test
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -67,7 +67,7 @@ Executable liquid
   default-language: Haskell98
   Build-Depends: base >=4.8.1.0 && <5
                , ghc
-               , ghc-boot == 8.0.2
+               , ghc-boot
                , cmdargs
                , time
                , deepseq
@@ -94,8 +94,8 @@ executable target
 Library
    Default-Language: Haskell98
    Build-Depends: base >=4.8.1.0 && <5
-                , ghc == 8.0.2
-                , ghc-boot == 8.0.2
+                , ghc == 8.2.1
+                , ghc-boot == 8.2.1
                 , template-haskell >= 2.9
                 , time >= 1.4
                 , array >= 0.5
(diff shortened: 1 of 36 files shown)


nikivazou:
That is great!!!! Thank you both @phadej and @Gabriel439!!! 

That would close https://github.com/ucsd-progsys/liquidhaskell/issues/1122 


You would need to do something in the lines of https://github.com/ucsd-progsys/liquidhaskell/pull/1016/files  to update the desugarer with expression locations and preserve local liquid error messages

phadej:
now I get to run `liquid`, but it gives me errors like (obviously I do something wrong):

```
% /liquid ./tests/pos/AdtPeano1.hs
LiquidHaskell Copyright 2013-17 Regents of the University of California. All Rights Reserved.


**** DONE:  A-Normalization ****************************************************
 

**** DONE:  annotate ***********************************************************
 

**** RESULT: ERROR *************************************************************


 /home/ogre/.cabal/share/x86_64-linux-ghc-8.2.1/liquidhaskell-0.8.0.5/include/Data/Maybe.spec:3:9-14: Error: Illegal type specification for `measure isJust`
  
 3 | measure isJust :: forall a. Data.Maybe.Maybe a -> Bool

....

 Fixpoint.Types.dummyLoc:0:0: Error: Illegal type specification for `measure lqdc##$select##GHC.Tuple.(,,,)##2`
     measure lqdc##$select##GHC.Tuple.(,,,)##2 :: (a, b, c, d) -> b
     Type constructor (,,,) expects a maximum 0 arguments but was given 4 arguments

```

phadej:
Maybe it's related to me removing `sortDiff` line in `liftableTyCons` in `src/Language/Haskell/Liquid/Bare/Measure.hs`.

nikivazou:
Thats super weird: why `(,,,)` does not expect any arguments???

I suspect something should change on how we computer `realTcArity` called here https://github.com/ucsd-progsys/liquidhaskell/blob/develop/src/Language/Haskell/Liquid/Bare/Check.hs#L320

nikivazou:
I mean `realTcArity` of `(,,,)` should be 4. This makes sense, right?

phadej:
I pushed a commit removing last `undefined` I introduced, so the incorrect behaviour isn't caused by lazy programmer, but maybe incompetent one.

I'm quite sure that I introduced some silly error when doing `FunTy` & `TyConBinder` refactors. I hope that new eyes could spot what's wrong. (look for diffs *not* in `Desugarer`).

Also another ghc-commit reference for `anormalize` change. Ben refactored `initDs` in https://github.com/ghc/ghc/commit/4bd2232754f2c0b365177d177e6d749c64e73176

Now it's simpler to use (maybe?)

nikivazou:
What? They switched back into `FunTy`? OK, I will try to take a look ASAP.

phadej:
@nikivazou yes, I tried to link to the relevant commits in the OP.

nikivazou:
Oh, ghc did not change anything. I made a `ForAllTy` pattern to minimize my edits when porting to ghc-8. It is good that you are removing it!

nikivazou:
@phadej I moved your edits here https://github.com/ucsd-progsys/liquidhaskell/pull/1136 so that I have edit access too. 

AdtPeano1 is now SAFE!!!! Let's see what else breaks.

Issue #1132, opened 2017-10-14
Commenting out an unused function breaks verification


peti:
liquidhaskell-0.5.0.1-3330-g8c6f55f5 (current git HEAD) verifies the following program just fine:

~~~
import Data.Time.Calendar

{-@ type MonthNum = { i:Int | 0 < i && i <= 12 } @-}
type MonthNum = Int

{-@ assume toGregorian :: Day -> (Integer, MonthNum, Int) @-}

{-@ showMonthNum :: MonthNum -> String @-}
showMonthNum :: MonthNum -> String
showMonthNum i = monthNames !! (i-1)

showMonth :: Day -> String
showMonth d = monthNames !! (m-1) where (_,m,_) = toGregorian d

{-@ monthNames :: { mn:[String] | len mn == 12  } @-}
monthNames :: [String]
monthNames = ["Jan","Feb","Mar","Apr","May","Jun","Jul","Aug","Sep","Oct","Nov","Dec"]
~~~

Now, when I remove the definition of `showMonthNum` -- which is not used anywhere in that code --, then the tool fails, complaining:

~~~
 13 | showMonth d = monthNames !! (m-1) where (_,m,_) = toGregorian d
                                   ^^^
  
   Inferred type
     VV : {v : Int | v == m - ?a
                     && v == ?b}
  
   not a subtype of Required type
     VV : {VV : Int | 0 <= VV
                      && VV < len Main.monthNames}
  
   In Context
     m : {m : Int | m /= 0
                    && m >= 0
                    && m > 0}
      
     Main.monthNames : {Main.monthNames : [[Char]] | len Main.monthNames == 12
                                                     && len Main.monthNames >= 0}
      
     ?b : {?b : Int | ?b == m - ?a}
      
     ?a : {?a : Int | ?a == (1 : int)}
~~~

That doesn't seem right?

nikivazou:
Interesting! It seems that commenting out `showMonthNum` totally ignores your `MonthNum` definition.... 

I am looking at it now

BTW, I suggest you use develop instead of master, there are a lot of fixes there. We work on develop instead of master and we just forget to update master...

peti:
I switched to `develop` now, too, just in case. Is there any particular reason to prefer that branch over `master`?

ranjitjhala:
Hi @peti and @nikivazou, this looks like its fixed in `develop` ? Can you confirm? Thanks! 

[...]

PR #1134, opened 2017-10-14
get quals from all used vars: fix for #1132

nikivazou:


ranjitjhala:
Thanks @nikivazou!

gridaphobe:
Be VERY careful with this. I tried enabling this same scraping from imported variables a while ago and it had disastrous performance consequences. I never fully tracked it down, but suspect it's due to some highly polymorphic qualifiers somewhere in the include files.

I think I scraped **all** imported variables rather than just the **used** variables though...

ranjitjhala:
Ah yes ... that was still tucked into an option IIRC?

gridaphobe:
Looks so :)

https://github.com/ucsd-progsys/liquidhaskell/blob/develop/src/Language/Haskell/Liquid/UX/CmdLine.hs#L263-L276

PR #1141, opened 2017-10-21
Gradual

nikivazou:
This adds the `gradual` executable. Only minor edits in the Liquid Haskell code (3 exports). 

Minor edits, but better merge sooner than later :)

I could just port gradual in a different repo, but for now it is more convenient to have it here, I can move it when I am done?


ranjitjhala:
Sounds good!

PR #1143, opened 2017-10-23
Add back popl tests

atondwal:
Not sure why Ackermann hangs now, but the rest work again.


ranjitjhala:
can you try ackermann with --no-adt?

atondwal:
sorry, just saw this response: yeah, I tried it with `--no-adt` and no dice

atondwal:
oh, I think Ackermann.hs was actually using fuel instead of PLE, and it might just be that PLE is very very slow when it comes to evaluating Ackermann (thanks to the exponential number of disjunctions we throw at z3 at each step of evaluation of the Ackermann function when we check to see if we should continue).

PR #1160, opened 2017-11-07
Lower-casing the `lazy` keyword

bartavelle:
Apparently this should be lower case to work. Perhaps it is the same for the `Decrease` example, but I have not tested it yet.


ranjitjhala:
Oops you're right! Thanks @bartavelle !!

PR #1183, opened 2017-12-06
Remove zip3 from split

nikivazou:


ranjitjhala:
Hmm. looks like it breaks some tests...

ranjitjhala:
looks like this doesn't work. shall i keep it open or close?

PR #1200, opened 2017-12-18
totality checking hack error

nikivazou:


ranjitjhala:
“totality” has one “l” not two.

Btw what we really need is a GHC style message that lists at least one
missing pattern...

I spent many hours trying to do a proof because I couldn’t tell which case
was missing (and there were only 4 needed cases out of many
possibilities...)

ranjitjhala:
why did you replace `false` with `5 < 4` btw?

nikivazou:
I replaced false with a "totality" specific checl "5<4" so that I can syntactically detect it in the generated subtyping constraints and replace it with the totality failure error. 

You are correct, that we could figure out exactly which case is missing (is generating the error), but it is not trivial to present it to the user...

ranjitjhala:
Yes that's what I guessed...

PR #1207, opened 2017-12-21
binah tests

jbrown215:
Adds binah tests to pos which should pass and neg which should fail


ranjitjhala:
Fantastic, many thanks @jmbrown215!!

PR #1209, opened 2017-12-26
Adding information that is saving hours of newcomers

varosi:
I have tried to build a few SMT solvers, like Z3 and CVC. It's not trivial under Windows 10 on a non-clean installations. Then I found ready to install Z3 binary from Microsoft at last. But for some reason it doesn't add its *bin* folder to PATH. And I had first to figure out how LiquidHaskell interact with SMT solvers (I'm a newcomer). With reading the code I found that it spawns child process of solvers that should be found on PATH. If I have this written in instructions it would save me hours of time and hours invested in learning LiquidHaskell instead. So I propose those changes to the manual.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -10,10 +10,11 @@ To run `liquid` you need to install:
 
 Download and install *at least one* of
 
-+ [Z3](https://github.com/Z3Prover/z3/releases)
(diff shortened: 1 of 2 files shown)


ranjitjhala:
This is great, many thanks!!

Issue #1215, opened 2017-12-28
Shorter file path for the only file (from user's stand point) in the Web demo


varosi:
Error messages will be a bit more readable in the Web demo if file paths are replaced with "demo.hs" or something short like this. Even line numbers and positions are not needed in the hint text as they are already selected in the text.

![image](https://user-images.githubusercontent.com/6051718/34420510-a9b781a8-ec12-11e7-89a1-559a418a804d.png)

ranjitjhala:
Thanks for the suggestion! Its implemented on the online demo now:

<img width="331" alt="screen shot 2017-12-28 at 2 15 15 pm" src="https://user-images.githubusercontent.com/1650232/34424590-a29b799e-ebd9-11e7-9892-188e01b69e88.png">

varosi:
Thanks! Looks very nice and clean!

PR #1216, opened 2017-12-28
GHC 8.2.2

### LINKED ISSUE #1211: Compilation fails on GHC 8.2.2
It seems it's due to some minor changes in the GHC API: for example, the type of `getRuntimeRep` has changed, which causes the `DsUtils` module to break.

ranjitjhala:
Merging in [Luca Ciciriello's](https://twitter.com/lucaciciriello?lang=en) edits to the desugarer that 
let us upgrade to GHC 8.2.2. 

Fixes #1211

### DIFF
--- a/AwakeTODO.md
+++ b/AwakeTODO.md
@@ -1,6 +1,0 @@
-- see the Category issue
-- set-up atom
-- see Jeff's thread on Monoids
-- write a prefix checker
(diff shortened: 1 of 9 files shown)


ranjitjhala:
Also drop filepath from JSON output; fixes #1215

PR #1222, opened 2018-01-19
Include commit count (and revision) in version info

waddlaw:
- [x] Remove the `version` magic number
- [x] Add commit count (and revision) in version info

Now, `liquid` says:

```bash
$ stack exec -- liquid
LiquidHaskell Version 0.8.2.2, Git revision bc9c3c64cf2be0a65ad7bfc1804152bd5a7f7cfc (8377 commits) Copyright 2013-17 Regents of the University of California. All Rights Reserved.
```

Currently implementation was change `Language.Haskell.Liquid.UX.CmdLine` for the better, but I think that it should be implemented `src/Liquid.hs`, if as possible.
Because, `TH` is expanded at compile time, and then `library` is not always compiled.

## references

- [Main.hs in stack](https://github.com/commercialhaskell/stack/blob/master/src/main/Main.hs#L122)
- [optparse-simple pkg](https://www.stackage.org/haddock/lts-10.3/optparse-simple-0.1.0/Options-Applicative-Simple.html#)
- [gitrev pkg](https://www.stackage.org/lts-10.3/package/gitrev-1.3.1)

### DIFF
--- a/devel/Paths_liquidhaskell.hs
+++ b/devel/Paths_liquidhaskell.hs
@@ -1,12 +1,18 @@
 {-# LANGUAGE TemplateHaskell #-}
+
 module Paths_liquidhaskell where
 
 import Language.Haskell.TH
 import System.Directory
 import System.FilePath
+import Data.Version (Version, makeVersion)
 
 getDataFileName :: FilePath -> IO FilePath
 getDataFileName f = do
   let loc = $(do { loc <- location; f <- runIO (canonicalizePath (loc_filename loc)); litE (stringL f); })
   let root = takeDirectory (takeDirectory loc)
   return (root </> f)
+
+-- | dummy version (devel only)
+version :: Version
+version = makeVersion [0,0,0,0]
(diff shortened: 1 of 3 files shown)


ranjitjhala:
Thanks so much! (Waiting to see if this works on travis ci too...)

ranjitjhala:
Unfortunately the build breaks on travis with this strange error:

screen shot 2018-01-18 at 10 49 39 pm

I had earlier tried to use the `version` field myself for the command 
line and it worked fine locally but for some strange reason it broke 
TRAVIS in the same way.

Does anyone (@waddlaw ? or @gridaphobe ) know what the problem may be?
(i.e. why does the `Paths_liquidhaskell` on Travis not export these fields?)

waddlaw:
@ranjitjhala Passed to CircleCI check.
Changes:
- using the dummy version number when if use devel flag.

ranjitjhala:
Thanks so much for figuring this out!!!

PR #1223, opened 2018-01-19
use Haskell defs into refinements

nikivazou:


ranjitjhala:
Please add a test case that shows why this Change is added; i.e. that fails currently but succeeds with this PR.

PR #1227, opened 2018-01-29
Delete duplicate `build-type` stanza in .cabal file

RyanGlScott:
`cabal-install-2.1` warns you about this:

```
The field "build-type" is specified more than once at positions 4:1, 13:1
```


ranjitjhala:
Thanks @RyanGlScott !!

Issue #1170, opened 2017-11-20
Use `len` instead of `vlen` for `Vector`?


ranjitjhala:
(Via slack channel, user @waddlaw)

```{-@ LIQUID "--no-termination" @-}

import Data.Vector (Vector, (!))

{-@ vectorSum :: {v:Vector Int | 0 <= vlen v} -> Int @-}
vectorSum :: Vector Int -> Int
vectorSum vec = go 0 0
  where
    go acc i
      | i == sz = acc
      | i < sz = go (acc + (vec ! i)) (i + 1)
    sz = length vec
```

gives the error 

```$ liquid ex1.hs
ex1.hs:12:29-35: Error: Liquid Type Mismatch

 12 |       | i < sz = go (acc + (vec ! i)) (i + 1)
                                  ^^^^^^^

   Inferred type
     VV : {v : Int | v <= sz
                     && v >= 0
                     && v == i}

   not a subtype of Required type
     VV : {VV : Int | VV >= 0
                      && VV < vlen vec}

   In Context
     sz : {sz : Int | sz >= 0
                      && sz == len vec}

     vec : {vec : (Vector Int) | 0 <= vlen vec}

     i : {i : Int | i <= sz
                    && i >= 0}
```

This is an odd consequence of the `Traversable` instance for `Vector`.

The following modification of your `vectorSum` is safe (add the following two lines)

```
import Prelude hiding (length)
import Data.Vector (length,Vector, (!))
```

What’s happening is without the above, the `length` 
refers to the generic `length` from `Traversable` (or 
whatever the class is, I forget now), and specifically, 
you see that in your error message you get:

```   In Context
     sz : {sz : Int | sz >= 0
                      && sz == len vec}
```

That is, the binder `sz`  that was assigned `length vec` has the type `sz = len vec`. However, for Vector bounds checking, you need `sz = vlen vec` since we use `vlen` as the size of the Vector!

Thats why if you instead use `sz = Vector.length vec` then everything works out.

That said — all of this points to something we should _fix_ in the LH prelude, namely instead of using `vlen` we should just update the specifications for `Vector` to use `len` so as to be compatible with the `Traversable` thing…! [I will make an issue on GH]

PR #1230, opened 2018-02-01
[WIP] Allow len for vector measure.

waddlaw:
See #1170

For backward-compatibility, I have left `vlen`.

```haskell
assume ! :: forall a. x:(Data.Vector.Vector a) -> vec:{v:Nat | v < vlen x || v < len x } -> a
```

Is this correct? And, What should I do about the test case?

### DIFF
--- a/include/Data/Vector.spec
+++ b/include/Data/Vector.spec
@@ -4,21 +4,20 @@ import GHC.Base
 
 data variance Data.Vector.Vector covariant
 
-
 measure vlen    :: forall a. (Data.Vector.Vector a) -> Int
 
-invariant       {v: Data.Vector.Vector a | 0 <= vlen v } 
+invariant       {v: Data.Vector.Vector a | 0 <= vlen v && 0 <= len v }
 
-assume !           :: forall a. x:(Data.Vector.Vector a) -> vec:{v:Nat | v < vlen x } -> a 
+assume !           :: forall a. x:(Data.Vector.Vector a) -> vec:{ v:Nat | v < vlen x && v < len x } -> a
 
-assume unsafeIndex :: forall a. x:(Data.Vector.Vector a) -> vec:{v:Nat | v < vlen x } -> a 
+assume unsafeIndex :: forall a. x:(Data.Vector.Vector a) -> vec:{ v:Nat | v < vlen x && v < len x } -> a
 
-assume fromList  :: forall a. x:[a] -> {v: Data.Vector.Vector a  | vlen v = len x }
+assume fromList    :: forall a. x:[a] -> { v: Data.Vector.Vector a  | vlen v = len x && len v = len x }
 
-assume length    :: forall a. x:(Data.Vector.Vector a) -> {v : Nat | v = vlen x }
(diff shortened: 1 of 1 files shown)


include/Data/Vector.spec
```diff
@@ -7,18 +7,18 @@ data variance Data.Vector.Vector covariant
 
 measure vlen    :: forall a. (Data.Vector.Vector a) -> Int
 
-invariant       {v: Data.Vector.Vector a | 0 <= vlen v } 
+invariant       {v: Data.Vector.Vector a | 0 <= vlen v || 0 <= len v }
```
ranjitjhala: Make this `&&` instead of `||` because both facts are true.

include/Data/Vector.spec
```diff
@@ -7,18 +7,18 @@ data variance Data.Vector.Vector covariant
...
 
-assume !           :: forall a. x:(Data.Vector.Vector a) -> vec:{v:Nat | v < vlen x } -> a 
+assume !           :: forall a. x:(Data.Vector.Vector a) -> vec:{v:Nat | v < vlen x || v < len x } -> a
 
-assume unsafeIndex :: forall a. x:(Data.Vector.Vector a) -> vec:{v:Nat | v < vlen x } -> a 
+assume unsafeIndex :: forall a. x:(Data.Vector.Vector a) -> vec:{v:Nat | v < vlen x || v < len x } -> a
 
-assume fromList  :: forall a. x:[a] -> {v: Data.Vector.Vector a  | vlen v = len x }
+assume fromList  :: forall a. x:[a] -> {v: Data.Vector.Vector a  | vlen v = len x || len v = len x }
```
ranjitjhala: Similarly, `fromList`, `length`, `replicate`, `imap` and `map`
should also have `&&` (instead of `II`) in the output.

ranjitjhala:
Hi @waddlaw -- I left some comments above (haven't used this feature on Github so apologies in advance!) can you see it? Basically I think in the `invariant` and the various functions where `len` appears in the *output* , the spec should be an `&&` instead of an `||`. With that perhaps the tests will also pass!

waddlaw:
I see. Updated commit!

include/Data/Vector.spec
waddlaw: Thanks for advice!  :)

waddlaw:
These test with failed.

## failed positive test case

- [vector2.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/pos/vector2.hs)
- [kmpVec.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/pos/kmpVec.hs)
- [ClojurVector.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/pos/ClojurVector.hs)
- [BinarySearchOverflow.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/pos/BinarySearchOverflow.hs)
- [BinarySearch.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/pos/BinarySearch.hs)

nikivazou:
Should we close this?

PR #1233, opened 2018-02-09
Backport GHC differential D4388 to Desugar.

shlevy:
This allows liquidhaskell to build with the new CostCentre interface.

Note that the referenced change is not going to be backported to an official GHC 8.2 release and may not make it into an official 8.4 release, but it is in the pipeline to be included in GHC 8.2 and 8.4 provided with nixpkgs, as it fixes an otherwise unworkaroundable blocker in doing distributed builds with nix and GHC and profiling libs. I doubt this PR will be desired as-is, it seems like there are three different paths here:

1. Do nothing for now, and once liquidhaskell is bumped to GHC 8.6 this will have to get fixed one way or another
2. Bump liquidhaskell to track GHC 8.6 now
3. Add a cabal flag or similar to switch between the current and new code paths depending on whether D4388 is applied to ghc or not.

Please let me know which path you'd like to take.

### DIFF
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -62,6 +62,10 @@ Flag include
   Description: use in-tree include directory
   Default:     False
 
+Flag deterministic-profiling
+  Description: Support building against GHC with https://phabricator.haskell.org/D4388 backported
+  Default:     False
+
 Executable liquid
   default-language: Haskell98
   Build-Depends: base >=4.8.1.0 && <5
@@ -281,6 +285,8 @@ Library
      hs-source-dirs: devel
    if flag(devel)
      ghc-options: -Werror
+   if flag(deterministic-profiling)
+     cpp-options: -DDETERMINISTIC_PROFILING
    Default-Extensions: PatternGuards
 
 test-suite test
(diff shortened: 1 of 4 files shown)


ranjitjhala:
Thanks @shlevy !!! Is there a link to the CostCentre interface that you can point us to?

In the meantime, I think option 3 would be best, if possible?

Thanks again for this!

shlevy:
@ranjitjhala The patch as a whole is here https://phabricator.haskell.org/D4388 , the changes to CostCentre are here https://phabricator.haskell.org/differential/changeset/?ref=188068&whitespace=ignore-most . Basically just replaces the `Unique` in each cost centre with a numeric index that increases per centre emitted in a given file, which in turn ensures that the symbol names emitted are deterministic.

I'll get started on the CPP, thanks!

shlevy:
@ranjitjhala Updated with a cabal flag.

ranjitjhala:
Thanks @shlevy !!!

I think this is ready to merge. *When* we move to GHC-8.4 we may ping you with some Qs, is that ok?

Thanks!

shlevy:
Absolutely!

ranjitjhala:
Thanks very much for this @shlevy!!

PR #1240, opened 2018-02-18
Parsing support for singleton and empty lists in the refinements

nikivazou:
See https://github.com/ucsd-progsys/liquidhaskell/blob/d1b94e27043d8cc049fdfb2958a741465e75f252/tests/pos/SingletonLists.hs

matches https://github.com/ucsd-progsys/liquid-fixpoint/pull/355


ranjitjhala:
Thanks @nikivazou !!! I think Edsko asked for this feature at some point? Maybe we can close that issue too?

Issue #1198, opened 2017-12-15
Measure Invariants


ranjitjhala:
@nikivazou why does the below program fail 

```haskell
module Term where

{-@ data Tree [sz] @-}
data Tree = Leaf | Node Tree Tree

{-@ measure sz @-}
{-@ sz :: Tree -> Nat @-}
sz :: Tree -> Int
sz Leaf = 0
sz (Node t1 t2) = 1 + sz t1 + sz t2
```

**FAIL** with the odd (termination) error: 

```
/Users/rjhala/research/sf/liquid-sf/term.hs:10:23-27: Error: Liquid Type Mismatch

 10 | sz (Node t1 t2) = 1 + sz t1 + sz t2
                            ^^^^^

   Inferred type
     VV : {v : Tree | (Term.sz v < Term.sz ?a => Term.sz v >= 0)
                      && v == t1}

   not a subtype of Required type
     VV : {VV : Tree | Term.sz VV < Term.sz ?a

                       && Term.sz VV >= 0}

   In Context
     t1 : {t1 : Tree | Term.sz t1 < Term.sz ?a => Term.sz t1 >= 0}

     ?a : {?a : Tree | Term.sz ?a < Term.sz ?a => Term.sz ?a >= 0}
```

Where is the antecedent `Term.sz t1 < Term.sz ?a => ...` that guards `Term.sz t1 >= 0` coming from? 

Why is it necessary? 

Note that the code passes if you add:

```haskell
{-@ invariant {v:Tree | 0 <= sz v } @-}
```

but this is unnecessary, as it should be generated from the measure? What am I missing here?

nikivazou:
I run into this again.... 

The guard is necessary, I believe for soundness: it says that the result refinement (that `0 <= sz t`) is valid only on `t` that is provably smaller than the argument of the function. 

Should I remove it? 
I believe it is required.....

In other words... I am not sure how to prove termination of the "termination metric" function.
And I suspect we need some notion of syntactic termination checking for those...

ranjitjhala:
The `<` is not useful in the above because the ordering 
is undefined; i.e. the refinement logic can do nothing with 
it other than conservatively assume that it does not hold.

Instead we *definitely* need to implement the standard 
`inductive-datatype` definition, doing a syntactic check 
e.g. where the following _is inductive_,

```haskell
data List a = Emp | Cons a (List a)
```

but where the following is _not_:

```haskell
data List a = Cons a (List a)
```

[...]

PR #1244, opened 2018-02-21
add structural induction checks for termination metrics

nikivazou:
For #1198


ranjitjhala:
Wow, awesome!!!

nikivazou:
I have errors from programs similar to 

```
{-@ data LL [llen] @-} 

{-@ invariant {v:LL a | (llen v) >= 0} @-}

{-@ measure llen :: (LL a) -> Int 
    llen(N)      = 0
    llen(C x xs) = 1 + (llen xs)
  @-}
```

Since `llen` is a logic function but not a Haskell one. 
I will edit the programs similar to the above that generate an error: we are going towards to all logic functions should be defined in  Haskell, correct?

ranjitjhala:
Yes, if possible!

(Is the error here that the default size is NOT defined by a core expr and
hence outside your new checker?)

nikivazou:
The error is 
```
     Not in scope: variable or data constructor 'llen'
```

It is generated in Bare when I want to turn the symbol `llen` into a ghc `Var`...

ranjitjhala:
Ah ok got it.

ranjitjhala:
Only 32 failures v good! Thought there would be hundreds!

ranjitjhala:
In short: all default sizes have to now be GHC vars, right?

nikivazou:
I am fixing the tests. 

@ranjitjhala: why do I need to export all the Haskell functions that I declare as measures?

ranjitjhala:
@nikivazou -- the measures (and reflects, inlines etc.) have to be 
exported because effectively `measure` is a **global name**. If you have:

```haskell
module A (func) where 

{-@ measure foo @-}
foo = ...

{-@ func :: ... uses `foo` in its type @-}
func = ...
```

and then you **import** `A` elsewhere, then we need to **resolve** 
the  name `foo` that appears in the signature for `A.foo` and for that, 
GHC requires that `foo` be exported from `A`.

nikivazou:
This is because importing module `A` imports all measures of `A`. We should change that (if possible). I might import `A` but not all its measures...

BTW, now all my failing tests have to do with name resolution...........

ranjitjhala:
1. Btw, maybe you can just issue a 

`WARNING: Measure 'bslen' not a GHC Variable, omitting structural well-foundedness check`

or similar for "legacy" measures that are not GHC variables?

2. I found out recently that you can rerun a subset of tests matching a pattern with:

```
stack test liquidhaskell --fast --test-arguments="-p Unit"
```

if you change, e.g. `-p Unit` to`-p ByteString` then it will just run those tests...

ranjitjhala:
Yes, ideally we'd do that, but its very non-trivial, because unlike with
GHC,
where importing `foo` just imports `foo`, for us, we have to end up
importing
everything that appears (transitively) in the signature for foo...

nikivazou:
The problem is that `lookupGhcVar` throws the ghc error, and I do not know of any way to catch it and emit the warming you are suggesting... 

Ok, will make yet another flag to deactivate the checks for BS & Text. 
Yet, now my errors in text* are 

```
 Data/Text/Internal.hs:69:13: Error: GHC Error

 69 | {-@ measure tlen @-}
                  ^

     Not in scope: variable or data constructor 'tlen'
```

because Internal exports the measure `tlen`, but my current module is not importing it... 
This is annoying

*Cannot run BS locally even with your suggestion... I am missing something...

ranjitjhala:
ok, the flag is a reasonable workaround for now...I can then take a look.

ranjitjhala:
Yay! Thanks @nikivazou!!

nikivazou:
@ranjitjhala again text and bytestring reveal a name resolution mess, so if you want to give it a try, up to you :)

ranjitjhala:
Sounds good : they work with your flag rigght?

nikivazou:
They all work: merging!

PR #1249, opened 2018-02-23
Add blog on case expand

nikivazou:


ranjitjhala:
I really don't like `CoreToLogic.lg` btw, that needs to eventually go...

docs/blog/2018-02-23-measures-and-case-splitting.lhs
```diff
@@ -0,0 +1,148 @@
...
+Measures
+---------
+
+Let's define a simple data type with three alternatives 
+
+\begin{code}
+data ABC = A | B | C 
+\end{code}
+
+and a measure that turns `ABD` into an integer
```
ranjitjhala: `ABD` ---> `ABC` ?

nikivazou:
Oups, sorry, this edit was in my local version, but I need it anyways so let it too go on develop!

I do not like it either, but I have another issue on that.

PR #1259, opened 2018-03-21
Make all equational combinators return their right-hand side.

rosekunkel:
From a proof standpoint, these definitions are equivalent. However, defining the equality combinators to return their right hand side has two advantages:

1. It makes defining optimized functions using the equality combinators more natural. This way, we can write
```haskell
foo = correctButInefficientDefinition
```
and then use the operators to refine it:
```haskell
foo = correctButInefficientDefinition
      === simplifiedDefinition
      === optimizedDefinition
```
and end up with `foo` having the optimized definition.

2. It brings the definitions in line with the inequality combinators, which all return their right hand sides.


ranjitjhala:
Great idea, this has bitten me before too! Thanks @wkunkel !!!

Issue #1267, opened 2018-03-28
Imprecisions with no-case-expand 


nikivazou:
The following is UNSAFE. It should get safe even with the `no-case-expand` flag.

```
{-@ LIQUID "--no-case-expand" @-}

module NoCaseExpand where

data ABC = A | B | C 

foo :: Int -> ABC -> ()
foo 0 A  =  ()

foo x A | x /= 0 = ()

foo _ A = error " " 

foo _ t = ()
```

ranjitjhala:
Interesting. Wonder what the ghc-core is for this program...

nikivazou:
Exactly what I was looking into....

The core is this 

```
fooCore x y = 
  let fail _ = case y of 
                 A -> case x /=  0 of 
                       False -> error " " 
                       True  -> () 
                 _ -> () 
  in case x of 
       0 -> case y of 
               A -> () 
               _ -> fail ()
       _ -> fail () 
````

Seems that to avoid the error, we need to inline `fail`...

ranjitjhala:
!!! how does it come up with that bizarre code.

Regardless, @nikivazou I am not wholly persuaded
by your conclusion, because if you type that exact
code in, LH is happy to prove SAFE, even with `"--no-case-expand"`

http://goto.ucsd.edu:8090/index.html#?demo=permalink%2F1522267761_23890.hs

?

nikivazou:
Apart that `ABC` should have 3 data constructors! Not two as in your example above. 
When the default case expands to exactly one data constructor, LH always expands it.

ranjitjhala:
Good point!

PR #1269, opened 2018-03-29
fix for #1267

nikivazou:
Strengthen the DEFAULT case with data con alternative checker info, if available.


ranjitjhala:
Good idea! this basically says “default is NOT equal to one of the matched
patterns”right?

nikivazou:
Yes. 

Still imprecise though... 

For instance, if I define a measure `isB`, LH does not know that `isB v <=> is$ModuleName.B`. 
So, it cannot automatically prove `isB x` after the cases where x is matched with B.
But I do not know how to make is more precise...

ranjitjhala:
You mean there are extra facts that get put in the env when we do “full”
case expand, right?

nikivazou:
Yes.

Issue #1095, opened 2017-09-20
LH runs forever (>5min) on complicated patterns


nikivazou:
The following program takes too long to verify (>6 min)

```
{-@ LIQUID "--exactdc"     @-}
{-@ LIQUID "--higherorder" @-}

module Foo where

data Foo 
  = A Foo Foo Foo 
  | B Foo Foo
  | C Foo
  | D Foo
  | E Foo Int 
  | F Int
  | G Int
  | H 
  | I
  | J  
  | K
  | L
  | M
  | N 
  | X

{-@ 
data Foo [size]
  = A {a1 :: Foo, a2 :: Foo, a3 ::  Foo} 
  | B {b1 :: Foo, b2 :: Foo}
  | C {c  :: Foo}
  | D {d  :: Foo}
  | E {e1 :: Foo, e2 :: Int} 
  | F {f  :: Int}
  | G {g  :: Int}
  | H 
  | I
  | J  
  | K
  | L
  | M
  | N 
  | X
  @-}

size :: Foo -> Integer 
{-@ measure size @-}
{-@ invariant {t:Foo | 0 <= size t} @-}
{-@ size :: Foo -> {v:Integer |  0 <= v }  @-}
size (A x y z) = 1 + size x + size y + size z 
size (B x y)   = 1 + size x + size y 
size (C x)     = 1 + size x 
size (D x)     = 1 + size x 
size (E x _)   = 1 + size x 
size (F _)     = 1 
size (G _)     = 1 
size _         = 0 


foo  :: Foo -> Foo
foo (A X _ _) = X 
foo (A _ X _) = X 
foo (A _ _ X) = X 
foo (B X _)   = X 
foo (B _ X)   = X 
foo (C X)     = X 
foo (D X)     = X 
foo (E X _)   = X 
foo X         = X 
foo (A x y z) = A (foo x) (foo y) (foo z)
foo (B x y)   = B (foo x) (foo y)
foo (C x)     = C (foo x)
foo (D x)     = D (foo x)
foo (E x i)   = E (foo x) i 
foo x         = x 
```

The problem is that `foo` has very complicated case analysis. 
In such cases, my hack was to use ` --no-case-expand` to simplify the ghc internal cases analysis, but this would break reflection of `foo`. 

Note that writing `foo` as follows runs in <2 min

```
{-@ reflect hasX @-}
hasX :: Foo -> Bool 
hasX (A X _ _) = True 
hasX (A _ X _) = True 
[...]

PR #1279, opened 2018-04-02
Case Expand Upto Depth

ranjitjhala:
Rather than make `--case-expand` a Yes/No proposition, allow a "depth" upto which to expand.

```
--max-case-expand=N
```

expands `DEFAULT` at pattern matches whose nesting (i.e. the number of `case-of` that the `DEFAULT` occurs under is `<= N` (the top level case-of being at depth=`1`.)

So, `--no-case-expand` is just `--max-case-expand=0`.

Currently am experimenting with the default depth being `2`, while also emitting a WARNING if there are a ton of cases at the given depth.


ranjitjhala:
Looks like with a default of 2, even the example from #1095 takes about 8s (with `-max-case-expand=0` it takes ~1s). So I think this is reasonable to merge in.

@nikivazou and @jprider63 any further thoughts?

(I guess we can always revert if we don't like this...)

nikivazou:
Do tests break with the default 2? 

Did you check if the current expansion could be optimized? That is still do the expansion but reduce the code explosion. 

I am a bit skeptical, because we are basically adding another (fuel-like) heuristic with this optimization. I suggest we leave as default the expand all option we had before. 

PS: I though you did not like warnings :)

ranjitjhala:
* I'm not sure what such an optimization would look like; its an interesting problem but I don't see any particularly good ideas for how to solve it. The expanding really is a simple heuristic for getting a LOT of precision in many cases. e.g. look at `tests/pos/gadtEval.hs` -- it fails with `--no-case expand` and that's unfortunate as it is very idiomatic code, that _should_ be handled by default. 

* Yet, the default of _expand-everywhere_ is terrible as shown by #1095.

* So while the numeric thing _is_ fuel-like, IMO we had the same fuel before, only you could just choose 0 or infinity; now you 

(a) have finer grained control, and 
(b) there is a default "middle ground" that seems reasonable for most cases, and 
(c) while I don't like warnings, these warnings tell you how to make them to go away :-)

That said, there is something odd that GHC is doing, consider the following function:

```haskell
merge :: (Ord a) => [a] -> [a] -> [a]
merge [] ys = ys
merge xs [] = xs
merge (x:xs) (y:ys)
  | x > y     = y : merge (x:xs)   ys
  | otherwise = x : merge    xs (y:ys)
```

GHC gives us  core that looks like the below: the odd thing is the "nested" case-split on `ds_d1VH` 
which is somehow matched against DEFAULT **two** times. Seems pretty strange! (And is the kind of thing that would cause case-expand to blow up...)


```haskell
 Main.merge
   = \ (@ a)
       ($dOrd_a1UP :: GHC.Classes.Ord a)
       (ds_d1VH :: [a])
       (ys :: [a]) ->
       join {
         fail_d1VS :: GHC.Prim.Void# -> [a]
         [LclId[JoinId(1)]]
         fail_d1VS _ [Occ=Dead, OS=OneShot]
           = Control.Exception.Base.patError
               @ 'GHC.Types.LiftedRep
               @ [a]
               "/Users/rjhala/research/stack/liquidhaskell/tests/todo/Merge.hs:(4,1)-(8,38)|function merge"# } in
       case ds_d1VH of {
         __DEFAULT ->
           case ys of {
             __DEFAULT ->
               case ds_d1VH of {
                 __DEFAULT -> jump fail_d1VS GHC.Prim.void#;
                 : x xs ->
                   case ys of {
                     __DEFAULT -> jump fail_d1VS GHC.Prim.void#;
                     : y ys ->
                       case scc<CAF>
                            scc<CAF> break<2>(x,y) (scc<CAF> GHC.Classes.> @ a $dOrd_a1UP) x y
                       of {
                         GHC.Types.False ->
                           break<8>(ys,xs,x,y)
                           GHC.Types.:
                             @ a
                             x
                             (scc<CAF>
                              scc<CAF>
                              scc<CAF>
                              break<7>(ys,xs,y)
                              (scc<CAF> (scc<CAF> Main.merge @ a $dOrd_a1UP) xs)
                                (break<6>(ys,y) GHC.Types.: @ a y ys));
                         GHC.Types.True ->
                           break<5>(ys,xs,x,y)
                           GHC.Types.:
                             @ a
                             y
                             (scc<CAF>
                              scc<CAF>
                              scc<CAF>
                              break<4>(ys,xs,x)
                              (scc<CAF>
                               (scc<CAF> Main.merge @ a $dOrd_a1UP)
                                 (break<3>(xs,x) GHC.Types.: @ a x xs))
                                ys)
                       }
                   }
               };
             [] -> break<1>(ds_d1VH) ds_d1VH
           };
         [] -> break<0>(ys) ys
       };,
```

ranjitjhala:
Though yes, I do agree the warnings are annoying. One could switch them off, but maybe its better to experiment with this for a while and then if things are too annoying to _then_ go back to the boolean expand.

nikivazou:
Do you know what are the semantics of `join`? Maybe `join` is at fault for this duplication.

ranjitjhala:
Yes, the GhcListSort.hs tests break,
because they use odd things like

```
case compare x y of
  | GT -> e1
  | _  -> e2
```


which even Hlint says to rewrite as


```
| x < y     = e1
| otherwise = e2
```

(which then works fine ...)

Yes there's probably something going on with
the `join` will have to re-read the PLDI paper
on that more closely...

nikivazou:
After discussion with James,  if we are to turn this into a research problem ("How should we expand so that we have enough precision, but no exponential blow ups") it would be cool if we are to address this example: https://github.com/plum-umd/lmonad-meta/blob/master/src/Simulations/EraseEvalErase.hs#L313

Where expansion is required to reflect the definitions of the reflected functions at each case.

PR #1281, opened 2018-04-03
Termination Checker: Allow sub-sub-terms and lexicographic ordering

nomeata:
i.e. accept

	{-@ measure sz @-}
	sz :: List a -> Int
	sz Nil = 0
	sz (Cons _ Nil) = 1
	sz (Cons _ (Cons _ l)) = 2 + sz l

as terminating.

This commit also adds a failing test case of nested recursion, which
_should_ be accepted by my checker, but fails for other reasons (not
sure if nested recursion is accepted by LH at all.)


nikivazou:
Thanks! 
Well, LH does not care if you recursion is nested as long as you have defined the proper sizing measure. That is, if you assume that your nested `sz` sizing function is terminating, then LH can use it to prove termination on functions that recurse on your Trees.

PR #1282, opened 2018-04-03
add structural termination flag

nikivazou:
Using `--structural` flag, the structural termination checker is used to check all functions. 

Initial purpose of the flag: let @nomeata develop the structural termination checker. e.g., the added [example](https://github.com/ucsd-progsys/liquidhaskell/commit/560e5b78820ca5e43b1febeb5c5c5b1a12b6a5b3#diff-c79be91fc15879ad54fbfa50c647251f) now fails.

### DIFF
--- a/src/Language/Haskell/Liquid/Termination/Structural.hs
+++ b/src/Language/Haskell/Liquid/Termination/Structural.hs
@@ -22,6 +22,8 @@ import qualified Data.HashSet                           as S
 
 
 terminationCheck :: GhcInfo -> Output Doc 
+terminationCheck info | structuralTerm (getConfig info)
+                      = mconcat $ map (resultToDoc . checkBind (cbs info)) (allRecBinds $ cbs info)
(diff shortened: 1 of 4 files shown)


ranjitjhala:
@nikivazou minor nit: can you make the flag `--structural-termination` ? Just `--structural` is a bit unclear IMO...

nomeata:
(Although if I understand Niki correctly, then this is just something for me to play around with, and will be removed afterwards, when the termination checker is tried on all functions, and the ones that are found to be terminating are treated like `lazy`. So maybe no need to bikeshed about a shed that will be dismanteled soon)

ranjitjhala:
Oh, right! good point!

nikivazou:
--structural-termination is the name of the [flag](https://github.com/ucsd-progsys/liquidhaskell/pull/1282/files#diff-074daff8be693a7b782a2e1cca19583cR155), but any prefix works!

PR #1294, opened 2018-04-15
squash cabal warning

atondwal:


ranjitjhala:
Those lines are in the file for a reason (but I don't know what that reason is!)

I tried in the past to use `hpack` instead which seems much nicer (avoids duplication) but somehow it was too complicated to port our .cabal file...

atondwal:
yeah, I thought I'd found the configuration to make the missing-modules problems go away, but looks like that only worked on my machine because of cached object files laying around

PR #1307, opened 2018-04-26
remove unresonable check

nikivazou:


ranjitjhala:
@nikivazou please don't remove this check; as you can see from the failed test, it is actually quite a reasonable one!

ranjitjhala:
maybe the real q is: why is this check failing on @jbrown215 's code?

nikivazou:
Yes! This is why I removed it. 
The check is needed but it is too restrictive, e.g., it rejects @jbrown215 's well sorted example. 

I do not think I can properly relax the check any time soon :(

ranjitjhala:
Ok fine, let’s comment out that failing test, and
we can come back to this when I torch and
rewrite the absref code ... 😀

PR #1310, opened 2018-04-28
Implicit Arguments

atondwal:
Matching lf PR: https://github.com/ucsd-progsys/liquid-fixpoint/pull/363

Note that both these PRs include fixes to `circle..yml`: stackage changed the location of their stack binary.


ranjitjhala:
Cool! Can you wait a bit, would like to review this before merging! Thanks!

atondwal:
[Cleaned up some of the commits a little so that they're not so hard to review]

PR #1327, opened 2018-06-05
Equational Reasoning operations without the type class nonsense 

nikivazou:
The trick: give `?` higher priority so that that the "explanation" is just embedded in the term that is compared. 

This way, verification is faster (since I removed the type class nonsense for optional arguments) and all the intermediate steps are checked!


ranjitjhala:
Thanks @nikivazou ! Can you add an example somewhere, e.g. the equivalent 
of [this file](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/include/Language/Haskell/Liquid/Example.hs) but using the new combinators?

nikivazou:
I will, but nothing is changing just use `==.` instead of `===`.

nikivazou:
BTW, @ranjitjhala can you update the liquid server with this (aka the merged) develop?

PR #1349, opened 2018-07-25
Circle2

atondwal:


atondwal:
@ranjitjhala here's the circle2 config .... I can't seem to be able to get --liquid-opts to work, so maybe we could default to cores=1 in test or something, or should we not bother?

ranjitjhala:
Awesome! I suggest you make whatever edits needed to get it to work for now
(ie disable liquid-opts if needed?) we can revisit later if it’s an issue?

atondwal:
grr, Client521.hs is nondeterministically failing... build #1349 fails, but #2321 succeeds, both on the this commit!

PR #1386, opened 2018-12-12
Incremental PLE

ranjitjhala:
This PR changes how LH checks "proofs" (to make verification much faster, specifically, to basically allow PLE to be on "by default" -- though that is not done, yet.)

1. No polymorphic instantiation inside proofs, i.e. we don't generate fresh "liquid" KVars for polymorphic functions - *except* when using abstract refinements (see the new signature for `ProofCombinators.?`) 

2. PLE itself builds a TRIE and then incrementally traverses it to collect equalities at each node, thereby "sharing" the work across constraints.

3. The ONLY place that (step 1) caused a serious regression is the diff for the case `eval_safe g s (EVar x)` for the commit 578a2ee

 
Matching PR for https://github.com/ucsd-progsys/liquid-fixpoint/pull/381


ranjitjhala:
PLE-OPT

```
rjhala@borscht ~/r/s/liquidhaskell (ple-opt)> time stack exec -- liquid tests/ple/pos/STLC2.hs --full
       15.66 real        14.71 user         1.06 sys
rjhala@borscht ~/r/s/c/src (master)> time stack exec -- liquid Ch03_ex_11.hs --full
        5.88 real         5.19 user         0.65 sys
```

OLD-DEVELOP

```
rjhala@borscht ~/r/s/liquidhaskell (develop)> time stack exec -- liquid tests/ple/pos/STLC2.hs --full
       22.43 real        24.25 user         1.55 sys
rjhala@borscht ~/r/s/c/src (master)> time stack exec -- liquid Ch03_ex_11.hs   --full
       12.04 real        12.16 user         1.18 sys
```

PR #1392, opened 2019-01-09
Upgrade to GHC 8.4.4

mhwombat:
These changes will allow liquidhaskell to build with GHC 8.4.4. I have tested building it with both Stack and Nix/NixOS. You may not be ready for that yet, so feel free to disregard this pull request.

One change I made that may not be what you want: `stack.yaml` now expects to find liquid-fixpoint in `../liquid-fixpoint` instead of `./liquid-fixpoint`. I made that change in my fork because I found nested repositories a bit confusing to work with. It's my understanding that nested repos aren't recommended. However, if you'd like me to undo that, let me know.

There are a lot of changes to default.nix and shell.nix. That's because I'm using the versions that are automatically generated by cabal2nix. I think this will be simpler for others who might work on the package.

For future reference, here's how I generated those files.

```
cabal2nix . > default.nix
cabal2nix . --shell > shell.nix
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,40 @@
+{ mkDerivation, aeson, array, base, bifunctors, binary, bytestring
+, Cabal, cereal, cmdargs, containers, data-default, deepseq, Diff
+, directory, exceptions, filepath, fingertree, ghc, ghc-boot
+, ghc-paths, ghc-prim, gitrev, hashable, hpc, hscolour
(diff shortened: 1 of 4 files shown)


ranjitjhala:
Thanks @mhwombat much appreciated! Let me see what is failing...

nikivazou:
I assume we don;t need that anymore :(

Issue #1396, opened 2019-01-19
Totality checking of lazy functions depends on argument order


rosekunkel:
Given the following definitions:
```haskell
data Peano = Z | S Peano

{-@ lazy ack @-}
ack :: Peano -> Peano -> Peano
ack Z n = S n
ack (S m) Z = ack m (S Z)
ack (S m) (S n) = ack m (ack (S m) n)

{-@ lazy ackFlipped @-}
ackFlipped :: Peano -> Peano -> Peano
ackFlipped n Z = S n
ackFlipped Z (S m) = ackFlipped (S Z) m
ackFlipped (S n) (S m) = ackFlipped (ackFlipped n (S m)) m
```
Liquid Haskell fails to prove `ackFlipped` is total, but it has no trouble with `ack`:
```text
Error: Liquid Type Mismatch
  
 11 | ackFlipped n Z = S n
      ^^^^^^^^^^
  
   Inferred type
     VV : {v : GHC.Prim.Addr# | v == "/home/rose/research/liquidhaskell/tests/terminate/pos/Ackermann.hs:(11,1)-(13,58)|function ackFlipped"}
  
   not a subtype of Required type
     VV : {VV : GHC.Prim.Addr# | 5 < 4}
```
If you give `ackFlipped` an explicit termination metric, it succeeds:
```haskell
data Peano = Z | S Peano

{-@ measure peanoToNat @-}
{-@ peanoToNat :: Peano -> Nat @-}
peanoToNat :: Peano -> Int
peanoToNat Z = 0
peanoToNat (S n) = 1 + peanoToNat n

{-@ lazy ack @-}
ack :: Peano -> Peano -> Peano
ack Z n = S n
ack (S m) Z = ack m (S Z)
ack (S m) (S n) = ack m (ack (S m) n)

{-@ ackFlipped :: n:_ -> m:_ -> _ / [peanoToNat m, peanoToNat n] @-}
ackFlipped :: Peano -> Peano -> Peano
ackFlipped n Z = S n
ackFlipped Z (S m) = ackFlipped (S Z) m
ackFlipped (S n) (S m) = ackFlipped (ackFlipped n (S m)) m
```
This is also a problem for the new structural termination checker: even if `ackFlipped` passes it, it fails the totality check.

ranjitjhala:
Interesting; I wonder what the core for ackFlipped looks like. In
particular why does GHC not think it total?

rosekunkel:
[...]

PR #1397, opened 2019-01-25
Make structural termination checker argument-order-agnostic.

### LINKED ISSUE #1381: Extend Structural Termination
The new structural checker by @nomeata is awesome. Would be great to extend it to handle things like:

```haskell
data Pea = Z | S Pea

foo :: Int -> Pea -> Int 
foo n Z     = n
foo n (S p) = foo (n + 1) p  
```

rosekunkel:
Fixes https://github.com/ucsd-progsys/liquidhaskell/issues/1381.

This will likely cause some regression in error messages, at least until https://github.com/ucsd-progsys/liquidhaskell/issues/1396 is fixed. Also, there are still some structurally-recursive programs that this won't be able to check, I'll make a new issue for that limitation (but it has a very simple workaround).

### DIFF
--- a/src/Language/Haskell/Liquid/Termination/Structural.hs
+++ b/src/Language/Haskell/Liquid/Termination/Structural.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE DeriveGeneric       #-}
 
 module Language.Haskell.Liquid.Termination.Structural (terminationVars) where
 
@@ -14,15 +15,18 @@ import VarSet
 
 import           Text.PrettyPrint.HughesPJ hiding ((<>)) 
 import qualified Data.HashSet  as S
+import Data.Hashable (Hashable)
+import GHC.Generics (Generic)
+import Data.Maybe (mapMaybe)
 
 terminationVars :: GhcInfo -> [Var]
-terminationVars = resultVars . terminationCheck 
+terminationVars = resultVars . terminationCheck
 
-resultVars :: Result -> [Var]
-resultVars OK         = [] 
+resultVars :: Result a -> [Var]
+resultVars (OK _)       = []
 resultVars (Error es) = teVar <$> es 
 
-terminationCheck :: GhcInfo -> Result 
+terminationCheck :: GhcInfo -> Result ()
 terminationCheck info 
   | isStruct  = mconcat $ map (checkBind cbs) (allRecBinds cbs)
   | otherwise = mconcat $ map (checkBind cbs) (S.toList $ gsStTerm $ gsTerm $ giSpec info)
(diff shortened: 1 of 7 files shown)


ranjitjhala:
Excellent let’s see what CI says

ranjitjhala:
Yay all tests passed!

ranjitjhala:
@rosekunkel this is awesome! Does it make sense to add some neg tests too? I guess it’s ok that the old neg tests still fail?

rosekunkel:
More tests are always good, but this change should just be accepting more correct programs, not rejecting more wrong programs. So I'm not sure what new neg tests I would add.

ranjitjhala:
Thanks @rosekunkel !!!

src/Language/Haskell/Liquid/Termination/Structural.hs
```diff
@@ -154,14 +160,59 @@ check env args = \case
...
+
+instance Hashable Call
+
+-- This is where we  check a function call. We go through  the list of arguments
+-- and find the  indices of those which are decreasing.  Note that this approach
+-- is only guaranteed to  work when the arguments to the  function are named, so
+-- e.g.
+-- foo (x:xs) (y:ys) = foo xs (y:ys)
+-- won't necessarily work, but
+-- foo (x:xs) yys@(y:ys) = foo xs yys
```
nomeata, line 177: I don’t understand this part. Core has no patterns in function definitions!

src/Language/Haskell/Liquid/Termination/Structural.hs
```diff
@@ -154,14 +160,59 @@ check env args = \case
...
+        if S.null ca then
+          raiseErr calls
+        else
+          structDecreasing env $ (S.fromList . mapMaybe (removeDecreasing ca) . S.toList) calls
+  where
+    raiseErr calls =
+      retErr env $ text "Non-structural recursion in calls" <+>
+      (text $ showPpr (S.map callSource calls))
+    nonDecCalls = S.filter (S.null . decreasingArgs) calls
+    commonArgs calls = foldl1 S.intersection (S.map candidateArgs calls)
```
nomeata, line 212: Code smell, sometimes you have `calls` as a parameter here, sometimes you just use the `calls` from `structDecreasing`. That’s confusing.

nomeata (commented):
I read through the algorithm, and it seems to make sense to me.

ranjitjhala:
My bad, @nomeata  -- I merged too excitedly :-)

@rosekunkel can you take a look at Joachim's remarks? [Maybe add comments to clarify if ok?]

src/Language/Haskell/Liquid/Termination/Structural.hs
ranjitjhala: I believe this has to do with #1398

rosekunkel: Yes, this is referring to the problem in #1398. It's not unique to patterns in function definitions, the same should happen if you used an explicit `case` statement.

src/Language/Haskell/Liquid/Termination/Structural.hs
rosekunkel: This is a good point, I'll change it.

src/Language/Haskell/Liquid/Termination/Structural.hs
nomeata: Oh, ok. I wasn’t reading the example to the end, the problem is not the pattern, but the reconstruction.

PR #1399, opened 2019-01-25
Make helpers for structDecreasing consistent.

rosekunkel:
Addresses @nomeata's concerns about the code style in my previous PR: https://github.com/ucsd-progsys/liquidhaskell/pull/1397#discussion_r251080249


nomeata:
Heh, you actually resolved it the other way than I would have done (no need to abstract over `calls` when you always pass `calls` anyways). But this is getting philosophical…

rosekunkel:
Well, I do call `raiseError` on both `calls` and `nonDec calls`. Granted, I don't think these error messages actually go anywhere...

nomeata:
Good point.

nomeata (approved):

ranjitjhala:
So this looks good right, shall I merge?

Issue #1393, opened 2019-01-11
well formedness confusion on imported modules


nikivazou:
Consider the following three modules: 

```haskell
module C where 

{-@ reflect cProp @-}

cProp :: Eq a => a -> a -> Bool 
cProp x y = x == y 
```

```haskell
module B where 

import C

data B a = B {b1 :: a, b2 :: a }
{-@ data B a = B {b1 :: a, b2 :: {v:a | cProp b1 v } } @-}
```

```haskell
module A where 

import B 

a :: Eq a => B a -> Bool 
a (B x y) = x == y 
```

`liquid` is SAFE on modules `B` and `C`, but running liquid on module `A` returns 

```
 B.hs:6:10-11: Error: Illegal type specification for `B.b2`

 6 | {-@ data B a = B {b1 :: a, b2 :: {v:a | cProp b1 v } } @-}
```

This is very confusing, since module `B` is safe (but `cProp` is out of scope in module `A`).
I do not think checking module `A` should check the specifications of `B` on `A`'s module environment.

ranjitjhala:
This, unfortunately is a  hard problem, that I’ve seen before. The problem is this: how do you hide parts of a refinement of a constructor for example because those parts are no longer in scope? Right now the simplest workaround is just to transitively import all the modules. This is a bit silly I agree but honestly I don’t know what a good solution is.

nikivazou:
There are other potential "silly" solutions, like 
- prune from A all predicates not in scope or 
- make B reexport all the refinements that appear on the data constructors. 

I too do not know which one is better. But if we keep it as is, we should change the error message, when checking the imported modules.

ranjitjhala:
[For the record] I like @nikivazou 's suggestion that the saved `.bspec` files should be "closed", i.e. should pull in transitive dependencies. I do worry whether the GHC names are in scope though, but that's another issue...

facundominguez:
Closing this since I can't reproduce it in the latest LH.

PR #1400, opened 2019-01-28
Explicit logic imports

nikivazou:
First step to address https://github.com/ucsd-progsys/liquidhaskell/issues/1393

### DIFF
--- a/src/Language/Haskell/Liquid/GHC/Interface.hs
+++ b/src/Language/Haskell/Liquid/GHC/Interface.hs
@@ -313,7 +313,6 @@ processModules cfg logicMap tgtFiles depGraph homeModules = do
   where                                             
     go = processModule cfg logicMap (S.fromList tgtFiles) depGraph
 
-
 processModule :: Config -> LogicMap -> S.HashSet FilePath -> DepGraph -> SpecEnv -> ModSummary
               -> Ghc (SpecEnv, Maybe GhcInfo)
 processModule cfg logicMap tgtFiles depGraph specEnv modSummary = do
@@ -748,7 +747,7 @@ listLMap  = toLogicMap [ (dummyLoc nilName , []     , hNil)
 --------------------------------------------------------------------------------
 -- | Pretty Printing -----------------------------------------------------------
 --------------------------------------------------------------------------------
-{- 
+
 instance PPrint GhcSpec where
   pprintTidy k spec = vcat
     [ "******* Target Variables ********************"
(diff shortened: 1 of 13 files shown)


nikivazou:
@ranjitjhala seems that the measures defined using the old commented syntax do not get qualified names, is there a reason for that?

ranjitjhala:
The only things that do get qualified names are those that have matching
ghc elements (datacons, Tycon, var etc)

Old style measures have no such corresponding object hence...

of course this is not a great reason — if you find some way to do it go for
it. I didn’t spend time on it because we don’t write such measures any more
..

nikivazou:
Makes sense! 
But then, when we check for duplicate definitions, e.g. https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/errors/ShadowMeasure.hs we need to remove the module names?

nikivazou:
The tests: https://github.com/ucsd-progsys/liquidhaskell/pull/1400/commits/7ec8a4c76627c23d89dacc922d0345215e936c70 get accepted so let's merge (if Circle says ok). 

@ranjitjhala question: if there is a bspec, then why do we also parse the initial file? See: https://github.com/ucsd-progsys/liquidhaskell/blob/develop/src/Language/Haskell/Liquid/GHC/Interface.hs#L329-L331

src/Language/Haskell/Liquid/GHC/Interface.hs
```diff
@@ -339,7 +339,7 @@ updLiftedSpec :: Ms.BareSpec -> Maybe Ms.BareSpec -> Ms.BareSpec
 updLiftedSpec s1 Nothing   = s1 
 updLiftedSpec s1 (Just s2) = (clear s1) `mappend` s2 
   where 
-    clear s                = s { sigs = [], aliases = [], ealiases = [], qualifiers = [] }
+    clear s                = s { sigs = [], aliases = [], ealiases = [], qualifiers = [], dataDecls = [] }
```
ranjitjhala, line 342: Why is it ok to wipe the `dataDecls`? Is it because the data-constructor signatures are still kept so we don't really need the datatype?

nikivazou: No, it is because the dataDecls of the parsed file are not qualified. 
In my example, e.g., they use cProp instead of C.cProp. 
So I have to remove the parsed ones and replace them with the qualified ones....

ranjitjhala:
👍

ranjitjhala:
cool, thanks! I'm *hoping* this also magically fixes another issue that I've found recently re: imports :)

PR #1419, opened 2019-03-10
Class laws definition and checking

nikivazou:
See for example

1. https://github.com/ucsd-progsys/liquidhaskell/blob/b0216fad35e700b0fc23331a08a4dff54b32e77e/tests/class-laws/SemiGroup.hs

2. https://github.com/ucsd-progsys/liquidhaskell/blob/b0216fad35e700b0fc23331a08a4dff54b32e77e/tests/class-laws/Monoid.hs


ranjitjhala:
Are the instances being checked? Would be good to have some neg tests where the tests fail.

Also: I guess it’s now possible to create an instance without the laws? (That May be a feature rather than a limitation...)

nikivazou:
The instances are only being checked only when you write `{-@ instance laws .... @-}`, e..g, 
https://github.com/ucsd-progsys/liquidhaskell/blob/b0216fad35e700b0fc23331a08a4dff54b32e77e/tests/class-laws/Monoid.hs#L37-L43

ranjitjhala:
Yes I meant:

1. I assume if you have a bogus proof inside the instance laws it should be rejected right? Can you add a test for that?

2. Can you clarify the exact syntax of the “definitions” in the instance laws block? Specifically, are the rhs of the equations always just variables (identifiers) as the tests suggest?

nikivazou:
1. Proofs are checked using the standard LH. So, if there is a bug in the proof but the specification is indeed the instance of the class law, then the {-@ instance law @-} should be fine and the proof rejected. Will add a test now. 

2. Yes. The rhs is (currently) a variable. Implementation wise the instance laws defined a substitution that I am using to get from the class laws definitions the instances, so it can quite easily be extended so that the rhs is logic expressions, but I cannot think a use case of that. Any suggestions on what you would like to write?

PR #1434, opened 2019-03-27
add a simple spec for fromGregorian / toGregorian

peti:
I needed those refined types in my own code and thought
they might be useful in general.


nikivazou:
Thanks!

Issue #1435, opened 2019-03-27
Error: Illegal type specification for `GHC.List.splitAt`


peti:
The current `develop` version of LH won't cope with `Control.Lens`. I'll paste the (longish) error message below. Note that LH succeeds if I replace `Control.Lens` by `Control.Lens.TH`, so it's apparently not the derived lens instances that cause trouble; it's something else. Anyway, here is the error case I ran into:

~~~ Haskell
{-# LANGUAGE TemplateHaskell #-}

module Address where

import Control.Lens
import Data.Text

data Address = Address { _street :: Text
                       , _postcode :: Text
                       , _city :: Text
                       }
  deriving (Show, Eq)

makeLenses ''Address
~~~

~~~ bash
$ liquid Address.hs
LiquidHaskell Version 0.8.6.0
Copyright 2013-19 Regents of the University of California. All Rights Reserved.

[1;94m
**** DONE:  A-Normalization ****************************************************
[0m[1;90m 
[0m[1;94m
**** DONE:  annotate ***********************************************************
[0m[1;90m 
[0m[1;95m
**** RESULT: ERROR *************************************************************
[0m

 /nix/store/qsz7kiig0w8w0gq38d544jh6f6zrg2rd-liquidhaskell-0.8.6.0-pre-release-data/share/ghc-8.6.4/x86_64-linux-ghc-8.6.4/liquidhaskell-0.8.6.0/include/GHC/List.spec:(38,12)-(39,0): Error: Illegal type specification for `GHC.List.splitAt`
  
 38 | splitAt :: n:_ -> x:[a] -> ({v:[a] | (if (n >= 0) then (Min (len v) (len x) n) else ((len v) = 0))},[a])<{\x1 x2 -> (len x2) = (len x) - (len x1)}>
 39 | span    :: (a -> GHC.Types.Bool) 
  
     GHC.List.splitAt :: forall a##xo .
                         n:{VV : GHC.Types.Int | $HOLE[n:=VV]}
                         -> x:[a##xo]
                            -> ({v : [a##xo] | (n >= 0 => Min (len v) (len x) n)
                                               && (not (n >= 0) => len v == 0)}, [a##xo])<\x1 VV -> {VV : [a##xo] | len VV == len x - len x1}>
     Sort Error in Refinement: {v : [a##xo] | ((n >= 0 => Data.Semigroup.Min (len v) (len x) n)
                                               && (not (n >= 0) => len v == 0))}
[...]

PR #1438, opened 2019-03-27
include/GHC/List.spec: inline 'Min' predicate to avoid name clash with Data.Semigroup.Min

peti:
Fixes https://github.com/ucsd-progsys/liquidhaskell/issues/1435.


nikivazou:
SORRY @peti my fault. I broke the develop while you were editing... But it is fixed now. Can you please rebase the PRs? (I don't think I have permission to do that myself)

peti:
Done.

Issue #1423, opened 2019-03-15
Abstract refinements + reflection + GADTs causes fixpoint crash


rosekunkel:
The following file causes a crash:
```haskell
{-# LANGUAGE GADTs #-}
{-@ LIQUID "--reflection" @-}
{-@ LIQUID "--exact-data-con" @-}
module Gadt where

{-@
data Foo a <p :: Int -> Bool> where
   A :: Foo<{\_ -> True}> ()
@-}
data Foo a where
  A :: Foo ()

{-@ reflect getFoo @-}
getFoo :: Foo a -> a
getFoo A = ()
```
The output is
```text
LiquidHaskell Version 0.8.4.0, Git revision e557f2f4f992afcaf25dd70d1e73865334668138
Copyright 2013-18 Regents of the University of California. All Rights Reserved.


**** DONE:  A-Normalization ****************************************************
 

**** DONE:  Extracted Core using GHC *******************************************
 

**** DONE:  Transformed Core ***************************************************
 
liquid: Constraint.instantiatePvs
CallStack (from HasCallStack):
  error, called at src/Language/Fixpoint/Misc.hs:151:14 in liquid-fixpoint-0.8.0.1-5lBHQ0phVgG9Ya0ErK2yUa:Language.Fixpoint.Misc
```

ranjitjhala:
Oh no. Regrettably this has touched the dark corners of the abstract
refinement implementation that only @nikivazou (if anyone) understands ...

nikivazou:
Well, I have no idea how are GADTs represented in core and how they are different than ADTs. 
Will take a look next week.

ranjitjhala:
Yes it’s quite likely that the GADTs you some different kind of
representation and that the code for abstract refinements need to to be
updated to account for that...

nikivazou:
Fixed, but check my comment in #1439

PR #1439, opened 2019-03-27
fix for #1423

nikivazou:
This is a patch for #1423 but not sure it will take us far in GADTs. 
The issue is that user-defined data refinements refine the wrapper data constructors. 
In case analysis the worker data constructor is used. 

In most cases these two are the same and everything is smooth (in that the Abstract Refinements and in general the type specs are the same for both). 
With GADTs these two might not be the same. Up to now the worker had no abstract refinements and thus the crash. This fix ports the abstract refinements of the wrapper to the worker and for this example it is fine since the fields on the data constructor are not actually using the abstract refinement. But in the general case, where the abstract refinement is actually used, this could get unsound. 
A sound alternative to this fix would be to disallow data refinements in all data types whose wrapper and worker are not the same.

### DIFF
--- a/src/Language/Haskell/Liquid/Types/PredType.hs
+++ b/src/Language/Haskell/Liquid/Types/PredType.hs
@@ -129,6 +129,7 @@ meetWorkWrapRep c workR wrapR
   = workR { ty_binds = xs ++ (ty_binds wrapR)
           , ty_args  = ts ++ zipWith strengthenRType ts' (ty_args wrapR)
           , ty_res   = strengthenRType (ty_res workR)    (ty_res  wrapR)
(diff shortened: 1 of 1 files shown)


ranjitjhala:
@nikivazou can you explain more why the worker/wrapper thing can 
introduce unsoundness? They are just two different GHC binders used 
to represent the `DataCon`?

nikivazou:
The refinement type of the wrapper is syntactically defined by the user refinement-data definition. 
Since it is unclear to me how the worker type is derived (when it is different from the wrapper) I am not sure if the refinements of the worker are sound (i.e., if the user specified refinements are correctly propagated via the worker)

PR #1440, opened 2019-03-27
Holes

nikivazou:
Allow haskell expression holes in the verified programs, see https://github.com/ucsd-progsys/liquidhaskell/blob/4265255175092e3b389bac9f6001c614052b9d2d/tests/pos/T1440.hs


ranjitjhala:
What does this do? Fix a bug or add a feature? If the latter, what’s a test
case that exercises it?

nikivazou:
Feature, added test case in the description. 

The main edit is calling ghc with the `-fdefer-typed-holes` flag. 
With this, Haskell programs with holes get compiled from ghc (before that compilation would error and liquid terminate). 
When run with `-v` you can see the liquid types of the holes.

nikivazou:
@ranjitjhala happy now? Can I merge? 

If the source file has no holes, then no changes should be made so there should be no slowdown. 
(Saying should, since it appears that the core might be changed when ghc runs with the `-fdefer-typed-holes` flag).

ranjitjhala:
So is the idea you can now use holes instead of undefined or similar hacks?

ranjitjhala:
Actually not yet happy :)

If the idea is to output the types of holes then

1 don’t put it under -v as that is really for debugging LH there is a lot
of junk; instead just print out the types of holes by default
2 Add a test to check the right type or something like it is being printed
out; eg see the Error-Tests in test.hs

nikivazou:
Yes, the idea is you can use holes (prefixed with `_`)
This has already been supported by ghc but compilation ends with an error unless you use the `-fdefer-typed-holes` flag, which basically translates holes internally into an error. 

I will add proper printing instead of the current `-v`, but it is not stable to add tests. 
The main problem is that currently, there is no src location for the holes (which would be needed to report the type, I assume).

nikivazou:
Now the holes and their type are printed as errors (which is quite cool because you can see the info on VS code :))

Exactly like ghc, internally the holes as warnings, but (turns out) LH prints warnings as errors (I think we had in the past a discussion on that.)

PR #1444, opened 2019-04-02
Optimize Windows builds

varosi:
Cache removed for now and use liquid target to build liquid-fixpoint.
This resulted in <50min build times. This should fix the problem with AppVeyor timeout on 1 hour.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,11 +22,6 @@ install:
 - z3 --version
 
 build_script:
-# Test build liquid-fixpoint
-- cd liquid-fixpoint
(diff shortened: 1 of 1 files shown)


varosi:
42min build time - https://ci.appveyor.com/project/ranjitjhala/liquidhaskell/builds/23552432

varosi:
32bit builds are much faster, but there are no 32 bit GHC 8.6.4 for Windows. They said that there will be for GHC 8.6.5.

varosi:
Thanks!

PR #1486, opened 2019-05-09
Fix typo: extentionality -> extensionality

googleson78:


ranjitjhala:
Thanks @googleson78 ! (Overriding CI failure because those are unrelated...)

PR #1492, opened 2019-05-15
link with waitSMT

nikivazou:
Link to https://github.com/ucsd-progsys/liquid-fixpoint/pull/399


nikivazou:
I compared running times and there is a small slowdown (26 sec in all tests, see here: https://docs.google.com/spreadsheets/d/1JHUCLXceh__YDomI5ga48cuqNWiLx0WIF_2etfE_zjY/edit?usp=sharing)

But, when I run the tests locally (with laster text version `text 1.2.3.1` and `Z3 version 4.8.3`) the following tests 
```
- ple-neg/T1192.hs
Macro
- unit-pos/zipper.hs
- unit-pos/selfList.hs
- unit-pos/stacks0.hs

Benchmarks
- icfp-neg/DataBase.hs
- icfp-pos/DataBase.hs
```

loop forever with this branch and throw an `hGetLine: end of file` exception with develop. 
My understanding is that the test suite ignores the `hGetLine: end of file` exception, but I am confused as that circle does not loop forever in this branch. 

@ranjitjhala can you run `liquid tests/ple/neg/T1192.hs` (on develop) to check if you also get the `end of file` exception?

zgrannan:
@nikivazou I just tried the test on the `develop` branch and wasn't able to reproduce the `end of file` exception. This was also using `text 1.2.3.1` and `Z3 version 4.8.3`.

As part of trying to reproduce it, I tried introducing some latency by creating a wrapper to slow down z3
```bash
#!/usr/bin/env bash
sleep 1
$REAL_Z3_EXECUTABLE $@ | while read line; do
  sleep 1
  echo $line                              
done
```
The slow Z3 wasn't able to reproduce the issue either.

Maybe `Z3` is crashing for some reason running those tests? That would seem to explain the behavior on the `develop` and `waitSMT` branches, because if the stream terminated it would constantly be in the `EOF` state.

nikivazou:
Thanks @zgrannan. 
Just to be sure: you run these tests using the `liquid` command line, (e.g., `liquid tests/neg/T1192.hs`? 
Not simply running `stack test`, because in this case this error is suppressed. 

If that is the case, you should be correct that something is wrong with my z3 and close this issue.

zgrannan:
Ahh so actually I only used `Z3` version `4.8.3` on the first test (`T1192`). I re-ran it on all the tests and the tests for 
```
- icfp/neg/DataBase.hs
- icfp/pos/DataBase.hs
```
both failed with

```
liquid: fd:13: hGetLine: end of file
```

__However, the tests pass for me using `Z3` version `4.8.4`__. It looks like `Z3` version `4.8.3` is segfaulting. You can try using the command `z3 -smt2 -in < pos_database.txt` to replicate it. It fails using 4.8.3 but succeeds with 4.8.4

[pos_database.txt](https://github.com/ucsd-progsys/liquidhaskell/files/3274774/pos_database.txt)

nikivazou:
Oh, thanks! So now we can totally blame z3 and close it. 
I still don't fully understand why the LH code is fine, since we use `hGetLine` to read the SMT response and `hGetLine` does not block. But I am happy to close it.

PR #1497, opened 2019-05-20
Add a spec file for Data.Text

googleson78:
This is done in similar fashion to the `Data.ByteString` one.

Some functions that are missing in `Data.Text` are removed, and some are named differently/have different arguments. I tried to adjust this, hopefully not missing anything.


ranjitjhala:
Super, thanks!!!

PR #1504, opened 2019-06-07
Kill extensionality

nikivazou:


ranjitjhala:
@nikivazou was this not needed? or did you tweak the benchmarks? or what has happened to allow the removal of this feature (not that I am complaining...)

nikivazou:
I did tweak the POPL benchmarks, for the branch to pass. 
But this feature is clearly special cased for these benchmarks and does not work in general (thus removing). James was asking for some complicated functions in fixpoint and motivated this removal.

PR #1514, opened 2019-06-13
Fix Show for RTAlias

zgrannan:
Ran into a small issue with the definition of `show` while doing some debugging.

It looks like the format string has one more argument than necessary, resulting in an error
```
printf: argument list ended prematurely
```

The error appears to be introduced in: https://github.com/ucsd-progsys/liquidhaskell/commit/99754dbcd48c58f6ed7326d878bb8b4f97cd197b#diff-5fc4ae8750cd8b70894d6a005369102f


ranjitjhala:
Nice, thank you so much!!!

Issue #1498, opened 2019-05-22
Refinements on class instances are ignored (+ crash???)


googleson78:
The following file:
```haskell
class A a where
  f :: a -> Int

instance A Int where
  {-@ f :: _ -> {x : Int | 0 < x} @-}
  f n = n

{-@ x :: {x : Int | 0 < x} @-}
x :: Int
x = f ((-1) :: Int)
```
is declared as `SAFE` by `liquid`.

The following is also declared as `SAFE`:
```haskell
class A a where
  f :: a -> Int

instance A Int where
{-@
instance A Int where
  f :: _ -> {x : Int | 0 < x}
@-}
  f n = (-1)
```

More interestingly the following generates a crash:
```haskell
class A a where
  f :: a -> Int

instance A Int where
{-@
instance A Int where
  f :: _ -> {x : Int | 0 < x}
@-}
  f n = (-1)

{-@ g :: Int -> {n : Int | 0 < n} @-}
g :: Int -> Int
g x = f x
```
Crash log:
```
 /tmp/liquid/class/Classes.hs:13:9: Error: Uh oh.

 13 | g x = f x
              ^

     addC: malformed constraint:
{lq_tmp$x##610 : GHC.Types.Int | lq_tmp$x##610 == x##aVJ}
 <:
{VV : _ | $HOLE[lq_tmp$db##0:=VV]}
```

Am I misunderstanding [this section](https://github.com/ucsd-progsys/liquidhaskell#modules-with-code-type-classes)?

ranjitjhala:
Wow this is surprising to me!

I *think* what is happening here is that the signature for “f” is

1. Getting attached to the class method and
2. NOT getting checked at the instance.

However even if 1 is happening (counterintuitive, but possible because of
how name resolution happens) I don’t understand why 2 should happen?

ranjitjhala:
Ok, here's what's happening, which is consistent with all the above.

1. When you don't write "instance" inside the `{-@ ... @-}` the spec is associated with the class,
2. When you write "instance" it is associated with the instance.

Either way, the text in [that section](https://github.com/ucsd-progsys/liquidhaskell#modules-with-code-type-classes) should be clarified to make explicit that we *never check instance signatures*. 
They are *always trusted*, and we just *use* the signatures at client "call" sites. 
I think this is a missing feature (and can lead to this kind of unsoundness). 

I _think_ we never got around to implementing this because its a bit hard 
to tell where they are created?

[...]

PR #1515, opened 2019-06-13
Fix error reporting for #T1498

nikivazou:
Re #1498.

- 0: 
```
class A a where
  f :: a -> Int

instance A Int where
  {-@ f :: _ -> {x : Int | 0 < x} @-}
  f n = n

{-@ x :: {x : Int | 0 < x} @-}
x :: Int
x = f ((-1) :: Int)
```
Reports 
```
 Invalid type specification for Main.f
 Use class or instance refinements instead.
```

- 1:
```
class A a where
  f :: a -> Int

instance A Int where
{-@
instance A Int where
  f :: _ -> {x : Int | 0 < x}
@-}
  f n = (-1)
```
Reports 
```
Error: Bad Data Specification

 7 |   f :: _ -> {x : Int | 0 < x}
            ^
 Cannot resolve type hole `_`. Use explicit type instead. for `f`
```

- 2: Similar to 1


ranjitjhala:
Niki, can you add a link in the read me to some examples that use class and
instance refinements so that we can point Users those to those to see what
the syntax is? Thanks!

nikivazou:
This is not good? https://github.com/ucsd-progsys/liquidhaskell#modules-with-code-type-classes

ranjitjhala:
If that’s current then perfect! (I somehow thought the syntax had changed
but maybe that was for the laws?)

nikivazou:
Yes, I did not change anything. 
Just adding some proper error messages when LH used to crash before.

PR #1516, opened 2019-06-13
Change function application syntax

zgrannan:
Liquid Haskell currently parses `f (x, y)` as `(f x) y`. This will no longer work if https://github.com/ucsd-progsys/liquid-fixpoint/pull/405 is merged. This PR updates some tests to use the normal Haskell function application syntax.


ranjitjhala:
Oops, sorry I missed this, thanks!!!

PR #1517, opened 2019-06-14
Remove colouring of output, because it makes problems in latest Stack 2.1

varosi:


nikivazou:
Thanks @varosi-chaosgroup, the appveyor is still failing :(

varosi-chaosgroup:
>       Hole00.hs:                      FAIL (0.24s)
>         tests\\test.hs:442:
>         Wrong exit code
>         expected: ExitSuccess
>          but got: ExitFailure 1
>       HiddenMethod00.hs:              OK (1.82s)
>       Class00.hs:                     OK (1.81s)

This test fail which is not connected with colouring output.

ranjitjhala:
Yes it works fine on the Linux and Mac which is puzzling!

Can you tell us how we can get the logs? Ie can we say into the machine
where the tests are run etc?

The other option would be to disable this test on windows?

varosi:
Seems that it's working now, so I'm closing this pull request.

PR #1544, opened 2019-09-16
add refinement for function valued selector

yiyunliu:
With reflection enabled, we might want to refine a function valued field and invoke the selector to propagate the refinements. The selector is currently not properly generated and the proof would go through only if we extract the field using pattern matching:
```haskell
{-@ data ThingMM = ThingMM {f :: Int -> Int, fprop :: v:Int -> {f v >= 0}} @-}

data ThingMM =
  ThingMM
    { f :: Int -> Int
    , fprop :: Int -> ()
    }

{-@ safe' :: t:ThingMM -> v:Int -> {f t v >= 0} @-}
safe' :: ThingMM -> Int -> ()
safe' (ThingMM _ fprop) i = const () $ fprop i

-- should be safe
{-@ unsafe :: t:ThingMM -> v:Int -> {f t v >= 0} @-}
unsafe :: ThingMM -> Int -> ()
unsafe t i = const () $ fprop t i
```

### DIFF
--- a/src/Language/Haskell/Liquid/Bare/DataType.hs
+++ b/src/Language/Haskell/Liquid/Bare/DataType.hs
@@ -662,7 +662,7 @@ makeRecordSelectorSigs env name = checkRecordSelectorSigs . concatMap makeOne
   where
   makeOne (Loc l l' dcp)
     | null fls                    --    no field labels
-    || any (isFunTy . snd) args   -- OR function-valued fields
+    || any (isFunTy . snd) args && not (higherOrderFlag env)   -- OR function-valued fields
     || dcpIsGadt dcp              -- OR GADT style datcon
     = []
     | otherwise 
--- a/tests/pos/T1544.hs
+++ b/tests/pos/T1544.hs
@@ -1,0 +1,17 @@
+{-@ LIQUID "--reflection"    @-}
+module T1544 where
+
+{-@ data ThingMM = ThingMM {f :: Int -> Int, fprop :: v:Int -> {f v >= 0}} @-}
+data ThingMM =
+  ThingMM
+    { f :: Int -> Int
+    , fprop :: Int -> ()
+    }
+
+{-@ fieldLemmaPat :: t:ThingMM -> v:Int -> {f t v >= 0} @-}
+fieldLemmaPat :: ThingMM -> Int -> ()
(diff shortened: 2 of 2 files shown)


nikivazou:
@yiyunliu can you also add this example to a new test `tests/pos/T1544.hs` to make sure we do not break it in the future? 

Thanks!!!

yiyunliu:
Updated

nikivazou:
Great thanks!

@yiyunliu ci/circleci  complains, but I think it is a false alarm. 
`tests/neg/T1013A.hs` seems to be safe now, but it seems that with your edit we can now soundly verify it. Do you agree? If so, can you please move it to the `tests/pos` so that circleci accepts your change?

yiyunliu:
The last time I checked it seems that z3 was complaining about duplicate
definition. (Exit code changed from expected -1 to -2). I'll look into that
soon

yiyunliu:
The SMT error was caused by checking out the wrong version of liquid-fixpoint. After updating my submodule, it starts to complain about sort error instead, which I believe happens before the subtype checking (that's where the original error message was emitted).

nikivazou:
I get a liquid type error, no SMT errors. 
We can chat tomorrow about this?

yiyunliu:
@nikivazou Could you rerun the tests? I merged the latest commit you pushed to develop and it seems to work locally (failing with ExitCode 1, like it used to).
I don't think the selector refinement would make T1013A.hs safe, since the file uses pattern matching already and doesn't involve any selectors.

nikivazou:
Correct! I just saw the error you are getting. The generated type of the `Bug.fmapCompose` is not correct. For some reason it gives to the `x` argument the type `x: a a` instead of `x: m a`. 
I can dig into it, but not before Friday. If you want to give it a try before that, you are more than welcome :)

yiyunliu:
The merged version passes the tests. The issue somehow gets fixed by the RankNType pull request..

jprider63:
The issue seems to be with RankNTypes and not refinements for function fields. Should we merge this and open a separate issue for RankNTypes?

nikivazou:
What a nice coincidence :) Kind of makes sense, because the failing test has `{-# LANGUAGE RankNTypes #-}`. 

Merging! Thanks!

PR #1548, opened 2019-09-25
Parse function composition

nikivazou:
Function Composition is properly parsed in refinements. e.g., you can write https://github.com/ucsd-progsys/liquidhaskell/blob/768e5ef799b0dee4ce2d3ea02ebe155d7079073f/tests/pos/T1548.hs#L10

There are two things to be addressed in different PR: 
1) In `(g . snd) (p)` the argument (here `p`) needs to be in parenthesis. This is a general limitation of our parser: when the function is not a variable, the argument needs to be parenthesized. 

2) More importantly, the definition of Haskell's function composition is hardcoded. See comment below

### DIFF
--- a/include/Data/Tuple.spec
+++ b/include/Data/Tuple.spec
@@ -1,4 +1,4 @@
 module spec Data.Tuple where
 
-fst :: x:(a,b) -> {v:a | v = (fst x)}
-snd :: x:(a,b) -> {v:b | v = (snd x)}
+fst :: {f:(x:(a,b) -> {v:a | v = (fst x)}) | f == fst }
+snd :: {f:(x:(a,b) -> {v:b | v = (snd x)}) | f == snd }
--- a/include/GHC/Base.spec
+++ b/include/GHC/Base.spec
@@ -18,6 +18,11 @@ len (y:ys) = 1 + len ys
 // null []     = true 
 // null (y:ys) = false
 
+
+// WiredIn in Language.Haskell.Liquid.Types.Names
+measure liquidCompose :: (b -> c) -> (a -> b) -> a -> c
+(.) :: {o3:(f:(b -> c) -> {o2: (g:(a -> b) -> {o1: (x:a -> {o:c | o == liquidCompose f g x && o == f (g x)}) | o1 == liquidCompose f g}) | o2 == liquidCompose f}) | o3 = liquidCompose }
+
 measure fst :: (a, b) -> a
 fst (a, b) = a
 
--- a/include/GHC/Base.spec
+++ b/include/GHC/Base.spec
@@ -18,11 +18,6 @@ len (y:ys) = 1 + len ys
 // null []     = true 
 // null (y:ys) = false
 
-
-// WiredIn in Language.Haskell.Liquid.Types.Names
-measure liquidCompose :: (b -> c) -> (a -> b) -> a -> c
-(.) :: {o3:(f:(b -> c) -> {o2: (g:(a -> b) -> {o1: (x:a -> {o:c | o == liquidCompose f g x && o == f (g x)}) | o1 == liquidCompose f g}) | o2 == liquidCompose f}) | o3 = liquidCompose }
-
 measure fst :: (a, b) -> a
 fst (a, b) = a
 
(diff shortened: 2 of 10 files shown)


nikivazou:
The definition of function composition is now hardcoded. I tried @conal 's suggestion to get this definition by the inline information (`maybeUnfoldingTemplate . realIdUnfolding`) of `GHC.Base..` but failed.  Below is the code where of the inline and debugging of the unfolding information of the functions, which is currently empty for `.`. 

https://github.com/ucsd-progsys/liquidhaskell/blob/parse-function-composition/src/Language/Haskell/Liquid/Bare/Axiom.hs#L215-L238

conal:
Thanks very much for this effort. Hopefully we'll figure out how to make the simpler implementation (`maybeUnfoldingTemplate . realIdUnfolding`) work and then be able use it for other inlinings.

What unfolding debugging info gets generated for `(.)`?

conal:
> https://github.com/ucsd-progsys/liquidhaskell/blob/768e5ef799b0dee4ce2d3ea02ebe155d7079073f/tests/pos/T1548.hs#L10

> In `(g . snd) (p)` the argument (here `p`) needs to be in parenthesis. This is a general limitation of our parser: when the function is not a variable, the argument needs to be parenthesized.

I think I'd almost always prefer function-level equations, which luckily avoid this parsing limitation:

``` haskell
{-@ h :: g:(b -> d) -> { snd . second g == g . snd } @-}
```

Another question worries me, however. I don't yet know how to persuade LiquidHaskell to apply such a property to an *application* in which the function compositions are implicit, e.g., to justify `snd (second g (a,b)) == snd (a,g b)`.

nikivazou:
> What unfolding debugging info gets generated for `(.)`?

I get `realIdUnfolding = No unfolding`. 
FYI, an inlined function is defined within the checked module, the id has the proper `realIdUnfolding`.

nikivazou:
> I think I'd almost always prefer function-level equations, which luckily avoid this parsing limitation:
> 
> ```haskell
> {-@ h :: g:(b -> d) -> { snd . second g == g . snd } @-}
> ```

To get that you can apply the `extensionality` axiom, e.g., 
https://github.com/ucsd-progsys/liquidhaskell/blob/6d226b645607aee4b945b74c736f1d4203571ca8/tests/pos/T1548.hs#L17-L18
> 
> Another question worries me, however. I don't yet know how to persuade LiquidHaskell to apply such a property to an _application_ in which the function compositions are implicit, e.g., to justify `snd (second g (a,b)) == snd (a,g b)`.

You can now prove the above directly with equational reasoning: https://github.com/ucsd-progsys/liquidhaskell/blob/6d226b645607aee4b945b74c736f1d4203571ca8/tests/pos/T1548.hs#L10-L14

This proof is ok after the recent edit, merged in develop, that gives singleton type to `snd`. 
Is this equational proof what you want, or you want to derive it via `h`?

conal:
> To get that you can apply the `extensionality` axiom ...

Great. That's just what I hoped. I have this same definition commented out in our `Theorems.hs`, but it didn't work.
I'll update my liquidhaskell repo and try again.

<blockquote>
You can now prove the above directly with equational reasoning:

https://github.com/ucsd-progsys/liquidhaskell/blob/6d226b645607aee4b945b74c736f1d4203571ca8/tests/pos/T1548.hs#L10-L14

This proof is ok after the recent edit, merged in develop, that gives singleton type to `snd`.
Is this equational proof what you want, or you want to derive it via `h`?
</blockquote>

Oops! My mistake. I think I really wanted `snd (second g (a,b)) == g (snd (a,b))`, or even just `snd (second g p) == g (snd p)`, for which I think we'll need `h`.

nikivazou:
> 
> Oops! My mistake. I think I really wanted `snd (second g (a,b)) == g (snd (a,b))`, or even just `snd (second g p) == g (snd p)`, for which I think we'll need `h`.

Then you can do 
```
{-@ example :: g:(b -> d) -> p:(a,b) -> { snd (second g p) == g (snd p) } @-}
example g p 
  =   snd (second g p) 
  ==. (snd . second g) p 
      ? hFun g
  ==. (g . snd) p 
  ==. g (snd p)
  *** QED 
```

conal:
> Then you can do ...

Thanks. Makes sense, though more explicit than I was hoping for. I guess there's a tension here, in that properties are often most elegantly expressed at the level of functions but applied at the level of *applications* of those functions. In an informal calculation, I think I'd usually say 
``` haskell
   snd (second g p) 
== {- hFun -}
   g (snd p)
```

I hadn't realized how much nontrivial reasoning this sort of "proof" elides. I'll get more experience with LH, and then maybe a more elegant but sufficiently explicit style will emerge.

nikivazou:
Looking forward for the new style :) 

But indeed LH currently needs more information than your informal proof. 
`hFun` has a hidden forall, LH definitely needs a way to instantiate this forall with `p`.

PR #1550, opened 2019-09-27
reflect non saturated functions

nikivazou:


ranjitjhala:
Nice! Can you update the test with some simple proofs that use the
(reflected) functions too?

nikivazou:
> Nice! Can you update the test with some simple proofs that use the (reflected) functions too?

Added!

PR #1553, opened 2019-09-30
do not instantiate type variables when reflection flag in on

nikivazou:


ranjitjhala:
Strange, I thought I’d implemented this awhile ago and had to add the
abstract refinement to various specs to recover instantiation — what did I
miss?

nikivazou:
You only apply that when PLE is on, not on reflection.

ranjitjhala:
Ah, looks like I ONLY disabled it when PLE (ie not reflection) was going on?

nikivazou:
@ranjitjhala I asked your review for that because it breaks many of your STLC proofs. 
You can deactivate this edits `{-@ LIQUID "--unsafe-polymorphism" @-}` (which is used for the tests that broke). 
Since there are more edits coming, I prefer to merge soon.

ranjitjhala:
Hmm unfortunately can’t look at them ATM: can you just add that flag so the
tests don’t break?

ranjitjhala:
@nikivazou please do not merge this — clearly disabling polymorphism is breaking way too many tests: I would like us to have a better solution for this because it’s just not clear that 

(a) it is unsound
(b) that this is a good fix

the better default is -natded or something, which explicitly disables instantiation (and hence is more targeted for now) rather than breaking half our tests.

ranjitjhala:
Niki I posted a comment — this break seems far too invasive and breaks way
too many things for somewhat unclear benefit; let’s instead explicitly
disable instantiation (and keep the defaults as is for now) until we have a
better understanding.

Another idea is to use a specially decorated type variable eg a# or
whatever to indicate no instantiation and use only those in your natded
proofs, extensionality axioms and so on.

But either way, In short, please don’t merge this.

ranjitjhala:
I am extra bothered by the addition of the “assume” on various quite simple tests.

ranjitjhala:
The more I think of it, this seems the right approach:

> Another idea is to use a specially decorated type variable eg a# or whatever to indicate no instantiation and use only those in your natded proofs, extensionality axioms and so on. 

All the weirdness was arising due to polymorphic instantiation in “assumed” functions. At least that is the case for the extensionality and natded example you showed.

So: just disable polymorphic instantiation for “assumed” signatures? This is not a hack: the trouble in BOTH those cases was that because of the assume, there were no internal constraints on the Tyvar coming from the BODY of the assume, which led to a FALSE. 

In the regular (non assume) case we don’t have this problem because suitable constraints show up on the type variable.

So:

Option 1: disable instantiation for assumed signatures (only) not all signatures

Option 2: use a special type variable to denote cases where instantiation should be disabled.

I think 1 is the cleanest option.

If you think otherwise, I’d like to see an example like your funext or natded ones, that *do not* use an assume and still cause false to get proved.

ranjitjhala:
> the trouble in BOTH those cases was that because of the assume, there were no internal constraints on the Tyvar coming from the BODY of the assume

In fact thinking some more, I realize that whenever a type variable appears POSITIVELY in a function signature, the constraints on the tyvar come from the BODY of the function. So, if the signature is ASSUMED then no constraints and you can instantiation with false.

So the exact solution is to *only disable instantiation for tyvars that appear positively in assumed signatures.*

nikivazou:
ok! I am not merging this (this is why I explicitly asked your review :)) 

> So the exact solution is to only disable instantiation for tyvars that appear positively in assumed signatures.

By positively, you mean type variables that only appear in negative positions on the arguments? If so, I am quite persuaded that this is what we want.

PR #1559, opened 2019-10-07
lighter rejection of type variable instantiation

nikivazou:
Implementation of Ranjit's suggestion in #1553. Let's see what breaks now.

### DIFF
--- a/benchmarks/popl18/ple/pos/BasicLambdas.hs
+++ b/benchmarks/popl18/ple/pos/BasicLambdas.hs
@@ -43,14 +43,6 @@ id :: a -> a
 id x = x
 
 
-{-@ fmap_id' 
-  :: x:(r -> a)
-  -> {v:Proof | (\r:r -> id (x r)) ==  (\r:r -> (x r) ) } @-}
-fmap_id' :: (r -> a) ->  Proof
-fmap_id' x
-   =   fun_eq (\rrr1 -> x rrr1) (\rrr2 -> id (x rrr2)) (\r -> x  r === id (x r) *** QED)
-
-
 
 {-@ assume fun_eq :: f:(a -> b) -> g:(a -> b) 
    -> (x:a -> {f x == g x}) -> {f == g} 
--- a/src/Language/Haskell/Liquid/Types/Types.hs
+++ b/src/Language/Haskell/Liquid/Types/Types.hs
@@ -74,6 +74,7 @@ module Language.Haskell.Liquid.Types.Types (
   , RTVar (..), RTVInfo (..)
   , makeRTVar, mapTyVarValue
   , dropTyVarInfo, rTVarToBind
+  , setRtvPol
 
   -- * Predicate Variables
   , PVar (PV, pname, parg, ptype, pargs), isPropPV, pvType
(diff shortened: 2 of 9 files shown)


nikivazou:
@ranjitjhala circleci and I are quite confident on this. Let me know if you want to wait for your review again. 

I am not sure if this edit only applies on assumed functions (as is now) or should also apply for the imported, but we can discuss this on a next call.

src/Language/Haskell/Liquid/Types/Types.hs
```diff
@@ -795,15 +796,19 @@ mapTyVarValue :: (tv1 -> tv2) -> RTVar tv1 s -> RTVar tv2 s
...
-dropTyVarInfo v = v{ty_var_info = RTVNoInfo}
+dropTyVarInfo v = v{ty_var_info = RTVNoInfo True }
 
 data RTVInfo s
-  = RTVNoInfo
+  = RTVNoInfo { rtv_is_pol :: Bool }
   | RTVInfo { rtv_name   :: Symbol
             , rtv_kind   :: s
             , rtv_is_val :: Bool
+            , rtv_is_pol :: Bool
```
ranjitjhala: Can you add some documentation here to explain what this field means?eg an example of when the field is true/false?

ranjitjhala (changes requested):
This looks great (assuming tests pass etc.) looks like very few changes? (Ie additions of assume?)

But can you add the comment about the field before merging? Thanks!

ranjitjhala:
Hi Niki, this looks good right it doesn’t break all the tests and so on? I
left one comment that is could you please add some description of what
happens at that field? Just for documentation

src/Language/Haskell/Liquid/Types/Types.hs
nikivazou: added documentation!

ranjitjhala (approved):
Great, thanks!

benchmarks/popl18/ple/pos/BasicLambdas.hs
```diff
@@ -43,14 +43,6 @@ id :: a -> a
 id x = x
 
 
-{-@ fmap_id' 
-  :: x:(r -> a)
-  -> {v:Proof | (\r:r -> id (x r)) ==  (\r:r -> (x r) ) } @-}
-fmap_id' :: (r -> a) ->  Proof
-fmap_id' x
-   =   fun_eq (\rrr1 -> x rrr1) (\rrr2 -> id (x rrr2)) (\r -> x  r === id (x r) *** QED)
```
ranjitjhala, line 51: Do you know why this test fails? (No need to hold up the merge, just curious?)

ranjitjhala:
@nikivazou -- looks good to me, thanks! this looks like it worked out quite well? (A little surprisingly well actually, but good you added the neg tests...)

benchmarks/popl18/ple/pos/BasicLambdas.hs
nikivazou: I have removed support for alpha equivalence, there is not reason why it passed. 
Before it was ansoundly passing because extensionality was not checking its proof argument at all.

nikivazou:
> @nikivazou -- looks good to me, thanks! this looks like it worked out quite well? (A little surprisingly well actually, but good you added the neg tests...)

Well, it basically addresses exactly the source of inconsistency of these two tests. 
Now, it the inconsistency can be generated differently, I am not sure...

Issue #1566, opened 2019-11-08
CircleCi is not fetching the last fixpoint submodule


nikivazou:
E.g. FP is properly updated but develop is fetched by circle in the following commit: 

https://github.com/ucsd-progsys/liquidhaskell/tree/48f6f82b3bb93c09115fa3a0fe930e302add0a92

nikivazou:
Reported in stack: https://github.com/commercialhaskell/stack/issues/5080

adinapoli:
Hey @nikivazou! I have been looking at this issue and I hope I can say with a certain degree of certainty I have found the culprit.  To reproduce the issue, I have created my own minified Github project which uses a submodule, to mimic what `liquidhaskell` does here with `liquid-fixpoint`:

https://github.com/adinapoli/circle-ci-git-submodules

As you can see, the idea is simple: I depend on a separate [git project used as a submodule](https://github.com/adinapoli/circle-ci-mysubmodule) to provide a simple `exposed :: Int` function that I can use in my main project as part of a testsuite:

```haskell
module Main where

import Test.Tasty
import Test.Tasty.HUnit

import MySubmodule (exposed)

main :: IO ()
main = defaultMain tests

tests :: TestTree
tests = testGroup "Tests" [unitTests]

--
unitTests = testGroup "Unit tests"
  [ testCase "Exposed" $
      exposed `compare` 42 @?= EQ
  ]
``` 

This way, I can test what happens when I bump my submodule to increment the value of `exposed`, and still have CI report success or failure if the number matches what I'm expecting in the main module.

After that, I have configured CircleCI for `circle-ci-git-submodules` and then opened this [PR](https://github.com/adinapoli/circle-ci-git-submodules/pull/1), that essentially replicates the usual scenario where `liquid-fixpoint` is bumped as part of a particular branch within `liquidhaskell`. After that, I was successfully able to replicate the CI problem, as you can see for example as part of this build:

https://circleci.com/gh/adinapoli/circle-ci-git-submodules/12

[...]

PR #1574, opened 2019-11-26
Eliminate etas in reflection equations

nikivazou:


nikivazou:
Oh, I bet this will fail since fixpoint is not properly sync-ed because of #1566.
That is: we first need to merge the relevant fixpoint request, since circle will not pull the linked FP module.

Issue #1576, opened 2019-12-16
Detect unsound behavior


jprider63:
I have a bit of time, so I plan to work on this. My plan is to:

- Generate a warning when unsafe behavior is encountered. This will happen when:
  - Functions in a black list (including `undefined`) appear in an expression
  - Refinements are declared with `lazy` or `assume`
- This warning will be enabled by default and will have the flags `-Wdetect-unsafe` and `-Wno-detect-unsafe`.
- I might add a `-Werror` flag.
- I might add an `unsafe` field to spec files.

Are there other sources of unsoundness I'm not thinking of? 

I thought we had an ongoing list of potential improvements, but I couldn't find an existing issue. Perhaps it was only over email/in the proposal.

#1339

nikivazou:
You can also break termination checking when you have negative data type definitions, but this looks more complicated to deted

ranjitjhala:
This would be amazing thanks James!!

jprider63:
I have [some code](https://github.com/jprider63/liquidhaskell/tree/unsound.jp) that works toward checking a blacklist. I'm not sure how to check whether a `Var` is in the list though. It appears that the name and unique identifier of `undefined` is being modified. For example, this code generates the following trace output. Do you have any suggestions? 

```
g :: ()
g = undefined

h :: ()
h = let x = undefined in x
```

```
[02v]
d2p6: lq_anf$##7205759403792802844
[02v]
d2p8: lq_anf$##7205759403792802846
[02v]
d2pa: lq_anf$##7205759403792802848
[02v]
d2pc: lq_anf$##7205759403792802850
[02v]
d2pe: lq_anf$##7205759403792802852
[02v]
d2pg: lq_anf$##7205759403792802854
[02v]
d2pi: lq_anf$##7205759403792802856
[02v]
d2pk: lq_anf$##7205759403792802858
[02v]
d2pl: lq_anf$##7205759403792802859
[02v]
d2pj: lq_anf$##7205759403792802857
[02v]
d2ph: lq_anf$##7205759403792802855
[02v]
d2pf: lq_anf$##7205759403792802853
[02v]
d2pd: lq_anf$##7205759403792802851
[02v]
d2pb: lq_anf$##7205759403792802849
[02v]
d2p9: lq_anf$##7205759403792802847
[02v]
d2pm: lq_anf$##7205759403792802860
[02v]
d2p7: lq_anf$##7205759403792802845
[02v]
[...]

PR #1577, opened 2019-12-19
Unsound.jp

jprider63:
Implements #1576. It's not perfect, but is a step in the right direction.


facundominguez:
@jprider63 @nikivazou, would you like bringing this PR to completion, or should it be closed?

facundominguez:
:wave: I'm closing this after the long time passed. But please feel free to revive it.

PR #1578, opened 2019-12-21
Extensionality

nikivazou:
Automate reasoning about function extensionality, see https://github.com/ucsd-progsys/liquidhaskell/blob/880c78f94520d76fa13880eac050f21dacb592fd/tests/pos/T1577.hs


ranjitjhala:
Wow nice !!!

PR #1582, opened 2020-01-13
Fix CircleCI submodule weirdness

adinapoli:
Based on my investigation as part of #1566, this should resolve the relevant issue.

This commit installs the `ssh` package before calling `git submodule sync`, to resolve some misbehaviour observed with CircleCI as part of #1566.


nikivazou:
Thanks a lot!

PR #1585, opened 2020-01-15
Allow tests to be run via stack or cabal

### LINKED ISSUE #1584: Make tests not stack-dependent
Currently `mkTest` [assumes](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/test.hs#L462) that `liquid` has been installed/provisioned via `stack`, but this might not be true for users using `cabal`.

In particular, it took me a while to realise why on my machine calling `cabal new-test` was hanging indefinitely, and only when I (accidentally) tried to run `stack test` (so that the relevant folders were generated on my machine) that I have realised what was going on 😉 I can see how others might end up into the same pit.

I think it would be nice for users not being tied to a particular program, even though `stack` is the de facto for this project.

@nikivazou , @ranjitjhala Would you accept a pull request in this area or is this something that doesn't warrant the effort? Thanks!

adinapoli:
Fixes #1584.

This commit introduces a new test option called '--liquid-runner' that defaults to `stack exec -- liquid` as before but allow users to override the command which will eventually run `liquid`.

Doing so allows Cabal users to pass something like `cabal new-exec liquid -- ` to the test executable, for example running this:

```
/path/to/test --liquid-runner 'cabal new-exec liquid -- '
```

As I was at it, I have also switched from `LineBuffering` to `NoBuffering` in the implementation of `mkTest`, as that should be pretty harmless and generally recommended for these kind of long-running tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ css/
 *.smt2
 .liquid
 .dir-locals.el
+.tasty-rerun-log
--- a/tests/test.hs
+++ b/tests/test.hs
@@ -73,6 +73,7 @@ main = do unsetEnv "LIQUIDHASKELL_OPTS"
                 testRunner
               , includingOptions [ Option (Proxy :: Proxy NumThreads)
                                  , Option (Proxy :: Proxy LiquidOpts)
+                                 , Option (Proxy :: Proxy LiquidRunner)
                                  , Option (Proxy :: Proxy SmtSolver) ]
               ]
     tests = group "Tests" $ microTests  :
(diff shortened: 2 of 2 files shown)


adinapoli:
uh-oh, Appveyor seems unhappy but it seems to be unrelated to this PR:

```
xml             > copy/register
xml             > Installing library in C:\Users\appveyor\AppData\Roaming\stack\snapshots\700db043\lib\x86_64-windows-ghc-8.6.5\xml-1.3.14-ESp8cJy4XJ8BJKfvGeNymT
xml             > Registering library for xml-1.3.14..
Access violation in generated code when reading ffffffffffffffff
Command exited with code 1
```

nikivazou (commented):
Thanks! 
This is ready to merge, correct?

adinapoli:
> Thanks!
> This is ready to merge, correct?

@nikivazou Correct -- if you are happy with the PR, we can definitely merge!

PR #1590, opened 2020-01-22
Support for GHC 8.10

adinapoli:
This draft PR is an initial attempt to kickstart an effort to support GHC 8.10.

At the moment this should build just fine on a machine which has installed [GHC 8.10 RC1](https://www.haskell.org/ghc/blog/20200124-ghc-8.10.1-rc1-released.html). As part of this PR I have also backported my outstanding PR for #1584 , so that we can run the tests by doing something like:

```
cabal v2-test liquidhaskell:test --flag include --flag devel --test-options='--liquid-runner="cabal v2-exec liquid -- "'
cabal v2-test liquidhaskell:liquidhaskell-parser
```

In case the testsuite fails with an error like `CoreToLogic.lg: openFile: does not exist` that's because calling `cabal v2-build` doesn't install the `data-files` inside the `datadir`. To circumvent this it's enough to do something like this before calling the test:

```
export liquidhaskell_datadir=/path/to/liquidhaskell/include/directory
```

This branch is also backward compatible with GHC 8.6.5, so it should build with both Stack and Cabal.


adinapoli:
Depends on https://github.com/ucsd-progsys/liquid-fixpoint/pull/430 .

adinapoli:
Interestingly AppVeyor seems to fail, I do wonder if it's due to https://gitlab.haskell.org/ghc/ghc/issues/13112.

stack.yaml.lock
```diff
@@ -0,0 +1,100 @@
+# This file was autogenerated by Stack.
```
nikivazou, line 1: @adinapoli do we really need these files?

stack-8.6.5.yaml.lock
```diff
@@ -0,0 +1,40 @@
+# This file was autogenerated by Stack.
```
nikivazou, line 1: @adinapoli do we really need these files?

nikivazou (approved):

PR #1602, opened 2020-02-05
T1601

### LINKED ISSUE #1601: Reflection captures bound name
This is a bit of a frustrating bug report to write.

There is a problem with reflection that I've encountered several times, but it's unclear to me how to minimize my code to isolate the issue. I am writing a tool for verifying neural networks using Liquid Haskell. At the moment, [my project](https://github.com/wenkokke/leadbeater) fails to compile with the following error:
```
:1:1-1:1: Error
  elaborate PLE1 195 failed on:
    ...
    && Leadbeater.Network.norm == Leadbeater.LinearAlgebra.map lam lam_arg##1 : real . Leadbeater.Prelude.rdiv lam_arg##1 (Leadbeater.LinearAlgebra.sumPos xs)
    ...
  with error
      Unbound symbol xs --- perhaps you meant: fst ?
  in environment
    ...
```
I've omitted a huge number of lines, basically containing the reflection of every single line of Haskell in my current codebase.

The problem seemingly lies with [the following snippet](https://github.com/wenkokke/leadbeater/blob/master/src/Leadbeater/Network.hs#L60-L63):
```haskell
{-@ reflect norm @-}
{-@ norm :: xs:VectorNE Rpos -> VectorX R xs @-}
norm :: Vector R -> Vector R
norm xs = let s = sumPos xs in map (`rdiv` s) xs
```
It seems that Liquid Haskell somehow reflects the `xs` in the let binding as a literal `xs`. This problem remains if I rewrite the `norm` function to inline the definition for `s`, or to factor the lambda out into a top-level function:
```haskell
{-@ reflect norm @-}
{-@ norm :: xs:VectorNE Rpos -> VectorX R xs @-}
norm2 :: Vector R -> Vector R
norm2 xs = map (`rdiv` sumPos xs) xs
```
```haskell
{-@ reflect norm @-}
{-@ norm :: s:Rpos -> xs:VectorNE Rpos -> VectorX R xs @-}
norm3' :: R -> Vector R -> Vector R
norm3' s xs = map (`rdiv` s) xs

{-@ reflect norm3 @-}
{-@ norm3 :: xs:VectorNE Rpos -> VectorX R xs @-}
norm3 :: Vector R -> Vector R
norm3 xs = norm3' (sumPos xs) xs
```
This is where things get weird. If I only compile and verify the library, which _includes_ the `norm` function, everything checks just fine. The error above _only happens_ if I also compile and verify my test. The error itself is the failure to elaborate [the following definition](https://github.com/wenkokke/leadbeater/blob/master/test/perceptron.hs#L26-L37), which doesn't even use `norm`:
```haskell
{-@ reflect example @-}
{-@ example :: NetworkN 2 1 @-}
example :: Network
example = NLast
          (
            Layer { bias       = 0.184
                  , weights    = (1 >< 2)
                                 [ 0.194 , 0.195
                                 ]
                  , activation = Sigmoid
                  }
          )
```
(The only function in the above definition is [`><`](https://github.com/wenkokke/leadbeater/blob/master/src/Leadbeater/LinearAlgebra.hs#L137-L140), the rest are all constructors and constants.)

Similarly, if I extract the code involved in defining `norm`, as I've done [here](https://gist.github.com/wenkokke/8e46debb9d6e8177e007f1e545110f47), everything checks just fine...

If you have any idea what may be causing this, please let me know, I'd be happy to investigate more and provide further evidence.

ranjitjhala:
Fixes #1601 by disabling `elimEta` which is currently buggy. The PR guards `elimEta` under the _highly experimental_ `extensionality` flag. We can revisit the defaults when @nikivazou addresses the issue.

### DIFF
--- a/src/Language/Haskell/Liquid/Constraint/ToFixpoint.hs
+++ b/src/Language/Haskell/Liquid/Constraint/ToFixpoint.hs
@@ -47,7 +47,7 @@ fixConfig tgt cfg = def
   , FC.ginteractive     = ginteractive       cfg
   , FC.noslice          = noslice           cfg
   , FC.rewriteAxioms    = Config.allowPLE   cfg
-  , FC.etaElim          = not (exactDC cfg)
+  , FC.etaElim          = False -- not (exactDC cfg) -- SEE: https://github.com/ucsd-progsys/liquidhaskell/issues/1601
   , FC.extensionality   = extensionality    cfg 
   }
 
--- a/src/Language/Haskell/Liquid/Constraint/ToFixpoint.hs
+++ b/src/Language/Haskell/Liquid/Constraint/ToFixpoint.hs
@@ -47,7 +47,7 @@ fixConfig tgt cfg = def
   , FC.ginteractive     = ginteractive       cfg
   , FC.noslice          = noslice           cfg
   , FC.rewriteAxioms    = Config.allowPLE   cfg
-  , FC.etaElim          = False -- not (exactDC cfg) -- SEE: https://github.com/ucsd-progsys/liquidhaskell/issues/1601
+  , FC.etaElim          = not (exactDC cfg) && extensionality cfg -- SEE: https://github.com/ucsd-progsys/liquidhaskell/issues/1601
   , FC.extensionality   = extensionality    cfg 
   }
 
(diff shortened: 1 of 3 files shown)


nikivazou:
So the fixpoint elimEta is buggy? @ranjitjhala do you have a small file that triggers the bug?

PR #1606, opened 2020-02-08
Drop only dictionaries of type Eq, Ord, Numerical, and ~

yiyunliu:
Once typeclass support is properly implemented, we should no longer drop dictionaries (term level) and class constraints (type level), with the exception of  Eq, Ord, Numerical (Num, Integral, etc.).
While tests for class-laws would definitely break, I want use this PR to see if there's anything else that will be broken that I did not expect.


facundominguez:
@nikivazou do you know what is the status of this PR? Should it be closed?

facundominguez:
Closing this due to inactivity. Feel free to revive.

Issue #1604, opened 2020-02-06
Reflection causes type specifications to be interpreted as assumptions


wenkokke:
It seems that using reflection causes type specifications to be interpreted as assumptions. For instance, the following code is SAFE, according to Liquid Haskell:
```haskell
{-@ LIQUID "--reflection" @-}
{-@ LIQUID "--ple" @-}

data Val = V { val :: Int }
{-@ data Val = V { val :: Int } @-}
{-@ type ValN N = {v:Val | val v == N} @-}

{-@ reflect ex1 @-}
{-@ ex1 :: ValN 5 @-}
ex1 :: Val
ex1 = V 4

{-@ test1 :: {v:Bool | v} @-}
test1 = val ex1 == 6

{-@ test2 :: () -> {v:() | val ex1 == 6} @-}
test2 () = ()

{-@ test3 :: () -> {v:() | 1 == 2} @-}
test3 () = ()
```
My guess is that the type specification for `ex1` is interpreted as an assumption, instead of as something to be checked, which causes Liquid Haskell to deduce `4 == 5`, which causes the logic to become inconsistent. As you can see, we can then deduce pretty much anything. (I'm not sure which style of proof is preferred, so I'm giving both `test1` and `test2`.)

The attached code fails when we comment out the type specification for `ex1`, or when we comment out the `reflect` pragma (and `test2` and `test3`).

ranjitjhala:
This is most alarming! Will look at it ASAP! Thanks for letting us know!

ranjitjhala:
Specifically, I think your hunch is right: the signature *should* be
checked before “strengthening” with the body, but somehow that’s not
happening!

nikivazou:
Actually, the program is SAFE when you have `ple` on (not `reflection`). 
So, @ranjitjhala I suspect the ple knowledge adds an inconsistency; are you on it?

ranjitjhala:
It would be great if you can take a look — the bug is not in PLE it’s that,
as Wen says, afaict we don’t check the signatures of reflected functions.

ranjitjhala:
As in, yes you are right: ple is what is adding the falls in but the real
problem is not the adding of the falls in because that’s already there in
the spec it is that the spec is wrong and should be checked...

nikivazou:
@wenkokke should be fixed. Can you confirm? Thanks!!!

wenkokke:
[...]

PR #1607, opened 2020-02-09
fix for #T1604

nikivazou:
Indeed we were putting the reflected signatures in the fixpoint environments.
With this edit we do not.


nikivazou:
Fix for #1604

PR #1611, opened 2020-02-12
T1490

### LINKED ISSUE #1490: Add proper support for NewTypes / "Crash SMTLIB2: Sorts are incompatible"
I managed to get a relatively small repro example for something that seems to be related to #1467.
My `liquid` is from `406dd8bc198fad9a2ff87025689b6035e60fd9be`.

Notably if I remove the B newtype and use `Bool` directly in `A` it works fine!

Here's the example:
```
module A where

import Foreign.Storable (Storable, poke, pokeByteOff)

data A = A B

newtype B = B Bool

instance Storable A where
  poke ptr (A (B b)) = do
    (\hsc_ptr -> pokeByteOff hsc_ptr 16) ptr b
```

```
tauren :: /tmp/liquid » cat A.hs                                                                                                                                                                                                                                            
module A where

import Foreign.Storable (Storable, poke, pokeByteOff)

data A = A B

newtype B = B Bool

instance Storable A where
  poke ptr (A (B b)) = do
    (\hsc_ptr -> pokeByteOff hsc_ptr 16) ptr b
tauren :: /tmp/liquid » liquid A.hs
LiquidHaskell Version 0.8.6.0, Git revision 406dd8bc198fad9a2ff87025689b6035e60fd9be [develop@406dd8bc198fad9a2ff87025689b6035e60fd9be (Fri May 10 09:09:19 2019 -0700)]
Copyright 2013-19 Regents of the University of California. All Rights Reserved.

Targets: A.hs

**** [Checking: A.hs] **********************************************************

**** DONE:  A-Normalization ****************************************************


**** DONE:  Extracted Core using GHC *******************************************


**** DONE:  Transformed Core ***************************************************

Working  83% [=======================================================..........]

**** DONE:  annotate ***********************************************************


**** RESULT: ERROR *************************************************************


:1:1-1:1: Error
  crash: SMTLIB2 respSat = Error "line 382 column 1196: Sorts Bool and Int are incompatible"
```

nikivazou:
Fixes #1490 by replaces newtypes with their synonyms before translation to sorts.


yiyunliu:
Thank you so much! This PR works perfectly. Now I can finally get newtype instances to work in type class.

ranjitjhala:
Thanks Niki!!!

nikivazou:
There source of trouble here was that 
if `newtype X = C` and `{-@ embed X as int @-}` then we need to make an `int` embed of `C` as well (since now SMT sees `C` and not `X`). 
But for the latter embed to work `C` should be imported...

PR #1618, opened 2020-02-19
Patch include .hs files so that LH can check them.

adinapoli:
This commit tweaks the Haskell files at include/ so that liquid can check them, and for each of them can yield 'SAFE'. Unused files have been removed.

Interestingly enough for `Foreign.hs` I had to resort to a trick similar to what @nikivazou did [here](https://github.com/ucsd-progsys/liquidhaskell/commit/3c5e585363b0a08123f9f27030be75df2efb5287), i.e. bring the `Word64` in scope to witness that this can be treated as an int.


include/Language/Haskell/Liquid/List.hs
```diff
@@ -1,5 +1,6 @@
 module Language.Haskell.Liquid.List (transpose) where
 
+{-@ lazy transpose @-}
```
adinapoli, line 3: Intuitively here it should be possible to give a proper refinement type, for using `lazy` was, well, the quickest way to make this pass, hehe.

nikivazou (approved):

Issue #1619, opened 2020-02-22
Polymorphic instantiation seems broken with typeclasses


rosekunkel:
Liquid deems `fmapPreservesLabel` unsafe, but its equivalent using explicit dictionary passing, `myFmapPreservesLabel`, is deemed safe. They should both be safe, since `fmap`/`myFmap` should be instantiated at `Tagged Label<label>`.
```haskell
{-# LANGUAGE Rank2Types #-}

{-@ data Label<label :: User -> Bool> where @-}
data Label
data User
data Tagged l a = Tag a
data MyFunctor f = MyFunctor { myFmap :: forall a b. (a -> b) -> f a -> f b }

instance Functor (Tagged l) where
  fmap = fmapTagged

taggedFunctor :: MyFunctor (Tagged l)
taggedFunctor = MyFunctor fmapTagged

fmapTagged :: (a -> b) -> Tagged l a -> Tagged l b
fmapTagged f (Tag x) = Tag (f x)

{-@
myFmapPreservesLabel :: forall <label :: User -> Bool>.
(a -> b) -> Tagged (Label<label>) a -> Tagged (Label<label>) b
@-}
myFmapPreservesLabel :: (a -> b) -> Tagged Label a -> Tagged Label b
myFmapPreservesLabel = myFmap taggedFunctor

{-@
fmapPreservesLabel :: forall <label :: User -> Bool>.
(a -> b) -> Tagged (Label<label>) a -> Tagged (Label<label>) b
@-}
fmapPreservesLabel :: (a -> b) -> Tagged Label a -> Tagged Label b
fmapPreservesLabel = fmap
```

ranjitjhala:
I don’t think it’s the type class, more likely the HKT — LH is (probably)
not doing any polymorphic instantiation when argument used in the type
application is NOT of kind * Ie it ONLY does instantiation when the type IS
of kind * (that’s my guess.)


GHC desugars the fmap into

fmap @ Tagged Label @ a @ b

And my guess is LH is not doing any instantiation for the Tagged Label.

I wonder why the other version passes... maybe because LH instantiates the
l in the sig for the dictionary?

rosekunkel:
I should mention that this is currently blocking my attempt to rewrite Binah in the MAC style. (Well, not blocking exactly, but it means none of the code can use do-notation or monad transformers, which is very unfortunate).

ranjitjhala:
[...]

PR #1621, opened 2020-02-24
make functor a generic class for #1619

nikivazou:
So, in @rosekunkel 's example, the abstract refinement `label` was defaulted to true because `Functor` is not "generic", so no fresh type was created. 

@rosekunkel this fixes your issue, but if you want to have similar behavior on other type classes, you should make a similar edit and add, e.g., applicative, monads, etc in the generic set, like I did. 

@ranjitjhala I am not entirely sure that Functor (and the rest type constructors) are generic. It would be cool to formalize what generic means for such constructors. But the truth is, we are being too strict by not refreshing at all non generic types. I suspect we need to refresh the inner parts of the type (as here) but not the top level refinements. 

Anyway, this should unblock Rose.

### DIFF
--- a/src/Language/Haskell/Liquid/Constraint/Generate.hs
+++ b/src/Language/Haskell/Liquid/Constraint/Generate.hs
@@ -1458,14 +1458,13 @@ isPLETerm γ
 
 -- | @isGenericVar@ determines whether the @RTyVar@ has no class constraints
 isGenericVar :: RTyVar -> SpecType -> Bool
-isGenericVar α t =  all (\(c, α') -> (α'/=α) || isOrd c || isEq c ) (classConstrs t)
+isGenericVar α t =  all (\(c, α') -> (α'/=α) || isGenericClass c ) (classConstrs t)
   where 
     classConstrs t = [(c, ty_var_value α')
                         | (c, ts) <- tyClasses t
                         , t'      <- ts
                         , α'      <- freeTyVars t']
-    isOrd          = (ordClassName ==) . className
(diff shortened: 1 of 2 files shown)


ranjitjhala:
Instead of hacking “isGeneric” this way why Not use abstract refinements on
the signature of Fmap?

nikivazou:
Good question :) 
Did you try it?

nikivazou:
ok, @ranjitjhala you are suggesting to add an abstract refinement 
1. in a class method (I don't think we support that :) )[1]
2. in a type variable for a type constructor (I don't think we support that either)[2]

I suggest, if this blocking Rose to hack isGeneric, otherwise properly fix it but this would need time. 


[1] The following creates a parse error 
```
{-@ class Functor f where
  fmap :: forall <p :: f -> Prop>. (a -> b) -> f a -> f b
  (<$) :: forall a b. a -> f b -> f a
  @-}
``` 
[2] The following creates a parse error 
```
{-@   fmap'' :: forall <p :: f -> Bool>. (a -> b) -> (f<p>) a -> (f<p>) b @-}
fmap'' :: (a -> b) -> f a -> f b 
fmap''   = undefined
```

ranjitjhala:
How about just an assume for the signature of fmap?

nikivazou:
We still have problem number 2.

ranjitjhala:
@nikivazou why was this not merged in? (I can't tell by reading the comments but somehow this PR was closed?)

facundominguez:
:wave: As with other old PRs, let me know if you want to work on it to get it merged.

PR #1627, opened 2020-03-12
Add re-exported TyCons to gsTcs

adinapoli:
This PR is my attempt at fixing one of the issues me, @ranjitjhala and @nikivazou were discussing later this week, where using the experimental source plugin to compile a standalone version of `GHC.Types` was yielding an error like:

```
ghc: panic! (the 'impossible' happened)
  (GHC version 8.10.0.20200123:
        src/GHC/Types.spec:29:1-12: Error: Unknown variable `GHC.Types.W#`
    rawTySigs
```

I can give you here the TL;DR but I have notes of a more thorough debugging session that I can present if interested. The bottom line is that in order for [lookupGhcVar](https://github.com/ucsd-progsys/liquidhaskell/blob/5af4f07495348637c3c15ea18588963e77d143e9/src/Language/Haskell/Liquid/Bare/Resolve.hs#L430) to succeed, the [makeEnv](https://github.com/ucsd-progsys/liquidhaskell/blob/5af4f07495348637c3c15ea18588963e77d143e9/src/Language/Haskell/Liquid/Bare/Resolve.hs#L84) should include, for that particular `Symbol`, a GHC `Var`, and for the case of `W#` we didn't have one, but we had vars (under the form of "identifiers", i.e. `AnId`) for things like `GHC.Types.True` and `GHC.Types.False`. The reason was that deep down the guts of `makeEnv`, we call [srcTyCons](https://github.com/ucsd-progsys/liquidhaskell/blob/5af4f07495348637c3c15ea18588963e77d143e9/src/Language/Haskell/Liquid/Bare/Resolve.hs#L224) which reads a bunch of information from the input `GhcSrc`, tries to extract a list of `TyCon` and then all the gathered `Var`s out of it.

So what was happening was that for example symbols like `GHC.Types.Bool` were being resolved to a `TyCon`, and the call to [dcs](https://github.com/ucsd-progsys/liquidhaskell/blob/5af4f07495348637c3c15ea18588963e77d143e9/src/Language/Haskell/Liquid/Bare/Resolve.hs#L219) a few lines above was converting them into `GHC.Types.True` and `GHC.Types.False`. 

However, for `W#` this wasn't happening because `GHC.Types.Word` wasn't inside the [gsTcs](https://github.com/ucsd-progsys/liquidhaskell/blob/5af4f07495348637c3c15ea18588963e77d143e9/src/Language/Haskell/Liquid/Bare/Resolve.hs#L225) of the input `GhcSrc`, and this is because now in this new world we are not defining the types in the target module anymore. Remember that:

```haskell
module GHC.Types (module Exports) where

import GHC.Prim
import "ghc-prim" GHC.Types as Exports
```

Now, from the point of view of liquidhaskell (and of the input `ModGuts`) the content of `mg_tcs` is empty, because we do not define _any_ datatype in this module, because all we do is to re-export everything from the wrapped module!

**To solve this**, for now I have tried the simplest thing I could think of: I extend the final `gsTcs` with the _exported_ `TyCon`s, by looking at the content of `mg_exports`: that gives us a bunch of `Name`s we can resolve using the `lookupTyThings`, which I have generalised further.

Can you think about any potential shortcoming of using this approach that might lead to false positive / false negatives? So far the testsuite seems to be happy, but I cannot anticipate whether or not we might run into problems for more sophisticated programs.
 
For now I have marked this PR as a draft because while it is compiling, I wanted to discuss the general approach with you and, most importantly, test this extensively in my branch, but I wanted to get the discussion started. Obviously merging this PR in the first place should create (I hope) too much breakage.

### DIFF
--- a/src/Language/Haskell/Liquid/GHC/Interface.hs
+++ b/src/Language/Haskell/Liquid/GHC/Interface.hs
@@ -34,6 +34,7 @@ import GHC.Serialized
 
 import qualified Language.Haskell.Liquid.GHC.API as Ghc
 import Annotations
+import Avail
 import Class
 import CoreMonad
 import CoreSyn
@@ -452,7 +453,8 @@ makeGhcSrc cfg file typechecked modSum = do
   let dataCons       = concatMap (map dataConWorkId . tyConDataCons) (mgi_tcs modGuts)
   -- let defVs          = definedVars coreBinds
   (fiTcs, fiDcs)    <- liftIO $ makeFamInstEnv hscEnv 
-  things            <- lookupTyThings hscEnv typechecked modGuts 
+  things            <- lookupTyThings  hscEnv typechecked modGuts 
+  availableTcs      <- availableTyCons hscEnv typechecked (mg_exports modGuts')
   -- _                 <- liftIO $ print (showpp things)
   let impVars        = importVars coreBinds ++ classCons (mgi_cls_inst modGuts)
   incDir            <- liftIO $ Misc.getIncludeDir
@@ -466,7 +468,7 @@ makeGhcSrc cfg file typechecked modSum = do
     , giUseVars   = readVars coreBinds
     , giDerVars   = S.fromList (derivedVars cfg modGuts) 
     , gsExports   = mgi_exports  modGuts 
-    , gsTcs       = mgi_tcs      modGuts
+    , gsTcs       = nub $ (mgi_tcs      modGuts) ++ availableTcs
     , gsCls       = mgi_cls_inst modGuts 
     , gsFiTcs     = fiTcs 
     , gsFiDcs     = fiDcs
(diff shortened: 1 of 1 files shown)


nikivazou:
Great! This looks a much cleaner solution than the (hacky) ones we discussed on Tue! 
And, does not seem to break anything :)

nikivazou (approved):

ranjitjhala:
Super, lets merge when @adinapoli  is ready!

adinapoli:
@ranjitjhala I am happy for this strand of work to go it, with the only caveat being that I have later on (in `adinapoli/source-plugin`) further expanded on this approach also for other things (like typeclass instances, for example) that we needed for the "mirror package" approach, and might be nice to get those in as well.

What about I cherry-pick the relevant commits and add those as part of this PR, so that we can merge those and have a less gargantuan PR by the time the source plugin one lands? 😉

ranjitjhala:
Oh, please merge if tests pass!

adinapoli:
Not relevant anymore, closing it!

PR #1630, opened 2020-03-18
NEW PLE IMPLEMENTATION

nikivazou:
Following [the FP PR](https://github.com/ucsd-progsys/liquid-fixpoint/pull/426).  

The only non supported test is [MonadId.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/53fb63cf3e8d8e266ebb4879e1519b8538f8d22f/benchmarks/popl18/ple/pos/MonadId.hs) that is using lambdas and is flagged with `"--oldple"`.

Time comparison is [here](https://github.com/ucsd-progsys/liquidhaskell/blob/53fb63cf3e8d8e266ebb4879e1519b8538f8d22f/tests/logs/compare-PLE.csv). Check the last column for sorted. The implementation is critically changed (following the POPL description). There is critical slow down is some tests but speedup in others. 

I need to merge it because 
1. it is supporting [tests](https://github.com/ucsd-progsys/liquid-fixpoint/pull/422) not supported before. 
2. there is an important bug fix ([see this](https://github.com/ucsd-progsys/liquid-fixpoint/pull/423).
3. I now understand what is happening and can edit accordingly :) 

The old behavior is restored with `"--oldple"`.


ranjitjhala:
Niki I’ll try to review today but if I don’t go ahead and merge

PR #1634, opened 2020-03-30
expand type synonyms before body expansion

nikivazou:
see tests/pos/T1634.hs for a test failing before.


ranjitjhala:
nice :-) thanks Niki!!!

PR #1640, opened 2020-04-06
flag polymorphic instantiation

nikivazou:
Most equational style proofs are not using the advantages polymorphic inference (esp. the ones that reason about trivially refined functions). But these proofs still pay the inference extra time. 
In this PR I flag the polymorphic inference feature. 
The example I am working on drops from ` 379.96 s` to `5.66 s`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,16 @@ To allow reasoning about function extensionality use the `extensionality flag`. 
 {-@ LIQUID "--extensionality" @-}
 ```
 
+
+Fast Checking
+-------------
+
+The option `--fast` or `--nopolyinfer` greatly recudes verification time, can also reduces precision of type checking. 
+It, per module, deactivates inference of refinements during 
+instantiation of polymorphic type variables. 
+It is suggested to use on theorem proving style when reflected 
+functions are trivially refined. 
+
 Incremental Checking
 --------------------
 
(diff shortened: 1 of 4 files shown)


ranjitjhala:
Strange I thought I’d fixed this already a while ago. Let’s make sure not
to duplicate functionality?

nikivazou:
Hmmm, so if global PLE is on `gsAutoInst` is supposed to have all the top level binders, correct? 
I assume this is broken.

nikivazou:
ok, @ranjitjhala what you have implemented only deactivates polyinfer when 1/ple-local is on and 2/for definitions explicitly defined to use ple. 
When the global `ple` flag is on, polyinference is just on. 

Let me try to see what happens with the test when polyinference is deactivated when global PLE is on, but I suggest we keep these two features (1/polyinference and 2/ple) separated, because there are cases that need both (e.g., reflection of functions with refinements).

ranjitjhala:
Yes, I agree it's better to keep things separate - the key is not to
implement it twice.

nikivazou:
@ranjitjhala I deactivated the old functionality, that was only working on local PLE. 
The flag now is global and called `nopolyinfer` or `fast` (documented). 

BTW, if it funny, because if this feature was actually working before the unsoundness of extensionality would be hidden.

nikivazou:
Also, it would be easy to implement a local, per definition `fast` checking mode. 
So, let me know if you also want that feature.

Issue #1657, opened 2020-05-01
Unsoundness using bounded refinements


ranjitjhala:
Why does the following report safe?

```haskell
{-@ data I <p :: Int -> Bool> = I _ @-}
data I = I Int

{-@ getI :: forall <p :: Int -> Bool>. 
             { {x: Int<p> | True} <: {x:Int | x > 0} }
             I <p>
@-}
getI :: I
getI = undefined 

{-@ pleaseFail :: I<{\_ -> True}> @-}
pleaseFail :: I
pleaseFail = getI
```

The above is a simplification of a bigger example

```haskell
module State where 

-- | Generic State-Transformer ---------------------------------------------------------------- 
{-@ data variance ST covariant contravariant covariant @-}
{-@ data ST a <pre :: World -> Bool, post :: World -> Bool> = ST _ @-}
data ST a = ST (World -> (World, a))

-- | World is a record with one field -----------------------------------------------------------
data World = World 
  { fI :: Maybe Int
  }

{-@ measure mfI @-}
mfI :: World -> Maybe Int
mfI (World x) = x

{-@ getI :: forall <p :: World -> Bool>. 
              { {w: World<p> | True} <: {w:World | isJust (mfI w)} }
              ST <p, p> Int 
  @-}
getI :: ST Int 
getI = undefined 

{-@ type STop a = ST <{\w -> True}, {\w -> True}> a @-}
{-@ pleaseFail :: STop Int @-}
pleaseFail :: ST Int 
pleaseFail = getI
```

ranjitjhala:
More mystery

```haskell
{-@ data I <pigbert :: Int -> Bool> = I Int @-}

data I = I Int
{-@ getI :: forall <pp :: Int -> Bool>. 
             { bloop :: Int <pp> |- {v: Int | v = bloop} <: {v:Int | v > 1984} }
             I <pp>
@-}
getI :: I
getI = undefined 

-- { {v: (Int<p>) | True} <: {v:Int | v > 1984} }

{-@ pleaseFail :: I<{\_ -> True}> @-}
pleaseFail :: I
pleaseFail = getI
```

There is no *mention* of `1984` in the generated constraints. Ok, will let @nikivazou solve this puzzle...

nikivazou:
@ranjitjhala I am playing with variance (correcting it). 
Your original example should be SAFE: 

```
{-@ data I <p :: Int -> Bool> = I Int @-}
data I = I Int
```
[...]

PR #1658, opened 2020-05-04
Fix for #1657: check bounds on zero argument, bounded functions

nikivazou:
The problem is that in the new tests the bounded function has zero arguments. 
When we check bounds for functions with arguments, this code is used: https://github.com/ucsd-progsys/liquidhaskell/blob/develop/src/Language/Haskell/Liquid/Constraint/Generate.hs#L849 which adds this constraints by https://github.com/ucsd-progsys/liquidhaskell/blob/develop/src/Language/Haskell/Liquid/Constraint/Generate.hs#L1135 

When the bounded function has no arguments, the bounds are handled by this: https://github.com/ucsd-progsys/liquidhaskell/blob/develop/src/Language/Haskell/Liquid/Constraint/Generate.hs#L684
where `addPost` just dropped the bounds (encoded as `RRTy e r OCons t`). 

Two `STMonad` tests are now failing, thus moved them in `todo`.


ranjitjhala:
Thanks @nikivazou !!!

Looks like the actual edit was a single line in `addPost` right?

1. Any idea why the two (old) tests are failing? (I'm guessing both use `fresh` which is another example of a "zero-argument" function?)

2. Also, does the modified version of the `pleaseFail` tests where the precondition is added to `pleaseFail` succeed?

nikivazou:
> Thanks @nikivazou !!!
> 
> Looks like the actual edit was a single line in `addPost` right?

Yes! 

> 
> 1. Any idea why the two (old) tests are failing? (I'm guessing both use `fresh` which is another example of a "zero-argument" function?)

It is `fresh` indeed.

> 2. Also, does the modified version of the `pleaseFail` tests where the precondition is added to `pleaseFail` succeed?

I don't understand this question.

ranjitjhala:
I mean that in e.g. `T1657.hs` if we instead wrote the signature

```haskell
{-@ shouldPass :: I<{\z -> z > 0}> @-}
shouldPass :: I
shouldPass = getI 
```

then does it verify? (Because if not, it means we're still not handling the 0-arg terms properly ...)

nikivazou:
> ```haskell
> {-@ shouldPass :: I<{\z -> z > 0}> @-}
> shouldPass :: I
> shouldPass = getI 
> ```

No :( 
But it does if you add an extract `()` argument to `getI`. 
So, I guess something is still off and need to look more...

ranjitjhala:
Interesting... I bet the “fresh” tests also have the same problem.

ranjitjhala:
Thanks @nikivazou !!!

PR #1674, opened 2020-05-18
Minor fixes.

kosmikus:


adinapoli:
Taking the baton of the gatekeeper / merger as this PR targets `adinapoli/source-plugin` and _not_ `develop`.

PR #1683, opened 2020-06-03
Source plugin for LiquidHaskell

adinapoli:
This PR is the "MVP" (_Minimum Viable Product_) that allows LiquidHaskell to be used as a compiler plugin and integrated as part of the GHC compilation pipeline. It requires the companion `liquid-fixpoint` PR, see https://github.com/ucsd-progsys/liquid-fixpoint/pull/435.

While this PR could be merged as-it-is, I will try documenting below what is outstanding in terms of "bells & whistles". None of what's below (with the notable exception of the CI failure) should stop merging this.

## Ergonomic adjustments

As said previously, we currently have the plugin living at `Language.Haskell.Liquid.GHC.Plugin`, which makes perfect sense in terms of our internal module hierarchy but it's a bit of a mouthful for users. We should probably have just a `LiquidHaskell` module which re-exports the plugin, so that users can use it just by having `-fplugin=LiquidHaskell` in their cabal files. I will try adding this as extra comments to this PR.

## Documentation

I have added some basic documentation but more polishing is needed, but that can come after merging this initial strand of work. In particular, I have tried to revamp the current `README` to make people aware of the new plugin and the migration path, but you might want to further expand on that.

## CI

While CircleCI is happily passing, AppVeyor seems to be struggling, with a failure similar to the one described [here](https://github.com/commercialhaskell/stack/issues/3765) but none of the proposed fixes actually work. I have seen this in the past while working on the plugin and I band-aided that by adding `-fexternal-interpreter` when building the library. Recently, however, it seems like `liquid-ghc-prim` is failing to build:

```
liquid-ghc-prim> configure (lib)
liquid-ghc-prim> Configuring liquid-ghc-prim-0.8.6.0...
liquid-ghc-prim> build (lib)
liquid-ghc-prim> Preprocessing library for liquid-ghc-prim-0.8.6.0..
liquid-ghc-prim> Building library for liquid-ghc-prim-0.8.6.0..
liquid-ghc-prim> [1 of 5] Compiling GHC.Prim
liquid-ghc-prim> 
liquid-ghc-prim> [;1msrc\GHC\Prim.hs:17:1: [;1merror:[;1m[;1m
liquid-ghc-prim>     Could not find module `GHC.Prim'
liquid-ghc-prim>     There are files missing in the `ghc-prim-0.6.1' package,
liquid-ghc-prim>     try running 'ghc-pkg check'.
liquid-ghc-prim>     Use -v (or `:set -v` in ghci) to see a list of the files searched for.
liquid-ghc-prim> [;1m   |
liquid-ghc-prim> [;1m17 | [;1mimport "ghc-prim" GHC.Prim as Exports
liquid-ghc-prim> [;1m   |[;1m ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
liquid-ghc-prim> 
liquid-ghc-prim> Access violation in generated code when reading 0xfffffffffffffff8
liquid-ghc-prim> 
liquid-ghc-prim>  Attempting to reconstruct a stack trace...
liquid-ghc-prim> 
liquid-ghc-prim>    Frame	Code address
```

 It's not really clear to me if this is a stack-specific issue or just a misconfigured build configuration. I will try playing with it, although the feedback cycle is a bit slow as I can't test this locally (I am on Mac OS X and I don't have a Windows VM lying around). 

## Migration plan

Currently I had to leave the "hardcoded prelude" files, living inside the `include` directory, so that the old executable could work just fine. Having said that, we should lay out a migration plan where we phase out the legacy executable and we clean up the the codebase from all these redundant files.

## Liquid-platform packages

Currently, I have done my best to move virtually all the files from `include` into their `liquid-*` mirror package counterparts. Having said that, there are still packages which haven't been fully tested or completed (for example `liquid-bytestring`) so we might want to spend a bit of time there. Last but not least **we should probably move away from the "mega-repo" approach** as we are doing now but rather have the different `liquid-*` projects into separate repository. I think this will help with versioning and revisioning.


gridaphobe:
Thanks @adinapoli!! I've been arguing for ages that LiquidHaskell should be a GHC plugin but don't have the time anymore to do the work.

nikivazou (approved):

PR #1684, opened 2020-06-06
Improve documentation and test case for mutual recursion.

jprider63:
This slightly improves the documentation and test case for mutually recursive functions. This is beneficial because you don't need to add an additional argument, so the function types remain unchanged.


ranjitjhala:
I’ll let @nikivazou weigh in on this

nikivazou (approved):

PR #1691, opened 2020-06-24
Better support for error messages and warnings

adinapoli:
This PR extends LH to emit warnings and errors in the same format used by GHC, which enables it to be used in tools like `ghcid` (ghcide is currently not supported mainly due to https://gitlab.haskell.org/ghc/ghc/issues/18070).

The internals of `GHC.Plugins` have been simplified a bit, and truth to be told I have another branch [here](https://github.com/ucsd-progsys/liquidhaskell/tree/adinapoli/ghcide-experiment) that simplify things even further, but I will decide whether or not it makes sense to merge the changes here (inflating the size of this PR) or delivering it separately. 

I will write a bit more detailed breakdown of the PR next week, as well as rebasing it with `develop`, hoping that by the time I do that, the CI flakiness (not dependent from us) will be resolved 😉


ranjitjhala:
So we're waiting on circle to de-flake till next week, right?

nikivazou (approved):

ranjitjhala:
🥳 🎉

PR #1696, opened 2020-07-01
Add proper version to liquid-* packages

adinapoli:
This PR adds "proper" versions to the `liquid-*` packages, using the tentative versioning scheme me and @kosmikus proposed a while back. Trying to put the correct constraints on libraries always felt a balance between science and art to me, so I have done my best here to not restrict things too much, while at the same time not making things too loose.

In particular, compiler-dependent packages like `liquid-base` and `liquid-ghc-prim` have been versioned with bounds "in lockstep" to their underlying libraries. For things like `containers`, `bytestring` etc fortunately I could use the latest versions as they are "old enough" to be used in not-so-new Stackage LTSs, which means `liquid-platform` should ideally build for a variety of Stack configurations.

Things like `liquid-prelude` and `liquid-platform` have the same version used for `liquidhaskell`, as they are not really "mirror" packages.

### DIFF
--- a/liquid-base/liquid-base.cabal
+++ b/liquid-base/liquid-base.cabal
@@ -100,7 +100,7 @@ library
   build-depends:      base                 == 4.14.0.0
                     , liquid-ghc-prim
                     , liquidhaskell        >= 0.8.10.1
-  default-language:   Haskell98
+  default-language:   Haskell2010
   default-extensions: PackageImports
                       NoImplicitPrelude
   if impl(ghc >= 8.10)
--- a/liquid-base/liquid-base.cabal
+++ b/liquid-base/liquid-base.cabal
@@ -1,5 +1,5 @@
 name:               liquid-base
-version:            0.8.10.1
+version:            4.14.0.0
 synopsis:           Drop-in base replacement for LiquidHaskell
 description:        Drop-in base replacement for LiquidHaskell.
 license:            BSD3
@@ -97,7 +97,7 @@ library
                     Liquid.Prelude.NotReal
                     Prelude
   hs-source-dirs:     src
-  build-depends:      base                 >= 4.11.1.0 && < 5
+  build-depends:      base                 >= 4.14.0.0 && < 5
                     , liquid-ghc-prim
                     , liquidhaskell        >= 0.8.10.1
   default-language:   Haskell98
--- a/liquid-base/liquid-base.cabal
+++ b/liquid-base/liquid-base.cabal
@@ -97,7 +97,7 @@ library
                     Liquid.Prelude.NotReal
                     Prelude
   hs-source-dirs:     src
-  build-depends:      base                 >= 4.14.0.0 && < 5
+  build-depends:      base                 == 4.14.0.0
                     , liquid-ghc-prim
                     , liquidhaskell        >= 0.8.10.1
   default-language:   Haskell98
--- a/liquid-bytestring/liquid-bytestring.cabal
+++ b/liquid-bytestring/liquid-bytestring.cabal
@@ -30,7 +30,7 @@ library
   build-depends:      liquid-base          < 5
                     , bytestring           >= 0.10.10.0 && < 0.11
                     , liquidhaskell        >= 0.8.10.1
-  default-language:   Haskell98
+  default-language:   Haskell2010
   default-extensions: PackageImports
   if impl(ghc >= 8.10)
     ghc-options: -fplugin=LiquidHaskell
--- a/liquid-bytestring/liquid-bytestring.cabal
+++ b/liquid-bytestring/liquid-bytestring.cabal
@@ -1,5 +1,5 @@
 name:               liquid-bytestring
-version:            0.8.10.1
+version:            0.10.10.0
 synopsis:           LiquidHaskell specs for the bytestring package
 description:        LiquidHaskell specs for the bytestring package.
 license:            BSD3
@@ -27,8 +27,8 @@ library
                       Data.ByteString.Lazy
                       -- Data.ByteString.Lazy.Char8
   hs-source-dirs:     src
-  build-depends:      liquid-base          >= 0.8.10.1
-                    , bytestring           >= 0.10
+  build-depends:      liquid-base          < 5
+                    , bytestring           >= 0.10.0.0
                     , liquidhaskell        >= 0.8.10.1
   default-language:   Haskell98
   default-extensions: PackageImports
--- a/liquid-bytestring/liquid-bytestring.cabal
+++ b/liquid-bytestring/liquid-bytestring.cabal
@@ -28,7 +28,7 @@ library
                       -- Data.ByteString.Lazy.Char8
   hs-source-dirs:     src
   build-depends:      liquid-base          < 5
-                    , bytestring           >= 0.10.0.0
+                    , bytestring           >= 0.10.10.0 && < 0.11
                     , liquidhaskell        >= 0.8.10.1
   default-language:   Haskell98
   default-extensions: PackageImports
--- a/liquid-containers/liquid-containers.cabal
+++ b/liquid-containers/liquid-containers.cabal
@@ -19,7 +19,7 @@ library
                       Data.Map
   hs-source-dirs:     src
   build-depends:      liquid-base          < 5
-                    , containers           >= 0.6.0.0
+                    , containers           >= 0.6.2.1 && < 0.7
                     , liquidhaskell        >= 0.8.10.1
   default-language:   Haskell98
   default-extensions: PackageImports
--- a/liquid-containers/liquid-containers.cabal
+++ b/liquid-containers/liquid-containers.cabal
@@ -1,5 +1,5 @@
 name:               liquid-containers
-version:            0.8.10.1
+version:            0.6.2.1
 synopsis:           LiquidHaskell specs for the containers package
 description:        LiquidHaskell specs for the containers package.
 license:            BSD3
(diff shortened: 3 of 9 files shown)


liquid-base/liquid-base.cabal
```diff
@@ -97,7 +97,7 @@ library
                     Liquid.Prelude.NotReal
                     Prelude
   hs-source-dirs:     src
-  build-depends:      base                 >= 4.11.1.0 && < 5
+  build-depends:      base                 >= 4.14.0.0 && < 5
```
kosmikus: I would have expected `== 4.14.0.0`.

liquid-bytestring/liquid-bytestring.cabal
```diff
@@ -27,8 +27,8 @@ library
                       Data.ByteString.Lazy
                       -- Data.ByteString.Lazy.Char8
   hs-source-dirs:     src
-  build-depends:      liquid-base          >= 0.8.10.1
-                    , bytestring           >= 0.10
+  build-depends:      liquid-base          < 5
+                    , bytestring           >= 0.10.0.0
                     , liquidhaskell        >= 0.8.10.1
```
kosmikus, line 32: I would have expected `== 0.10.0.0`. This may be less important than for `base`. But there should definitely be an upper bound.

kosmikus: Also, why `0.0`? It should be `>= 0.10.10.0` at least.

liquid-containers/liquid-containers.cabal
```diff
@@ -18,8 +18,8 @@ library
   exposed-modules:    Data.Set
                       Data.Map
   hs-source-dirs:     src
-  build-depends:      liquid-base          >= 0.8.10.1
-                    , containers           >= 0.5.9.2
+  build-depends:      liquid-base          < 5
+                    , containers           >= 0.6.0.0
```
kosmikus: This should also be `== 0.6.2.1` or at least `>= 0.6.2.1` plus an upper bound.

liquid-ghc-prim/liquid-ghc-prim.cabal
```diff
@@ -27,7 +27,7 @@ library
                     GHC.Tuple
 
   hs-source-dirs:     src
-  build-depends:      ghc-prim
+  build-depends:      ghc-prim             >= 0.6.1 && < 0.7
```
kosmikus: This should be `== 0.6.1`.

liquid-parallel/liquid-parallel.cabal
```diff
@@ -17,8 +17,8 @@ data-files:           src/Control/Parallel/Strategies.spec
 library
   exposed-modules:    Control.Parallel.Strategies
   hs-source-dirs:     src
-  build-depends:      liquid-base          >= 0.8.10.1
-                    , parallel             >= 3.0.0.0
+  build-depends:      liquid-base          < 5
```
kosmikus: This should be `< 4.15`, because that's what's in `parallel`.

liquid-parallel/liquid-parallel.cabal
```diff
@@ -17,8 +17,8 @@ data-files:           src/Control/Parallel/Strategies.spec
...
+                    , parallel             >= 3.2.0.0
```
kosmikus: This should be `== 0.3.2.0` or `>= 0.3.2.0` plus an upper bound.

liquid-platform/liquid-platform.cabal
```diff
@@ -25,14 +25,14 @@ flag legacy
 
 executable liquid
   main-is:            src/Liquid.hs
-  build-depends:      liquid-base       >= 0.8.10.1
-                    , liquid-containers >= 0.8.10.1
+  build-depends:      liquid-base       >= 4.14.0.0 && < 5
+                    , liquid-containers >= 0.6.0.0  && < 7
```
kosmikus: Should be `< 0.7`.

liquid-platform/liquid-platform.cabal
```diff
@@ -25,14 +25,14 @@ flag legacy
...
                     , liquid-prelude    >= 0.8.10.1
-                    , liquid-vector     >= 0.8.10.1
-                    , liquid-bytestring >= 0.8.10.1
+                    , liquid-vector     >= 0.12.1.2 && < 0.13
+                    , liquid-bytestring >= 0.10.0.0 && < 0.10.11.0
```
kosmikus: Should be `< 0.11`?

liquid-platform/liquid-platform.cabal
```diff
@@ -25,14 +25,14 @@ flag legacy
...
                     , liquidhaskell     >= 0.8.10.1
-                    , process           >= 1.6.0.0
-                    , cmdargs           >= 0.10
+                    , process           >= 1.6.0.0 && < 1.7.0.0
```
kosmikus: Should be `< 1.7`.

liquid-platform/liquid-platform.cabal
```diff
@@ -25,14 +25,14 @@ flag legacy
...
+                    , cmdargs           >= 0.10    && < 0.12
```
kosmikus: Why `0.12` and not `0.11`?

liquid-platform/liquid-platform.cabal
```diff
@@ -25,14 +25,14 @@ flag legacy
...
   default-language:   Haskell98
   ghc-options:        -W -O2 -threaded -fdefer-typed-holes
```
kosmikus: `-fdefer-typed-holes` does not belong in a release package.

liquid-base/liquid-base.cabal
```diff
@@ -97,7 +97,7 @@ library
...
                     , liquid-ghc-prim
                     , liquidhaskell        >= 0.8.10.1
   default-language:   Haskell98
```
kosmikus: Why `Haskell98` and not `Haskell2010`?

liquid-prelude/liquid-prelude.cabal
```diff
@@ -25,9 +25,9 @@ library
                     Language.Haskell.Liquid.Synthesize.Error
                     KMeansHelper
   hs-source-dirs:     src
-  build-depends:      liquid-base          >= 0.8.10.1
-                    , bytestring           >= 0.10
-                    , containers           >= 0.5
+  build-depends:      liquid-base          < 5
+                    , bytestring           >= 0.10.0.0
```
kosmikus: Should have an upper bound.

liquid-prelude/liquid-prelude.cabal
```diff
@@ -25,9 +25,9 @@ library
...
+                    , containers           >= 0.5.0.0
```
kosmikus: Should have an upper bound.

liquid-vector/liquid-vector.cabal
```diff
@@ -22,8 +22,8 @@ library
                       Data.Vector.Internal.Check
                       Data.Vector.Unboxed.Mutable
   hs-source-dirs:     src
-  build-depends:      liquid-base          >= 0.8.10.1
-                    , vector               >= 0.10
+  build-depends:      liquid-base          < 5
+                    , vector               >= 0.12.1.2 && < 0.13.0.0
```
kosmikus: Should be `< 0.13`.

liquid-vector/liquid-vector.cabal
```diff
@@ -22,8 +22,8 @@ library
                       Data.Vector.Internal.Check
                       Data.Vector.Unboxed.Mutable
   hs-source-dirs:     src
-  build-depends:      liquid-base          >= 0.8.10.1
-                    , vector               >= 0.10
+  build-depends:      liquid-base          < 5
```
kosmikus: Should be `< 4.15` because that's what `vector` has.

liquid-platform/liquid-platform.cabal
adinapoli: Yes, it's what the old executable used to provide, and this got copy/pasted as a result. I'll remove it.

liquid-base/liquid-base.cabal
adinapoli: I think this was a copy/paste of some other template I had lying around. I see no reason why we can't use `Haskell2010` here.

liquid-vector/liquid-vector.cabal
adinapoli: Yes, this (and all other comments) are fair enough. As said, I don't have a strong intuition whether putting too-stringent constraints might prevent people from actually experimenting with `liquidhaskell`, i.e. a wanted to avoid restricting ourselves into "this build only with this precise constellation of packages / LTS version".

I guess I will amend the PR with your suggestions and we'll see what happens once we release 😸

liquid-platform/liquid-platform.cabal
adinapoli: Ops, embarrassing, sorry 😉

liquid-vector/liquid-vector.cabal
kosmikus: Well, I think that if you want to make these more liberal, that would make sense if there was a Stackage snapshot that has less recent versions, but then I'd probably still make the LH version number equal to the lowest "supported" one. And I think it's fair to be flexible in the version bounds according to PVP, but I would not go further. It's easier to make revisions and relax them later than have breaking builds due to missing them.

nikivazou (approved):

PR #1699, opened 2020-07-06
Support ghcide (with a caveat)

adinapoli:
This PR adds support for integrating LH into `ghcide`. More testing is needed (I think @kosmikus is also trying this out) but this should now support emitting errors and warnings as part of a normal `ghcide` session.

There is a _catch_, however: in order for this to work, we would need a [special version of GHC 8.10.1](https://gitlab.haskell.org/ghc/ghc/-/issues/18070#note_286552), which is not something we can easily provide (however, we might be able to push for a `8.10.2` release if we find enough evidence that this special GHC version indeed fixes some plugin-related woes).

Armed with the code in this branch and this special GHC version, I was finally able to convince `ghcide` to refine my code:

 
Screenshot 2020-07-06 at 12 06 07

### DIFF
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -69,7 +69,7 @@ flag deterministic-profiling
 flag no-plugin
   default:     False
   manual:      True
-  description: Disables the use of the GHC plugin. Use the legacy executable for testing.
+  description: Use the legacy executable for testing.
 
 library
   exposed-modules:    Gradual.Concretize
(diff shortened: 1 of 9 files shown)


adinapoli:
Ps. I think this PR is worth merging _regardless_ of the `ghcide` business, as it introduces some simplifications in the `CmdLine` and `GHC.Plugin` modules.

ranjitjhala:
Lets merge it ... I'm assuming it doesn't force this ghc requirement for
all users? only those who want to use ghcide?

adinapoli:
> Lets merge it ... I'm assuming it doesn't force this ghc requirement for
all users? only those who want to use ghcide?

Aye, correct 😸

PR #1703, opened 2020-07-13
Test case improvements; layout-predicting changes.

kosmikus:
This is a part of the parser-related changes which I think we could consider merging right now. This does not contain any actual changes to the parser, but it changes libraries and test cases in such a way that they *will* be compatible with the new parser as currently proposed.

As the new parser introduces layout-based parsing, many declarations in the current file require some changes in indentation.

In addition, this PR contains some fixes to negative test cases that I discovered to be failing for reasons probably unrelated to their original intent, such as due to parse errors or no longer existing flags.

### DIFF
--- a/benchmarks/popl18/nople/neg/BasicLambdas.hs
+++ b/benchmarks/popl18/nople/neg/BasicLambdas.hs
@@ -1,6 +1,6 @@
 
 {-@ LIQUID "--higherorder"      @-}
-{-@ LIQUID "--autoproofs"      @-}
+{- LIQUID "--autoproofs"      @-}
 {-@ LIQUID "--exact-data-cons" @-}
 module Append where
 
--- a/include/Data/Either.spec
+++ b/include/Data/Either.spec
@@ -1,5 +1,5 @@
 module spec Data.Either where
 
 measure isLeft :: Data.Either.Either a b -> Bool
-isLeft (Left x)  = true
-isLeft (Right x) = false
+  isLeft (Left x)  = true
+  isLeft (Right x) = false
--- a/liquid-base/src/GHC/List.spec
+++ b/liquid-base/src/GHC/List.spec
@@ -53,7 +53,7 @@ qualif LenSum(v:[a], xs:[b], ys:[c]): len([v]) = (len([xs]) - len([ys]))
 GHC.List.!!         :: xs:[a] -> {v: _ | ((0 <= v) && (v < len(xs)))} -> a
 
 
- zip :: xs : [a] -> ys:[b] 
+zip :: xs : [a] -> ys:[b] 
             -> {v : [(a, b)] | ((((len v) <= (len xs)) && ((len v) <= (len ys)))
             && (((len xs) = (len ys)) => ((len v) = (len xs))) )}
 
--- a/tests/absref/pos/ListISort-LType.hs
+++ b/tests/absref/pos/ListISort-LType.hs
@@ -7,8 +7,8 @@ import Language.Haskell.Liquid.Prelude
 --------------------------------------------------------------------
 
 {-@ data List [llen] a <p :: x0:a -> x1:a -> Bool>  
-  = Nil 
-  | Cons { lHd :: a, lTl :: List <p> (a <p lHd>) }
+      = Nil 
+      | Cons { lHd :: a, lTl :: List <p> (a <p lHd>) }
   @-}
 
 {-@ measure llen @-}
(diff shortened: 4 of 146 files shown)


include/Data/Either.spec
```diff
@@ -1,5 +1,5 @@
 module spec Data.Either where
 
 measure isLeft :: Data.Either.Either a b -> Bool
```
ranjitjhala, line 3: These are especially important as its new syntax -- I will need to update various bits of documentation appropriately!

liquid-base/src/GHC/List.spec
```diff
@@ -53,7 +53,7 @@ qualif LenSum(v:[a], xs:[b], ys:[c]): len([v]) = (len([xs]) - len([ys]))
 GHC.List.!!         :: xs:[a] -> {v: _ | ((0 <= v) && (v < len(xs)))} -> a
 
 
- zip :: xs : [a] -> ys:[b] 
+zip :: xs : [a] -> ys:[b] 
             -> {v : [(a, b)] | ((((len v) <= (len xs)) && ((len v) <= (len ys)))
             && (((len xs) = (len ys)) => ((len v) = (len xs))) )}
```
ranjitjhala, line 58: Nice! I guess we don't have to indent the `&&` because its inside the `{ ... | ... }`

tests/absref/pos/ListISort-LType.hs
```diff
@@ -7,8 +7,8 @@ import Language.Haskell.Liquid.Prelude
 --------------------------------------------------------------------
 
 {-@ data List [llen] a <p :: x0:a -> x1:a -> Bool>  
-  = Nil 
-  | Cons { lHd :: a, lTl :: List <p> (a <p lHd>) }
+      = Nil
```
ranjitjhala, line 10: Beautiful!!!

ranjitjhala (approved):
Wow, this is a most epic PR! Thanks @kosmikus !!!

benchmarks/popl18/nople/neg/BasicLambdas.hs
```diff
@@ -1,6 +1,6 @@
 
 {-@ LIQUID "--higherorder"      @-}
-{-@ LIQUID "--autoproofs"      @-}
+{- LIQUID "--autoproofs"      @-}
```
adinapoli, line 3: Unrelated to this PR: I was never able to understand if such "unclosed" pragmas were:

* A genuine typo (i.e. somebody forgot the opening `@`);
* A leftover of the developer working on that test that needed a quick way to enable and disable the pragma,
  but eventually forgot about it.

I am wondering if we should somehow detect this (i.e. we have a block comment with a trailing `@-}` ) and emit
some sort of warning like "warning: undelimited LiquidHaskell pragma `--autoproofs`".

include/Data/Either.spec
adinapoli: Indeed -- I guess we should state straight away that the parser is now layout-sensitive and therefore users are expected to format their specs appropriately, as they would do with "normal" Haskell programs.

adinapoli (approved):
LGTM!

include/Data/Either.spec
kosmikus: `measure`s are probably the most controversial aspect of the new parser. One could argue that a Haskell function definition is not indented wrt its type signature ...

So for measures, you can currently (with the new parser) do either
```
measure
  isLeft :: Data.Either.Either a b -> Bool
  isLeft ... = ...
  isLeft ... = ...
```
or
```
measure isLeft :: Data.Either.Either a b -> Bool
  isLeft ... = ...
  isLeft ... = ...
```

It is perhaps possible to even allow
```
measure isLeft :: Data.Either.Either a b -> Bool
isLeft ... = ...
isLeft ... = ...
```
but it requires a bit of additional work, and I am not sure whether it's worth it. If you think it is, I can look into it.

liquid-base/src/GHC/List.spec
kosmikus: Indentation is declaration-level, much like in Haskell.

The `zip` type signature must start at the same column as the other declarations in the block (that's the actual change on this line). The lines that belong to the `zip` type signature must be indented more than the first line of the declaration. But the actual indentation of lines 57 and 58 does not matter here, as long as it's more than that of line 56.

In Haskell, you can also do
```
x = 7
              + 10
      + 3
          + 5
```
and it is all fine, as long as none of these lines start on the first column (or even to the left of the `x` token).

benchmarks/popl18/nople/neg/BasicLambdas.hs
kosmikus: Yes, it can be argued the flag should just be removed. I've kept it in the case someone who knows better than me what `autoproofs` is or was might still want to look at it in this context. Perhaps other occurrences of this pattern have similar history. It's certainly not good, and if @ranjitjhala or @nikivazou can say with confidence what the proper way for this negative test case to look like is now that `--autoproofs` does no longer exist, we can change it.

ranjitjhala: Hi all, I think `autoproofs` should be replaced with `ple`

ranjitjhala:
This is fantastic -- I *like* the new indented measure syntax.

Issue #1531, opened 2019-07-14
Datatype parsing error with `=` on a new line


vyorkin:
Looks like the LH parser doesn't like when `=` is placed on the next line, for example:

```haskell
> {-@
> data IncList a
>   = Emp
>   | (:<) { hd :: a, tl :: IncList {v:a | hd <= v} }
> @-}
```

But it doesn't complain when `=` is on the same line

```haskell
> {-@
> data IncList a =
>     Emp
>   | (:<) { hd :: a, tl :: IncList {v:a | hd <= v} }
> @-}
```

Might be related to #873

kosmikus:
Fixed via #1704.

Issue #1481, opened 2019-04-28
Cannot parse refinements definition on a type class 


kubum:
For example:

```haskell
{-# LANGUAGE MultiParamTypeClasses #-}

{-@ 
class Monad m => Filter filter m where
  q :: forall a . Int -> Int -> Int -> m (filter a)
  qq :: forall a . filter a -> m Int
@-}
class Monad m => Filter filter m where
  q :: Int -> Int -> Int -> m (filter a)
  qq :: filter a -> m Int
```

Results in the error

```bash
8 |   qq :: forall a . filter a -> m Int
          ^

     unexpected ':'
     expecting bareTyArgP, "->", "=>", "(", white space or end of input
```

If we delete `qq`, liquidhaskell 0.8.2.4 parses it all fine.

According to the Basics.lhs example the functionality is supported [1], but it is tricky to see why the parser is unhappy.

[1] https://github.com/ucsd-progsys/liquidhaskell/blob/develop/docs/slides/HS2014/Basics.lhs#L209-L217

ranjitjhala:
Can you remove the `Monad m => ` constraint from the _refined_ class definition?

ranjitjhala:
no that's not it...

ranjitjhala:
Hmm, not sure what's going on here - however, 
note that your spec doesn't actually put any 
refinements on the methods, so as such you 
can omit it?

However, I did try to put some refinements in 
which case it crashes LH. Thanks for posting 
this, we'll take a look.

(For now, I recommend working around this by
pulling the methods out as plain functions and
putting the refinements on them...)

kubum:
Hi @ranjitjhala! Thank you for the response, yeah I found in some test cases there was used `;` to satisfy parser, but it didn't work in this case.

How do you see the extraction? I did it for the instances, a boring example might look as follows:

```haskell
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE FlexibleInstances #-}

class Monad m => Filter filter m where
  q :: Int -> Int -> Int -> m (filter a)

data Table a = A | B

{-@ type PositiveInt = { v: Int | v > 0 } @-}

instance Monad m => Filter Table m where
  {-@
  instance Monad m => Filter Table m where
    q :: PositiveInt -> PositiveInt -> PositiveInt -> m (Table a)
  @-}
  q n1 n2 n3 = pure $ A
```

Result:

```bash
 13 | instance Monad m => Filter Table m where
[...]

Issue #1012, opened 2017-05-01
Parsing Error in .spec files


nikivazou:
Measures that return type application are not parsed properly in `.spec` files. See #994 for details, where the following definitions are mis-parsed.
```
measure select_Product_1 :: Product f g p -> (f p)
measure select_Product_2 :: Product f g p -> (g p)
```

kosmikus:
Fixed via #1704.

Issue #924, opened 2017-01-16
Error: Specified Type Does Not Refine Haskell Type for Data.ByteString.elemIndex


fhaust:
I compiled *Liquid Haskell* from commit a63ad626cfb81163c16f43b0f585f6331111b7f6 (giving me a kind `liquidhaskell-0.7.0.0` installation) . Converted an existing project to use `liquidhaskell-cabal` and just ran the build process. Liquid Haskell reports:

```
     /home/florian/Dokumente/Projects/Haskell/liquidhaskell/.stack-work/install/x86_64-linux/nightly-2016-05-21/7.10.3/share/x86_64-linux-ghc-7.10.3/liquidhaskell-0.7.0.0/include/Data/ByteString.spec:310:8: Error: Specified Type Does Not Refine Haskell Type for Data.ByteString.elemIndex
     
     310 |     :: Data.Word.Word8
                  ^
     
     The Liquid type
      
         GHC.Word.Word8
    -> Data.ByteString.Internal.ByteString
    -> GHC.Base.Maybe GHC.Types.Int elemIndices
      
     is inconsistent with the Haskell type
      
         GHC.Word.Word8
    -> Data.ByteString.Internal.ByteString
    -> GHC.Base.Maybe GHC.Types.Int
      
     defined at <no location info>
```

Not sure this is supposed to happen, `elemIndexEnd` is the first name from the [next non whitespace line](https://github.com/ucsd-progsys/liquidhaskell/blob/master/include/Data/ByteString.spec#L318). 

I am in the process of running `stack test` on the LH project and so far it hasn't reported any errors, so I would assume that a simple parsing bug is unlikely?

ranjitjhala:
Ugh. I think it *is* a parsing bug -- @gridaphobe I thought this one was fixed a long time ago?

ranjitjhala:
Can you check if changing line 304 in include/ByteString.spec to:

```haskell
index :: bs : Data.ByteString.ByteString -> { n : Int | 0 <= n && n < bslen bs } -> Data.Word.Word8
```

fixes it for now? (Or else post as a gist the file that is tickling this bug?)

ranjitjhala:
related: 

https://github.com/ucsd-progsys/liquidhaskell/issues/707

fhaust:
`stack test` finished successfully.

Changing that line does not help. Nor does changing the lines above and below it.

[...]

PR #1704, opened 2020-07-13
Parser rewrite, adding layout-sensitive parsing.

kosmikus:
[ matches PR https://github.com/ucsd-progsys/liquid-fixpoint/pull/444 ]

 ** @kosmikus suggests to merge this _after_ the new 8.10 release**

This builds on #1703 and contains the actual changes to the parser.
There is still a lot more to do. I have created a project
(https://github.com/ucsd-progsys/liquidhaskell/projects/4)
that shows known parser bugs and how they are affected by
this PR.

This should probably only be merged after the plugin release ...

The most important changes are:

- Rewrite to use megaparsec rather than parsec, according
  to similar changes applied to liquid-fixpoint.

- Add layout-sensitive parsing. This makes the parser more
  strict in some situations. In particular, multiple lines
  that belong to a single LH "command" now must be indented
  in such a way that subsequent lines start at a later column
  than the initial line. The benefit is that multiple commands
  can be included in a single spec comment, and that line breaks
  work as expected in all situations, i.e. anything can be
  split across multiple lines. Previously, there was a hack that
  disallows line-breaks in some situations to ensure that
  some mult-line directives such as e.g. (instance) measures
  are parsed correctly. Also, explicit block parsing via
  braces and semicolons is now possible, so layout is
  optional.

- Error message reporting range calculations should be
  somewhat improved, although more remains to be done in
  this area.

- An error in datatype parsing is fixed which meant that
  datatypes without record fields are often incorrectly
  parsed (#871).

- Other open issues fixed by these changes are
  #707, #788, #873, #892, #924, #1012, #1481 and #1531.

- The test system is improved so that negative/positive
  tests do not purely look for the exit code, but check that
  there was not simply a parse error. This should improve
  the value of negative tests.

- The test system is improved so that we can test spec-file
  parsing individually. This is used to demonstrate that
  the situations in which previously redundant parentheses
  needed to be added in order to make the parser recognise
  the end of a declaration correctly are gone.

- Add test cases for #338 and #884, which I think have
  been working already in the current version, independently
  of the parser changes.

- Add some more comments.


kosmikus:
This has now been updated / rebased on latest develop. Depends on ucsd-progsys/liquid-fixpoint#444.

kosmikus:
My current analysis is that the `cabal_810` CI failure is a dependency glitch that is a real problem, but one that in principle exists independently of this PR.

nikivazou:
Is it difficult to add the missing instance? 
The gradual functionality (and the relevant code) is not used tbh. So, you can add a dummy instance and soon I can clean all this unused/gradual code.

kosmikus:
@nikivazou There is no missing instance, really. It's a dependency problem that can occur due to unfortunately imprecise dependency bounds in various places. The stack builds which have locked dependencies are all ok. Alfredo and I will fix this soon.

ranjitjhala (approved):

PR #1705, opened 2020-07-14
Try resolving AConLike for Symbols

### LINKED ISSUE #1688: Dummy imports needed to resolve names from imported modules
@adinapoli and @kosmikus here is a minimized example of the "dummy imports" problem.

- `module Lib` defines a datatype `T` with constructor `C` and either _reflects_ a function foo (or makes another datatype `T2` with ctor `C2`) whose refinement uses C;

- `module Client` imports `Lib` but LH crashes when resolving the reflected definition of `foo` or `C2` because the name `C` is not in scope :-(

We have been working around this by forcibly-importing C by adding a line

dummy = (C, ...)

but this is painful at the least... Any ideas?

## Lib.hs

```haskell
{-# LANGUAGE GADTs #-}

{-@ LIQUID "--reflection"  @-}

module Lib where

{-@ measure propOf :: a -> b @-}
{-@ type ProofOf E = { proofObj:_ | propOf proofObj = E } @-}

data Expr = Bc Bool                   -- True, False
          | Ic Int                    -- 0, 1, 2, ...
          | Lambda Int Expr         -- \x.e          abstractions
          | App Expr Expr             -- e e'          applications

data Basic = TBool         -- Bool
           | TInt          -- Int

data StepProp where
    Step :: Expr -> Expr -> StepProp

data StepProof where
    EFake :: Vname -> Expr -> Expr -> StepProof

{-@ data StepProof where
    EFake :: x:Vname -> e:Expr -> v:_ -> ProofOf( Step (App (Lambda x e) v) e)
  @-}
```

## Client.hs

```haskell
{-# LANGUAGE GADTs #-}

{-@ LIQUID "--reflection"  @-}

module Client where

import Lib

-- PROBLEM: If we comment `dummy` then `Step` is not in scope and the resolution of `EFake` fails.
dummy = (\x y -> Step x y)

data HasFType where
    FTBC   :: Bool -> HasFType

{-@ ftypSize :: HasFType -> { n:Int | n >= 0 } @-}
ftypSize :: HasFType -> Int
ftypSize (FTBC {}) = 1
```

This is unfortunate. 

If we are loading `Lib` then why can't we also load the
constructor name `Step` in automatically? Presumably this constructor/name
also lives in the serialized spec? (Problem is the GHC doesn't know this name
when it is compiling `Client`?)

adinapoli:
Fixes #1688. 

In case we need to qualify a Symbol, we would call the `ResolveSym` instance defined for a GHC.Var, which would look only at Identifiers (i.e. only `Id`s in a TyThing).

This PR adds an extra fallback case to look also for identifiers associated to `AConLike`.

In a nutshell, this allows certain type synonyms to be correctly expanded _in the lib module_ rather than in the client module when it's too late.

As discussed with @ranjitjhala (and as the branch name suggests) this is a bit of a "quick hack" to get this working, and while it should be harmless for existing code, it's not the "morally correct" fix. On the other hand, such "morally correct" fix would entail a substantial rewrite of the name resolution engine, which would be more effort than the few lines this PR costed me.

### DIFF
--- a/src/Language/Haskell/Liquid/Bare/Resolve.hs
+++ b/src/Language/Haskell/Liquid/Bare/Resolve.hs
@@ -11,7 +11,6 @@
 {-# LANGUAGE PartialTypeSignatures #-}
 {-# LANGUAGE ConstraintKinds       #-}
 {-# LANGUAGE TupleSections         #-}
-{-# LANGUAGE TypeApplications      #-}
 
 module Language.Haskell.Liquid.Bare.Resolve 
   ( -- * Creating the Environment
--- a/src/Language/Haskell/Liquid/Bare/Resolve.hs
+++ b/src/Language/Haskell/Liquid/Bare/Resolve.hs
@@ -12,7 +12,6 @@
 {-# LANGUAGE ConstraintKinds       #-}
 {-# LANGUAGE TupleSections         #-}
 {-# LANGUAGE TypeApplications      #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 module Language.Haskell.Liquid.Bare.Resolve 
   ( -- * Creating the Environment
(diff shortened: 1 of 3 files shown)


src/Language/Haskell/Liquid/Bare/Resolve.hs
```diff
@@ -535,11 +537,24 @@ instance ResolveSym Ghc.DataCon where
...
-  resolveLocSym env name _ lx = case resolveLocSym env name "Var" lx of 
-    Left _               -> Right (val lx)
+instance ResolveSym F.Symbol where
+  resolveLocSym env name _ lx = case resolveLocSym env name "Var" lx of
+    -- If we can't resolve the input 'Symbol' from an 'Id', try again
+    -- by grabbing the 'Id' of an 'AConLike', if any.
+    -- See test T1688.
+    Left _               -> case resolveLocSym env name "Var" lx of
+      Left _  -> Right (val lx)
+      Right v -> Right (F.symbol $ unWrap (v :: Wrapped Ghc.Var))
```
kosmikus: Could we perhaps use this opportunity to actually document what is going on here? Write a comment for the `ResolveSym` class, and describe the entire logic in the `F.Symbol` case?

src/Language/Haskell/Liquid/Bare/Resolve.hs
```diff
@@ -535,11 +537,24 @@ instance ResolveSym Ghc.DataCon where
...
     Right (v :: Ghc.Var) -> Right (F.symbol v)
 
+newtype Wrapped a = Wrapped { unWrap :: a } deriving (PPrint)
+
+instance ResolveSym (Wrapped Ghc.Var) where
+  resolveLocSym = resolveWith "variable" $ \case
+                    Ghc.AConLike (Ghc.RealDataCon x) -> Just $ Wrapped (Ghc.dataConWorkId x)
+                    _                                -> Nothing
```
kosmikus: Again, could we document what the difference in practice is between this case and the `DataCon` case?

Independently, I don't see the value of introducing a dummy type constructor (`Wrapped`) just so that we can make this an instance, which then presumably is called in just a single place in the program, namely in line 547 above? So why not inline this code into the `Symbol` instance?

kosmikus (changes requested):

kosmikus (approved):
Thanks. Much better now, at least in my opinion.

ranjitjhala (approved):
Excellent suggestion by @kosmikus -- and thanks a ton for adding the explanation @adinapoli !!!

PR #1706, opened 2020-07-14
Liquid-Dev-Mode prototype

adinapoli:
_This PR has to be intended as a prototype and can be easily improved. To name something suboptimal, the current `Setup.hs` is the same for each package and is duplicated everywhere. Also, the actual _value_ of the `LIQUID_DEV_MODE` env var is not taken into account, but maybe it's nice if users could toggle dev mode by passing `LIQUID_DEV_MODE=true` and `LIQUID_DEV_MODE=false`._

This PR is a first attempt in speeding up the LiquidHaskell development experience when using the new GHC plugin. The main idea takes inspiration from the stages of a compiler (think GHC) where usually we build a "stage0" compiler to be used to compile the stage1. Similarly we introduce, for each of the `liquid-*` packages, a custom `Setup.hs` module that disables building the target modules if the `LIQUID_DEV_MODE` env var is set.

### Usage and recommended workflow

This is how one would use this:

* To begin with, perform a **full** build of **all** the libraries, by doing either `cabal v2-build` or `stack build`, **without** specifying any extra environment variables from the command line. This is morally equivalent to our "stage 0".

* At this point, the content of the `liquid-*` packages is considered "trusted" and "frozen", until we don't force another full, _non dev_ build;

* In order to quickly test changes to the `liquidhaskell` library without recompiling the `liquid-*` packages, we user need to build by passing the `LIQUID_DEV_MODE` env var as part of the build command. Examples:

#### Stack

```
LIQUID_DEV_MODE=true stack build
```

#### Cabal

```
LIQUID_DEV_MODE=true cabal v2-build
```

It's also possible (but not recommended) to add `LIQUID_DEV_MODE` to `.bashrc` or similar,but this would permanently disable building the `liquid-*` packages, and this might silently mask breaking changes to the liquidhaskell library that would manifest only when compiling these other packages.

* If the user wishes to _force_ building all the libraries again, it's sufficient to issue the same builds commands _without_ the `LIQUID_DEV_MODE`.

### DIFF
--- a/docs/mkDocs/docs/develop.md
+++ b/docs/mkDocs/docs/develop.md
@@ -2,6 +2,53 @@
 
 Here are some notes that are generally useful for people *developing* LH itself.
 
+## Fast (re)compilation
+
+When working on the `liquidhaskell` library, usually all we want is to make changes and quickly recompile
+only the bare minimum, to try out new ideas. Using a fully-fledged GHC plugin doesn't help in this sense,
+because packages like `liquid-base` or `liquid-ghc-prim` all have a direct dependency on `liquidhaskell`, and
+therefore every time the latter changes, an expensive rebuild of those packages is triggered, which
+might become tedious overtime. To mitigate this, we offer a faster, "dev-style" build mode which is based
+on the assumption that most changes to the `liquidhaskell` library do not alter the validity of
+already-checked libraries, and therefore things like `liquid-base` and `liquid-ghc-prim` can be considered
+"static assets", avoiding the need for a recompilation. In other terms, we explicitly disable recompilation
+of any of the `liquid-*` ancillary library in dev mode, so that rebuilds would also influence the 
+`liquidhaskell` library.
+
+### Usage and recommended workflow
+
+This is how you can use this:
+
+* To begin with, perform a **full** build of **all** the libraries, by doing either `cabal v2-build` or `stack build`,
+  **without** specifying any extra environment variables from the command line. This is needed to ensure that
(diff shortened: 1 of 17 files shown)


kosmikus:
Tried this with `cabal` today, and it seems to work.

ranjitjhala:
Super minor point @adinapoli -- can you also add the above text to the document for *how to* generate `liquid-*` modules so we have all the notes on how to work with these in one place? thanks!!!

adinapoli:
> can you also add the above text to the document for how to generate liquid-* modules so we have all the notes on how to work with these in one place? thanks!!!

Good point, although I am wondering whether or not that document would be the right place for it. What about another `LIQUID_DEV_MODE.md` markdown document, that we can in the future reference from the main docs entry point?

kosmikus:
I think the most important aspect here is that the documentation on how to use the dev mode does not remain hidden in the description of this PR. The second most important aspect is discoverability. So creating a new file is fine from my perspective. But ideally, already place a link to it from somewhere (or even several places), so that it can be found.

ranjitjhala:
👍 both these are things that I (and Niki and other potential LH developers
- as opposed to users - need to know, hence ultimately I would like them to
be gathered as such)

adinapoli:
@ranjitjhala @kosmikus Ok, I have added a documentation section in the "new" documentation that Ranjit is working on, as part of the "Developers' guide" inside `docs/mkDocs/docs/develop.md`, I hope that's OK.

As I was at it, I have factored the content of each and every `Setup.hs` into a separate module as part of the `liquidhaskell` library, and I would argue that's much cleaner now.

PR #1709, opened 2020-07-17
add stats to unsafe result

nikivazou:
Add stats in Unsafe Result. 

This is required when the `fail` keyword is used. 
Before, the result was unsafe so "0 contraints checked" was wrongly reported. 

For example, https://github.com/ucsd-progsys/liquidhaskell/blob/20341f29535680e18c081467f93a8e8356afe57d/tests/pos/T1709.hs was reporting 0 constraints checked before


kosmikus (approved):
Looks good to me.

PR #1714, opened 2020-07-21
Always catch all TError in Plugin

### LINKED ISSUE #1711: [plugin] Undeclared type-variable error not properly reported for GHC to render as a compilation error
In this example `Bool` is typo'd to `bool`, which is an undeclared/out-of-scope type variable. LH correctly identifies this error, but doesn't report it properly when LH is used as a GHC plugin. As a result, third party tools which integrate with GHC cannot detect the error condition. [In slack](https://liquidhaskell.slack.com/archives/C54QAL9RR/p1595280783112400?thread_ts=1595280231.110000&cid=C54QAL9RR) @ranjitjhala said, "we should trap these too" and suggested @adinapoli.
```haskell
{-@ bad :: {v:bool | v} @-}
bad = False
```

Expected behavior:
* Error is reported via GHC in a way that third-party programs (eg. GHCID) can detect.

Actual behavior:
* Error isn't reported via GHC in a way that third-party programs can detect.

--- 

Versions in use:
* `liquidhaskell@26cad4f`, `liquid-fixpoint@a4e303b`

adinapoli:
This PR, with a fairly witty title, fixes #1711 by (provisionally) catching exceptions thrown from pure code at the application boundary (i.e. the `GHC.Plugin`).

This allows the plugin to always report `TError`s using the GHC logging mechanism, so that 3rd party tools like `ghcid` and `ghcide` can correctly show the diagnostics. 

The morally-correct solution is obviously to never throw exceptions from pure code but rather converting the internals to return things like `Either (TError e) a`, but this would have been very invasive to do, in the current state of things.

### DIFF
--- a/src/Language/Haskell/Liquid/GHC/Plugin.hs
+++ b/src/Language/Haskell/Liquid/GHC/Plugin.hs
@@ -59,6 +59,7 @@ import qualified TysPrim
 import           GHC.LanguageExtensions
 
 import           Control.Monad
+import           Control.Exception                        (evaluate)
 
 import           Data.Coerce
 import           Data.List                               as L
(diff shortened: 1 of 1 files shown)


ranjitjhala:
Excellent, should we merge it @adinapoli ?

adinapoli:
@ranjitjhala go for it!

Issue #1708, opened 2020-07-16
liquid-legacy and liquid plugin behave differently


nikivazou:
in this example: https://gist.github.com/nikivazou/6932d10899f30e34a430c3e9978eaf9c

```
>> liquid-legacy tests/todo/fixme.hs # says SAFE 
>> stack exec -- liquid tests/todo/fixme.hs # crashes with SMTLIB2 error 
```

If I change the below definitions (of https://gist.github.com/nikivazou/6932d10899f30e34a430c3e9978eaf9c#file-fixme-hs-L11-L12) 

```
bool1 = undefined -- not supported by the plugin, but ok for the old LH
bool2 = undefined -- if changed to True it is ok by the plugin as well
```

to 

```
bool1 = True 
bool2 = True
```

both agree to SAFE. 

@ranjitjhala @adinapoli

ranjitjhala:
I’m surprised the old LH handled it: I didn’t think we supported
“undefined” in reflected functions?

nikivazou:
Yes, though it is not directly reflected. 
1. It would be convenient to support that
2. I am still surprised that the two ways give different response.

ranjitjhala:
What is the exact error you get?

kosmikus:
I've been looking at this briefly.

The error message with the plugin is:
```
**** LIQUID: ERROR :1:1-1:1: Error
  crash: SMTLIB2 respSat = Error "line 708 column 54421: invalid qualified identifier, sort mismatch" 
```
That's not a very helpful error message from z3, but perhaps that is normal. (I would hope it actually tells which identifier, but it doesn't. I'm not sure exactly how the line/column calculation works here either.)

I've then been looking at the actual stuff that's being sent between LH and z3 (which unfortunately does not seem traceable without changing the program; is this so rarely needed?), but I'm seeing rather substantial differences.

I guess the main candidate for differences is that the base library specs are included differently between the old and the new system. Possibly the plugin is somehow not including an instruction that the old system has. I think I'll need @adinapoli 's much better knowledge about the changes he has made for the plugin to make progress on this, but he's only going to be back on LH work on Monday.

adinapoli:
[...]

PR #1716, opened 2020-07-22
T1708

ranjitjhala:
Possible fix for #1708 - which also checks `reflect`-ed signatures and so causes reflected code with `undefined` (and other non-reflected) functions to be rejected by the well-formedness check.


ranjitjhala:
Looks like the ReWrite tests fail ; will take a look soon, but nevertheless should add a regression test to make sure this particular error doesn’t creep in later!

adinapoli:
@ranjitjhala Nice, I can confirm that this allows `frog.hs` defined as such:

```haskell
module Frog where

{-@ LIQUID "--reflection" @-}

{-@ reflect frog @-}
frog :: () -> Bool
frog () = undefined
```

To correctly fail (with the plugin) with:

```
[1 of 1] Compiling Frog             ( ../frog.hs, ../frog.o )

../frog.hs:6:13: error:
    • Illegal type specification for `Frog.frog`
    Frog.frog :: lq1:() -> {VV : GHC.Types.Bool | VV == Frog.frog lq1
                                                  && VV == GHC.Err.undefined}
    Sort Error in Refinement: {VV : bool | (VV == Frog.frog lq1
                                            && VV == GHC.Err.undefined)}
    Unbound symbol GHC.Err.undefined --- perhaps you meant: GHC.Enum.C:Bounded ?
    • 
  |
6 | {-@ reflect frog @-}
  |  
```

Having said that, I can see that this PR incorporates also the commits for the docs, would it be possible to rebase it so that it contains only the code that fixes #1708 ? In general, I think we should:

- [ ] Rebase this properly;
- [X] Add `frog.hs` (and maybe `fixme.hs` that Niki posted as part of #1708?) to the list negative tests, and "pattern match" on
  their returned error to be sure they are failing _for the right reason_;
- [x] Understand why the `ReWrite` tests are failing and fix them.

**Let me take care of adding the tests as part of this PR**, to maximise the timezone difference.

On the topic on **why the ReWrite tests are failing**: it seems to complain about `GHC.Num.fromInteger`, that comes from a typeclass (i.e. it's a typeclass method). Perhaps we have to take the `clsSigs` somehow into account when checking reflected functions?

ranjitjhala:
@nikivazou this may break code that you have that uses `undefined` and such in reflected functions ... we'll have to think of some "principled" way to do it, instead of relying on a bug :-)

PR #1718, opened 2020-07-23
Improvements to Rewriting

zgrannan:
Improvements to rewriting:

- Rewriting is now more reliable (it now will apply rewrites to subexpressions)
- A termination checker for rewriting has been added (for details see: https://github.com/zgrannan/rest/blob/master/REST-POPL21.pdf)
- Added the option `--max-rw-ordering-constraints`. Lower values can speed up verification but may prevent application of rewrite rules.
- Added the option `--rw-termination-check`. This enables the rewrite termination checker (may slow down verification)

Also see Fixpoint Repo: https://github.com/ucsd-progsys/liquid-fixpoint/pull/437


ranjitjhala:
@zgrannan -- To fix the conflicts, see the new versions of the `ReWriteX.hs` (for `X in {7,8,9}`) in `develop` there are some edits there as the old code should technically _not_ have passed the well-formedness checker (cf #1708), and I recently added code to fix that problem in #1716

nikivazou:
Thanks!!! Looks very good to me. Let us wait in case @ranjitjhala has any objection (same for the fixpoint).

PR #1722, opened 2020-07-27
Polish documentation

adinapoli:
This (WIP) PR (hopefully) improves and polishes the documentation, by acting on several fronts I am going to discuss below. Feedback and improvements welcome 🙂 

## Navbar & index

I have reworked the navbar and the index page in a "less in more" fashion, to make important content stood out more easily:

Screenshot 2020-07-27 at 12 31 29

And for the index file, this is the first thing the user will see:

Screenshot 2020-07-27 at 12 32 44

As you can see I have given priority to "How to install" rather than the educational aspect, which I have grouped into the umbrella tag "Learn". From the navbar it's also possible to quickly jump into the installation section:

Screenshot 2020-07-27 at 12 31 41

And the learning material, which also include things like Options & Specifications:

Screenshot 2020-07-27 at 12 31 51

## Sections reshuffling

I have also rebranded the "Community" and "Contributing" aspect under "Get involved", trying to break down a little bit better what the users might be doing in case they want to contribute:
 
Screenshot 2020-07-27 at 12 35 45

Generally speaking, every link should be easily accessible either via the navbar or the index page, in a click or two.

## Broken links cleanup

Last but not least I have fixed a lot of broken links (due to the fact `mkDocs` won't accept relative links for anything which is _not_ inside the `docs` directory). This ensures that we can always run `mkDocs serve -s` in _strict mode_ without errors (strict mode bails out in case any warning is reported, which is typically the case for broken links).

## What's left to  do

While this PR should feel like a step forward, there are still some little things to tweak:

- [ ] We currently (almost too casually) mention the `GHC.Plugin.Tutorial` documentation module in the `plugin.md` section,
   which might be easily missed. Furthermore, due to the fact we don't have a Hackage candidate with properly-built Haddock
   documentation, the best we can do for now is to link to the "un-rendered" [file](https://github.com/ucsd-progsys/liquidhaskell/tree/develop/src/Language/Haskell/Liquid/GHC/Plugin/Tutorial.hs) which is not the best experience. 
  Maybe it's time for us to upload the release on Hackage?

- [x] The majority of examples' (see `options.md`) demo the described options using the legacy executable, which might be
   confusing for the user. It's not clear to me what is the best approach. We could either not show any practical example (which
   I don't particularly like) or show _both_ syntax, i.e. the `ghc-options: -fplugin-opt=...` way and the executable way, which
  might get tedious very rapidly. Probably the sweet spot is to showcase _at the beginning of the section_ how you can pass
   options for both the plugin and the executable, once for all, without repeating it for every single option.

- [ ] The `specifications.md` file still doesn't feel right. In particular:
    * There are 1 or 2 links which are now broken because the relevant file they used to point to, doesn't exist anymore in
       the testsuite;
    * It contains sections which are clearly specific for the legacy executable (for example how to add specs to "external"
      modules), which I don't think we should now have, but we should rather advertise the `liquid-*` mirror package approach;
    * It contains a section on `.hquals` files, which I don't think we support anymore. When converting the hardcoded Prelude
      I have simply "unfolded" any of the `.hquals` files into the relevant `.spec` file, so perhaps we shouldn't suggest to write
      a separate `.hquals` file, but rather to add those directly into the `.spec` files.

### DIFF
--- a/docs/mkDocs/docs/options.md
+++ b/docs/mkDocs/docs/options.md
@@ -71,7 +71,6 @@ To allow reasoning about function extensionality use the `extensionality flag`. 
 {-@ LIQUID "--extensionality" @-}
 ```
 
-
 ## Fast Checking
 
 The option `--fast` or `--nopolyinfer` greatly recudes verification time, can also reduces precision of type checking. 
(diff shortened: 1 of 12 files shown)


docs/mkDocs/docs/options.md
```diff
@@ -80,57 +80,6 @@ instantiation of polymorphic type variables.
 It is suggested to use on theorem proving style when reflected 
 functions are trivially refined. 
 
-## Incremental Checking
```
ranjitjhala, line 83: Can we add this back as it may be useful even with the new plugin mode. (It focuses the check on _binders_ that have been modified which can be much quicker for large files during interactive development.)

ranjitjhala (approved):

ranjitjhala:
Hi @adinapoli this is AWESOME! Thanks!!! (Minor remark about restoring the bit about `--diff` -- I guess I should check it still works with the plugin, but still.)

adinapoli:
The AppVeyor build is a false positive, as usual 😏 : 

```
vector                           > Registering library for vector-0.12.1.2..
HttpExceptionRequest Request {
  host                 = "casa.fpcomplete.com"
  port                 = 443
  secure               = True
  requestHeaders       = []
```

adinapoli:
@ranjitjhala This includes the latest changes we discussed earlier. When CI is ready, we can merge 😸

PR #1725, opened 2020-07-28
T1697

nikivazou:
Seems that generic monad gets unsafe only when pattern-inline is on.


ranjitjhala (commented):
Shouldn’t the test have —no-pattern-in-line?

ranjitjhala:
Hmm this is rather dangerous. What exactly was the pattern-online rule for bind?

nikivazou:
My point is that the *unsoundness* comes from the pattern inline rule. Without that is it sound (if circle agrees). 
I think I have an alternative solution, but it is more drastic. Writing it now at the issue.

ranjitjhala:
thanks @nikivazou !!!

PR #1726, opened 2020-07-29
Release 0.8.10.1

adinapoli:
This PR cleans up the cabal manifest so that `cabal check` would pass without warnings and it bumps the lower bound on `liquid-fixpoint` to match the revised metadata of the tarball already uploaded to Hackage.

Depends on: https://github.com/ucsd-progsys/liquid-fixpoint/pull/438

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "liquid-fixpoint"]
 	path = liquid-fixpoint
-	url = https://github.com/ucsd-progsys/liquid-fixpoint.git
+	url = git@github.com:ucsd-progsys/liquid-fixpoint.git
 [submodule "ghc-options"]
 	path = ghc-options
 	url = https://github.com/ranjitjhala/ghc-options.git
--- a/docs/mkDocs/mkdocs.yml
+++ b/docs/mkDocs/mkdocs.yml
@@ -10,4 +10,4 @@ nav:
     - Community: contributing.md    
     - Development: develop.md         
 
-theme: litera
+theme: spacelab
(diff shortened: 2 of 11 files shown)


ranjitjhala (approved):

ranjitjhala:
conflict with LF version, will merge when CI says ok.

ranjitjhala:
There’s some oddity with appveyor unable to clone fixpoint?

adinapoli:
Thanks @ranjitjhala ! What happened to https://github.com/ucsd-progsys/liquid-fixpoint/pull/438 ? It's mentioned in my PR description as a "prerequisite" but not yet merged. Perhaps that's the reason why you got CI failures?

adinapoli:
> There’s some oddity with appveyor unable to clone fixpoint?

Uhm, yes, that's weird, but I find the "fix" a bit surprising, heh. To begin with, now our `.gitmodules` points to the SSH url of the project rather than the HTTPS one, and the former requires a set of credentials and it's likely to fail cloning unless the AppVeyor job is configured with the right keys. In principle I don't think that using the HTTPS url was intrinsically wrong, perhaps it was a temporary AppVeyor failure? 

More concerning, however, is commit https://github.com/ucsd-progsys/liquidhaskell/pull/1726/commits/d4e2942bfc1796eda438007ffc6445f576fb5665 , which seems to add a _lot_ of extra files. Were they really necessary?

ranjitjhala:
Alfredo, that particular commit mostly adds a bunch of benchmarks so I
thought it quite harmless. Is that ok?

adinapoli:
> Alfredo, that particular commit mostly adds a bunch of benchmarks so I
thought it quite harmless. Is that ok?

All good, I just wanted to check and confirm why we wanted that in the final release. 😉 I agree they do look harmless.

PR #1730, opened 2020-08-03
Update fixpoint with new PLE unfold logic

zgrannan:
Evaluation of function applications should enter a branch before expanding the body, if possible.
This reduces the number of intermediate SMT equalities generated by PLE.


ranjitjhala:
This is awesome 👏

PR #1731, opened 2020-08-03
Update fixpoint to recursively check ite

zgrannan:
Recursively enter branches via if-then-else during each PLE step.
This reduces the number of intermediate equalities added to the SMT environment


zgrannan:
No performance improvement here

PR #1734, opened 2020-08-05
Implement constant folding

zgrannan:


zgrannan:
This results in a significant improvement in the speed of this test (~60s to ~4s):

https://github.com/ucsd-progsys/liquidhaskell/blob/develop/benchmarks/popl18/ple/pos/Fibonacci.hs

Also see: https://github.com/ucsd-progsys/liquid-fixpoint/pull/442

PR #1736, opened 2020-08-06
Build on stack lts-16.8

### LINKED ISSUE #1729: Please support Diff-0.4.x
Current release versions of `liquidhaskell` cannot be built in an LTS-16.x environment because the package does not support `Diff-0.4.x`. Would that be difficult to rectify?

ranjitjhala:
Fixes #1729

### DIFF
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -204,7 +204,7 @@ library
                           KMeansHelper
 
   build-depends:      base                 >= 4.11.1.0 && < 5
-                    , Diff                 >= 0.3 && < 0.4
+                    , Diff                 >= 0.3 && < 0.5
                     , aeson
                     , binary
                     , bytestring           >= 0.10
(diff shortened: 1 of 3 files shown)


peti:
Wouldn't you have to modify the Cabal file, too, to allow the build to use Diff-0.4.x?

kosmikus:
Yes, agreed. The upper bound on `Diff` should be changed. It only works with `stack` because `stack.yaml` currently has `allow-newer` set.

ranjitjhala:
Oops should have stuck with @DanBurton's #1746

ranjitjhala:
Hi @peti -- can you let me know if this works for you? Thanks! Ranjit.

peti:
I ran `cabal build --constraint="Diff >= 0.4"` with a recent git checkout and it compiled fine. Thank you very much.

PR #1742, opened 2020-08-19
Release 0.8.10.2

### LINKED ISSUE #1739: Release 0.8.10.2 ?
GHC 8.10.2 is [out](https://www.haskell.org/ghc/blog/20200808-ghc-8.10.2-released.html) and that fixes a number of issues, including the infamous [18070](https://gitlab.haskell.org/ghc/ghc/-/issues/18070) which was preventing the plugin to work properly on Windows and with tools like ghcide.

Should we try and cut another release for LH, Fixpoint and all the liquid-* packages, so that users have some hope to get LH working with ghcide ?

Practically speaking, I think all it takes is to check that LH builds on 8.10.2 and bump the version (and the `base` constraint) for all the packages, but should be easy to do.

adinapoli:
Fixes #1739.

This PR is based on #1740 (to keep CI happy) but it also bump the various versions to `0.8.10.2`, and `liquid-base`, accordingly.

**See also:** https://github.com/ucsd-progsys/liquid-fixpoint/pull/441

Hackage releases to follow. (More specifically, we will need to release only `liquid-haskell`, `liquid-fixpoint` and `liquid-base`).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,7 +144,7 @@ jobs:
 
   cabal_810:
     docker:
-      - image: phadej/ghc:8.10.1-bionic-slim
+      - image: adinapoli/ghc:8.10.2-bionic-slim
     steps:
       - cabal_build_and_test:
           liquid_runner: "--liquid-runner=cabal -v0 v2-exec liquidhaskell -- -v0 \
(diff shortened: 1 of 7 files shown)


adinapoli:
Ah, I will need to bump the Docker image CircleCI is using.

ranjitjhala:
We want to keep this branch right?

adinapoli:
> We want to keep this branch right?

I would be in favour to keep it, yes 😉

PR #1745, opened 2020-08-24
Allow inlines expansion in reflected functions

### LINKED ISSUE #1738: Can't use imported `inline` inside a `reflected` function
LH complains 

If you create a *module* with an `inline`d function `incr`

```haskell
module InlineA where

{-@ inline incr @-}
incr :: Int -> Int
incr x = x + 1
```

and then *import* that `module` and try to use `incr` in a reflected function

```haskell
{-@ LIQUID "--reflection" @-}

module InlineB where

import InlineA

{-@ reflect bar @-}
bar :: Int -> Int
bar n = incr n
```

Then LH complains that `error:  Illegal type specification for 'InlineB.bar'` because `Unbound symbol 'InlineA.incr'`

Clearly we're not suitably serializing `inline` -- I don't think this is a regression per se, I suspect this was _always_ an issue because I didn't see a test like this in `tests/import`. (e.g. `tests/import/lib/ReflectClient0.hs` should also use an `inline` in reflected code, not just in the spec where it seems fine.)

adinapoli:
Hopefully fixes #1738 .

@ranjitjhala's intuition was correct, although the cause from the breakage was subtle, and it is the following:

When we call `qualifyExpand`, we do this on the input spec, but in case of reflected functions this is _not enough_, or rather is _not the right time_. If we take a look at the `Expand` instance which gets invoked for a `BareSpec`, is the following:

```haskell
expandBareSpec :: BareRTEnv -> F.SourcePos -> Ms.BareSpec -> Ms.BareSpec
expandBareSpec rtEnv l sp = sp 
  { measures   = expand rtEnv l (measures   sp) 
  , asmSigs    = expand rtEnv l (asmSigs    sp)
  , sigs       = expand rtEnv l (sigs       sp)
  , localSigs  = expand rtEnv l (localSigs  sp)
  , reflSigs   = expand rtEnv l (reflSigs   sp)
  , ialiases   = [ (f x, f y) | (x, y) <- ialiases sp ]
  , dataDecls  = expand rtEnv l (dataDecls  sp)
  , newtyDecls = expand rtEnv l (newtyDecls sp)
  } 
  where f      = expand rtEnv l 
```

It might seem that reflected-functions expansion happens as part of `reflSigs`, but that's not correct, because in the common case, the input `BareSpec` will record any binder to reflect as part of `reflects`, which is a `[Var]`. This means that during the first expansion we cannot do _anything_, because we have a list of opaque `Var`s at hand.

This also means that we produce a `gsRefSigs` which are not correctly expanded, as you can see from the trace for the offending test:

```
Trace: [gsRefSigs] : [T1738.bar : lq1:GHC.Types.Int -> {VV : GHC.Types.Int | VV == T1738.bar lq1 && VV == T1738Lib.incr lq1}]
```

Once I realised this, it was a matter of adding _another_ call to `qualifyExpand` when we produce the reflected signature. That's the right time to do so, methinks, because this time we have a `[(Var, LocSpecType)]`, and we know how to expand a `LocSpecType`.

This causes test `T1738` to pass, but CI will tell us if I have introduced any regression.

### DIFF
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -229,6 +229,7 @@ library
                     , optics               >= 0.2
                     , optparse-applicative < 0.16.0.0
                     , optparse-simple
+                    , optparse-applicative < 0.16.0.0
                     , githash
                     , parsec               >= 3.1
                     , pretty               >= 1.1
--- a/src/Language/Haskell/Liquid/Bare.hs
+++ b/src/Language/Haskell/Liquid/Bare.hs
@@ -199,7 +199,7 @@ makeGhcSpec0 :: Config -> GhcSrc ->  LogicMap -> [(ModName, Ms.BareSpec)] -> Ghc
 makeGhcSpec0 cfg src lmap mspecs = SP 
   { _gsConfig = cfg 
   , _gsImps   = makeImports mspecs
-  , _gsSig    = addReflSigs refl sig 
+  , _gsSig    = addReflSigs env name rtEnv refl sig
   , _gsRefl   = refl 
   , _gsLaws   = laws 
   , _gsData   = sData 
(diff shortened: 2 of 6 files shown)


ranjitjhala (approved):

ranjitjhala:
Nicely done @adinapoli -- I thought it would be a little more involved than that but happy the fix was so simple!

Issue #1737, opened 2020-08-06
Verification does not terminate when reflected function used instead of measure


zgrannan:
LiquidHaskell does not terminate on the following code:

```haskell
{-@ LIQUID "--reflection" @-}
{-@ LIQUID "--ple" @-}

{-@ reflect length' @-}
{-@ length' :: [a] -> Int @-}
length' :: [a] -> Int
length' []     = 0
length' (_:xs) = 1 + length' xs

{-@ type ListX  a X = {l:[a] | length' l == length' X } @-}

{-@ reflect zw @-}
{-@ zw :: (a -> b -> c) -> xs:[a] -> ListX b xs -> ListX c xs @-}
zw :: (a -> b -> c) -> [a] -> [b] -> [c]
zw f (x:xs) (y:ys) = f x y : zw f xs ys
zw _ _      _      = []
```

However is `reflect length'` is replaced with `measure length'` (as it ought to be), verification works.

ranjitjhala:
Wow, that's a bit of a surprise. (Has it always been the case?)

I assume its `ple` that's doing the diverging?

Can you take a look since you're working on it?

zgrannan:
@ranjitjhala Sure I will take a look. I think it does have to do with PLE.

It seems like a regression: this code works on an arbitrary commit I chose from the past (d6ed10575062f861c62eb3afe238f7755d9dec99)

ranjitjhala:
Interesting! probably a good idea to git bisect to find the breaking commit?

zgrannan:
I'm no longer convinced that this is an PLE issue necessarily. This works with
`--oldple` option, but I suspect that it may be due to the logic in the old ple
to prevent recursive function applications in `if` branches.

The issue here is that every time the "length" function is applied in PLE, it
takes the recursive branch.

I've created a more simple example:

```haskell
{-@ LIQUID "--reflection" @-}
{-@ LIQUID "--ple" @-}

data Lis = LNil | LCons Int Lis

{-@ reflect length' @-}
{-@ length' :: Lis -> Int @-}
length' :: Lis -> Int
length' LNil          = 0
length' (LCons _ xxs) = 99 + length' xxs

{-@ type ListX  X = {l:Lis | length' l == length' X } @-}

{-@ reflect zw @-}
{-@ zw ::  xs:Lis -> ListX xs -> ListX xs @-}
zw :: Lis -> Lis -> Lis
zw _    thelist@(LCons _ _) = thelist
[...]

PR #1748, opened 2020-08-27
Turn reflects to measures when possible

nikivazou:


facundominguez:
Closing since I cannot reproduce #1737.

PR #1755, opened 2020-09-16
Restore --diff functionality

### LINKED ISSUE #1754: Restore --diff option when using the Plugin
Currently the plugin doesn't produce the `.bak` file on disk anymore, making the `--diff` option useless. This is because in the plugin we call `checkTargetInfo` rather than `liquidOne`, which is the function that produces the `.bak` file on disk.

Since calling `liquidOne` in the plugin is not correct (this function does too much, like calling `exitWithResult`) it seems like the quickest fix is to simply tuck the function that generates the file elsewhere, either inside `checkTargetInfo` itself or deeper in the guts.

adinapoli:
Fixes #1754.

I have tested `--diff` on an existing test file:

Screenshot 2020-09-16 at 15 14 19

Now LH correctly generates the `.bak` file:

Screenshot 2020-09-16 at 15 14 40

And if we re-run the test, it seems that there _is_ some speed-up now:

Screenshot 2020-09-16 at 15 14 52

### DIFF
--- a/src/Language/Haskell/Liquid/Liquid.hs
+++ b/src/Language/Haskell/Liquid/Liquid.hs
@@ -156,7 +156,6 @@ liquidOne :: TargetInfo -> IO (Output Doc)
 --------------------------------------------------------------------------------
 liquidOne info = do
   out' <- checkTargetInfo info
-  unless (compileSpec cfg) $ DC.saveResult tgt out'
   exitWithResult cfg [tgt] out'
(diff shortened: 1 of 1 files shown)


ranjitjhala:
This is awesome thanks Alfredo! Please merge if CI gives the green light!

nikivazou:
Thank you @adinapoli !!!

Issue #1761, opened 2020-09-21
[bug] unable to use a type alias spec of an inlined haskell function in a refined data type spec


plredmond:
**Problem:**
* Define and inline a predicate over a value.
* Create an alias as shorthand to express that the predicate is satisfied in a specification.
* Use the specification in a refined data type.
```haskell
{-@ inline oneFunPred @-}
oneFunPred :: Int -> Bool
oneFunPred x = x == 1

{-@ type OneTyAlias a = {v:a | oneFunPred v} @-}

{-@ data One = One { field :: OneTyAlias Int }  @-}
data One = One Int -- ERROR (ranjit said it ought to pass)
```
**Expect:**
The refined data type has a constructor with a specification requiring the predicate to be satisfied.
**Actual:**
Unbound symbol error on `oneFunPred` for the line with the data definition.
```
lib/Bug.hs:10:12: error:
    • Illegal type specification for `Bug.One`
    Bug.One :: Bug.field##Bug.One:{VV : GHC.Types.Int | Bug.oneFunPred VV} -> {VV : Bug.One | Bug.field VV == Bug.field##Bug.One
                                                                                              && Bug.field VV == Bug.field##Bug.One
                                                                                              && (is$Bug.One VV <=> true)}
    Sort Error in Refinement: {VV : int | Bug.oneFunPred VV}
    Unbound symbol Bug.oneFunPred --- perhaps you meant: Bug.field, Bug.One ?
    •
   |
10 | data One = One Int
   |            ^^^^^^^
```

---

**Workaround:**
Inline the type alias into the refined data type specification.
```haskell
{-@ inline oneFunPred @-}
oneFunPred :: Int -> Bool
oneFunPred x = x == 1

{-@ data One = One { field :: {v:Int | oneFunPred v}}  @-}
data One = One Int

-- This example passes (as expected)
```

**Version:**
Bug present in hackage releases
```
liquidhaskell 0.8.10.2
liquid-base 4.14.1.0
liquid-prelude 0.8.10.2
```
and also present in built-from-source
```
liquidhaskell@de4f003920d23a2683987c7f89db1afcf7df3b80 (Sept 16)
```

[Link to slack conv](https://liquidhaskell.slack.com/archives/C54QAL9RR/p1600451110042700)

plredmond:
[...]

PR #1763, opened 2020-09-23
Tentatively adding gsTySigs to ghcSpecEnv

adinapoli:
Opening this as a draft so we can discuss the changes further.

@ranjitjhala This simple fix adds the `gsTySigs` to the `GhcSpecEnv`, making the code as part of #1761 pass. This works because `T1761.oneFunPred` is a signature of the current file, but when we check the sorts, we were not including this signature in the env. The `inline` was a red herring: adding or removing `inline` as part of that example doesn't seem to change things.

What is interesting, is that these changes causes exactly **one** regression, in the `MissingReflect.hs` test. Now, I find the regression and more so the test fairly strange:

```haskell
{-@ LIQUID "--reflection"                     @-}
{-@ LIQUID "--ple"                                @-}

-- | This fails with an error as `foo` is unbound sans the `reflect` annotation. 

{- reflect foo -}

{-@ foo  :: Nat ->  xs: [Nat] -> Nat -> Bool / [len xs ]@-}
foo :: Int ->  [Int] -> Int -> Bool
foo lo [] hi     = (lo<=hi)
foo lo (x:xs) hi = (lo<=hi) && foo lo xs x

{-@ empty_foo :: lo:Nat -> hi:{Nat | lo <= hi } -> {foo lo [] hi} @-}
empty_foo :: Int -> Int -> Proof
empty_foo lo hi = (foo lo [] hi) ==. (lo <= hi) ==.True  *** QED

main :: IO ()
main = pure ()
```

What is the purpose of this test, and why adding `--reflection` silently makes everything work, but this commit breaks things? 

I think that in principle I can circumvent this and add some bandaid to actually make `MissingReflect.hs` pass again, but I couldn't help but wonder if this test is actually encouraging behaviour _we want_: if we are not reflecting `foo`, like in the example, what do we expect to happen?

### DIFF
--- a/src/Language/Haskell/Liquid/Bare.hs
+++ b/src/Language/Haskell/Liquid/Bare.hs
@@ -176,12 +176,13 @@ ghcSpecEnv sp = fromListSEnv binds
   where
     emb       = gsTcEmbeds (_gsName sp)
     binds     = concat 
-                 [ [(x,        rSort t) | (x, Loc _ _ t) <- gsMeas     (_gsData sp)]
-                 , [(symbol v, rSort t) | (v, Loc _ _ t) <- gsCtors    (_gsData sp)]
-                 , [(symbol v, vSort v) | v              <- gsReflects (_gsRefl sp)]
-                 , [(x,        vSort v) | (x, v)         <- gsFreeSyms (_gsName sp), Ghc.isConLikeId v ]
-                 , [(x, RR s mempty)    | (x, s)         <- wiredSortedSyms       ]
-                 , [(x, RR s mempty)    | (x, s)         <- _gsImps sp       ]
+                 [ [(x,        rSort t) | (x, Loc _ _ t)  <- gsMeas     (_gsData sp)]
+                 , [(symbol v, rSort t) | (v, Loc _ _ t)  <- gsCtors    (_gsData sp)]
+                 , [(symbol v, vSort v) | v               <- gsReflects (_gsRefl sp)]
+                 , [(x,        vSort v) | (x, v)          <- gsFreeSyms (_gsName sp), Ghc.isConLikeId v ]
+                 , [(x, RR s mempty)    | (x, s)          <- wiredSortedSyms       ]
+                 , [(x, RR s mempty)    | (x, s)          <- _gsImps sp       ]
+                 , [(symbol x, rSort t) | (x, Loc _ _ t)  <- gsTySigs (_gsSig sp)   ]
                  ]
     vSort     = Bare.varSortedReft emb
     rSort     = rTypeSortedReft    emb
--- a/tests/pos/T1761.hs
+++ b/tests/pos/T1761.hs
@@ -1,0 +1,11 @@
+
+module T1761 where
+
+{-@ inline oneFunPred @-}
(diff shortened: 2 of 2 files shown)


ranjitjhala:
@adinapoli  -- in this test `foo` appears in the type signature for `empty_foo` -- we only want to allow that when `foo` has been explicitly "reflected", with `{-@ reflect foo @-}` otherwise, arbitrary functions should not be allowed
in type signatures. 

What exactly happens with your `gsTySigs` change? What output do you get? (I'd be pretty surprised if it's `SAFE`, does it either: 

(a) fail with some OTHER error than the one expected for this test - which is testing for specific error-messages; or
(b) return with SAFE

?

adinapoli:
> What exactly happens with your gsTySigs change? What output do you get?

Indeed it's not `SAFE` but it fails with:

```
    MissingReflect.hs:                        FAIL (0.95s)
      tests/test.hs:510:
      Did not match message: "Illegal type specification for `Main.empty_foo`"
        > 
        > **** LIQUID: UNSAFE ************************************************************
        > /root/project/tests/errors/MissingReflect.hs:17:1-64: error:  Liquid Type Mismatch
        >                                                                   .
        >                                                                   The inferred type
        >                                                                     VV : {v : () | (isAdmit Language.Haskell.Liquid.ProofCombinators.QED => false)
        >                                                                                    && (not (isAdmit Language.Haskell.Liquid.ProofCombinators.QED) => true)
        >                                                                                    && v == Language.Haskell.Liquid.ProofCombinators.*** (Language.Haskell.Liquid.ProofCombinators.==. (Language.Haskell.Liquid.ProofCombinators.==. (Main.foo lo GHC.Types.[] hi) (lo <= hi)) GHC.Types.True) Language.Haskell.Liquid.ProofCombinators.QED}
        >                                                                   .
        >                                                                   is not a subtype of the required type
        >                                                                     VV : {VV : () | Main.foo lo GHC.Types.[] hi}
        >                                                                   .
        >                                                                   in the context
        >                                                                     hi : {hi : GHC.Types.Int | hi >= 0
        >                                                                                                && lo <= hi}
        >                                                                      
        >                                                                     lo : {lo : GHC.Types.Int | lo >= 0}
```

ranjitjhala:
Alfredo, the issue in #1761 is that the function `oneFunPred`
should not even appear in the refinement; it is an "inline"
which means that its definition should be substituted in
the `Bare` phase when we do the `expand` -- and downstream
e.g. in Bare.Check where the error gets thrown,
the `oneFunPred` should no longer even be in the type sig.

That is, the real problem is that it is *not* getting substituted away...

adinapoli:
> the issue in #1761 is that the function `oneFunPred` should not even appear in the refinement;

Yes, that is also my understanding of the issue, and I cannot really understand why I have opened this draft to begin with, apologies. Perhaps I was taken in by the fact that the error didn't go away when I removed the `inline` , which got me thinking if it was possible to have the example pass if the function is not inlined to begin with.

But that's not what #1761 is all about, so yes, back to the drawing board 🙂 . Will take another deep dive next week.

adinapoli:
Closing this one as it's clearly taking a wrong turn.

PR #1774, opened 2020-10-07
Support for GHC 9

### LINKED ISSUE #1771: Explore support for GHC 9.0.1
Now that `ghc-9.0.1-alpha1` is out is a good time starting to explore how much changed (I think a lot) and how much LH will need to be amended to support the latest and greatest GHC.

adinapoli:
Fixes #1771. 

Depends on https://github.com/ucsd-progsys/liquid-fixpoint/pull/453 .

This is still marked as draft as it still does lack CI support for GHC 9 and I am fairly sure we will get some test regressions for older GHCs due to all the churning 😉 Plus, I had to add some provisional `head.hackage` constrains I had to add to make it build with `ghc-9.0.0.20200925`.

The gist of the design is that now we have a single module that must be imported when we work with the GHC API, namely `Language.Haskell.Liquid.GHC.API` which shelter users from breaking changes to the underlying `ghc` library. This was necessary because GHC 9 changed completely the module hierarchy, so we had to go "all in" with this approach.


tests/pos/ReWrite7.hs
```diff
@@ -47,7 +47,7 @@ const' x _ = x
       -> f  : (a -> m b)
       -> {v : m b | v == mbind om x f } @-}
 mbind :: OurMonad m -> m a -> (a -> m b) -> m b
-mbind = bind
+mbind (OurMonad b _) = b
```
nikivazou, line 50: Is this edit required?

nikivazou (approved):

nikivazou:
Looks great to me! Thanks @adinapoli !!!

tests/pos/ReWrite7.hs
adinapoli: Unfortunately yes. It seems like without it GHC 9 finds this code ambiguous. The edit provides enough information, and it's semantically equivalent.

nikivazou: So, GHC 9 complains, no Liquid Haskell? This makes more sense :)

PR #1789, opened 2020-11-01
Remove dependency on `text-format`

ranjitjhala:
Matching PR for https://github.com/ucsd-progsys/liquid-fixpoint/pull/456


ranjitjhala:
@kosmikus  -- can you advise re: the above and #1788 ?

PR #1798, opened 2020-11-21
Fix real exponential spec

renanroberto:
The actual spec claims that, for `x = b^n`, `b = 0` if, and only if, `x = 0`. Note that `0^0 = 1` is a counter-example to this statement.

This PR:
- Claims that `n = 0` implies `x = 1` and `(b = 0 and n /= 0)` _iff_ `x = 0`
- Restrict the exponent to non-negative integers. As you can see in the [source code](https://hackage.haskell.org/package/base-4.14.0.0/docs/src/GHC.Real.html#%5E) of `(^)`, a negative exponent raises an error

### DIFF
--- a/liquid-base/src/GHC/Real.spec
+++ b/liquid-base/src/GHC/Real.spec
@@ -2,7 +2,7 @@ module spec GHC.Real where
 
 import GHC.Types
 
-(GHC.Real.^) :: (GHC.Num.Num a, GHC.Real.Integral b) => a:a -> n:b -> {v:a | v == 0 <=> a == 0 }
+(GHC.Real.^) :: (GHC.Num.Num a, GHC.Real.Integral b) => n:a -> {m:b | m >= 0} -> {k:a | (m == 0 => k == 1) && ((n == 0 && m /= 0) <=> k == 0)}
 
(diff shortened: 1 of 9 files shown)


renanroberto:
An infrastructure failure has occurred in the CI. Should I change anything in the code or should we rerun the tests?

ranjitjhala:
Probably enough to just rerun the tests! Thanks!

renanroberto:
Just to add some information on this topic

## Unsoundness

The current spec causes unsoundness. In the following code, the type system thinks that `one` is 1, but in runtime it is 0. Therefore the `explode` function raises a runtime error even though the code is checked.

    {-@ type False = {b:Bool | b = False} @-}

    {-@ unsound :: False @-}
    unsound :: Bool
    unsound = 0^0 == 1

    {-@ one :: {n:Nat | n = 1} @-}
    one :: Int
    one = if unsound then 0 else 1

    {-@ explode :: {n:Nat | n = 1} @-}
    explode :: Int
    explode = div 1 one

ranjitjhala:
Oops, so sorry we didn't merge this! @nikivazou does this look good to you? can we merge?

renanroberto:
No problem! If any change is needed, please let me know

facundominguez (approved):
LGTM!

PR #1800, opened 2020-11-25
Improve totality error message

renanroberto:
Change the false proposition `5 < 4` on totality errors to a more descriptive false proposition.

To do that, we create a measure `totalityError` that always returns `false` and passes a descriptive error message to it. The current message is _"Pattern match(es) are non-exhaustive"_, which is the warning that GHC gives in case `-Wincomplete-patterns` is passed.

I've tried to create the measure directly in the spec with

    measure totalityError :: a -> Bool
        totalityError _ = false

But it doesn't work. I'm not sure why, but to create this measure, I've created the `totalityError` function on ` liquid-base/src/Liquid/Prelude/Totality.hs`. I'm not sure if this is the best approach. In `include/PatErr.spec` it was not necessary to create the function and I don't know why.

Another important point is that four functions are used to handle pattern-match erros: `Control.Exception.Base.patError`, `Control.Exception.Base.recSelError`, `Control.Exception.Base.nonExhaustiveGuardsError`, and `Control.Exception.Base.noMethodBindingError`. I'm not sure when this functions are used. In all my tests (that can be found [here](https://pastebin.com/cJmecjnz)) only `patError` was called. Since I don't know how this functions are used, maybe the current error message shouldn't be used in all of them.

I've tested some cases with both, plugin and executable, and it seems to work as expected.

### DIFF
--- a/include/PatErr.spec
+++ b/include/PatErr.spec
@@ -1,7 +1,13 @@
 module spec Prelude where
 
-assume Control.Exception.Base.patError :: {v:GHC.Prim.Addr# | 5 <4 } -> a
-assume Control.Exception.Base.irrefutPatError :: {v:GHC.Prim.Addr# | 5 < 4 } -> a
-assume Control.Exception.Base.recSelError :: {v:GHC.Prim.Addr# | 5 < 4 } -> a
-assume Control.Exception.Base.nonExhaustiveGuardsError :: {v:GHC.Prim.Addr# | 5 < 4 } -> a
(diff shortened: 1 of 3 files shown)


ranjitjhala:
Omg this is terrific, thanks!!!!

ranjitjhala:
Actually we should also add a test in the error-messages suite to make sure we produce the desired error message.

renanroberto:
> Actually we should also add a test in the error-messages suite to make sure we produce the desired error message.

@ranjitjhala to be honest I don't know how to do that. I only know how to check if it was a success or a failure. If you can point it out to me I can try to help. Maybe the code I linked can be useful

PR #1805, opened 2020-11-27
Provide better error messages for different totality errors

renanroberto:
In a discussion with @ranjitjhala and @nikivazou, it was clarified which error was returned in each case. Therefore, better error messages have been added for different error cases of totality errors


ranjitjhala:
Thank you so much!!!

PR #1806, opened 2020-11-27
Add "missing field in record construction" error

renanroberto:
This commit add `recConError` in the list of totality errors. With this change, records with missing fields will not be checked by LH. For example:

    data Foo = Foo {a :: Int, b :: Int}

    foo :: Foo
    foo = Foo {a = 1}

This code will give the error "_Missing field in record construction_"


ranjitjhala:
Oops looks like there’s a conflict  PR. Did you want to resolve it?

renanroberto:
Sure! No problem

renanroberto:
I think everything is ok now =)

ranjitjhala:
Thank you again!!

PR #1809, opened 2020-12-05
Improve null type

renanroberto:
Make `null` return information about the list being empty or not. If null returns true, then the list has length zero. And if null returns false, then the list is non-empty. Now the code below can be type-checked

    foo :: [a] -> [a]
    foo xs = if null xs then [] else tail xs


nikivazou:
Great! Thanks!!!

PR #1811, opened 2020-12-11
Lazy PLE

zgrannan:
"Lazy PLE" changes LH to use the following behavior:

1. First check all constraints without PLE
2. Run PLE on any unsafe constraints, then recheck them

See: https://github.com/ucsd-progsys/liquid-fixpoint/pull/462

This can result in a speedup if constraints are already safe without PLE. For example, I noticed a 25% speedup on the tests in `tests/ple/pos`

However, this can also lead to a significant slowdown in some cases (see: https://github.com/ucsd-progsys/liquidhaskell/blob/develop/benchmarks/sf/Lists.hs), when attempting verification without PLE would take a lot of time. Therefore the "--no-lazy-ple" flag can be added to restore the old behavior.

### DIFF
--- a/benchmarks/sf/Lists.hs
+++ b/benchmarks/sf/Lists.hs
@@ -1,5 +1,6 @@
-{-@ LIQUID "--reflection" @-} 
-{-@ LIQUID "--ple"        @-} 
+{-@ LIQUID "--reflection"  @-}
+{-@ LIQUID "--ple"         @-}
+{-@ LIQUID "--no-lazy-ple" @-}
 
 module Lists where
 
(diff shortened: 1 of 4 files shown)


ranjitjhala:
@zgrannan -- do you have the csv diff showing which benchmarks get better?

zgrannan:
@ranjitjhala Here are the results: https://gist.github.com/zgrannan/b5404a0d5b89bdca2eebb2769b0b8d4c

nikivazou:
Nice! There is a good improvement! 

@zgrannan do you have time to check why cabal_900 is failing?

PR #1812, opened 2020-12-17
T1812

nikivazou:


nikivazou:
This PR and #1811  fail because `th-abstraction-0.3.2.0` does not compile with ghc-9. 
I think we should merge. Any opinions @ranjitjhala @adinapoli ?

adinapoli:
> This PR and #1811 fail because th-abstraction-0.3.2.0 does not compile with ghc-9.
I think we should merge. Any opinions @ranjitjhala @adinapoli ?

@nikivazou I am confused; this PR doesn't seem to do anything interesting with `th-abstraction`, so how come all of the sudden the ghc9 build is now failing? Something seems off to me, considering that `th-abstraction` is now at version `0.4.1.0`, so it should be totally reasonable for Cabal to pick `0.3.2.0` for GHCs < 9 and the latest for GHC 9.

In any case, I am going on annual leave in a few hours, so if this is blocking you, I would suggest you folks go along and merge whatever you need to, and perhaps me or @kosmikus could look at restoring the GHC 9 build after the Xmas break.

PR #1817, opened 2021-01-29
Fail with an error on lambda reflection

nikivazou:


ranjitjhala:
@nikivazou -- can you add an `errorTest` test for this error message so we don't forget? (and then we can merge it in?) thanks!

nikivazou:
I added the test, so merging!!!

PR #1820, opened 2021-02-01
adjust synthesis tests to fix tests under nix-build

plredmond:
I have a fix for a testing issue on my [nix branch](https://github.com/plredmond/liquidhaskell/tree/nixify) where the synthesis test incorrectly defaults to using `stack` to locate `liquid`.

This PR upstreams my fix by changing the synthesis test to use `liquid` if it is on path, and otherwise fall back to calling it via `stack`.


nikivazou:
Thanks!!!!

Issue #1827, opened 2021-02-23
Liquid Haskell is not safe wrt Safe Haskell


oquechy:
[Safe Haskell](https://downloads.haskell.org/~ghc/7.8.4/docs/html/users_guide/safe-haskell.html) transitively infers safety of Haskell modules. It checks for usages of dangerous functions like `unsafePerformIO`. The safety of a module and its exported functions can be overridden with pragmas such as `{-# LANGUAGE Trustworthy #-}` which tells to count the module as safe despite usages of unsafe code.

Modules from liquid-base have different safety properties than modules from base. This makes it harder to transition from base to liquid-base for projects that care about safety. For example, `safe import Prelude` in such projects doesn't work with liquid Prelude while it works with Prelude from base. Then, the modules that import liquid Prelude without `safe import` can't be annotated with `{-# LANGUAGE Safe #-}`. 

For comparison:
 - https://hackage.haskell.org/package/base-4.14.1.0/docs/Prelude.html - Prelude from base is *Trustworthy*
 - https://hackage.haskell.org/package/liquid-base-4.14.1.0/docs/Prelude.html - Liquid Prelude is *None*

Possible fixes:
 - Remove unsafe dependencies such as GHC.Exts from LH where possible
 - Annotate liquid analogs of safe modules from base with `{-# LANGUAGE Trustworthy #-}`

kosmikus:
After discussion, we currently think that we should be able to fix most of the problems by just marking `Prelude` from `liquid-base` as `Trustworthy`. This one is getting inferred as `None` because it imports `GHC.Exts` (but doesn't re-export anything from it). Because `Prelude` from `liquid-base` is then imported implicitly by nearly every other module, everything is `None`. But with `Prelude` being marked `Trustworthy`, safety inference should be doing the right thing for most modules in the `liquid-*` package ecosystem.

PR #1828, opened 2021-02-26
Promote DataCon in matchTyCon

nikivazou:

PR #1829, opened 2021-02-27
Add trustworthiness pragma

oquechy:
Mark Prelude, Data.Bits, Data.Word, Data.Typable, Data.ByteString, and Data.ByteString.Char8 as Trustworthy to match modules from base and bytestring packages. Without the pragma it's impossible to `safe import Prelude` and other modules, see https://github.com/ucsd-progsys/liquidhaskell/issues/1827

### DIFF
--- a/liquid-base/src/Data/Bits.hs
+++ b/liquid-base/src/Data/Bits.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE Trustworthy #-}
 module Data.Bits ( module Exports ) where
 
 import "base" Data.Bits as Exports
--- a/liquid-base/src/Data/Typeable.hs
+++ b/liquid-base/src/Data/Typeable.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE Trustworthy #-}
 module Data.Typeable (module Exports) where
 
 import "base" Data.Typeable as Exports
--- a/liquid-base/src/Data/Word.hs
+++ b/liquid-base/src/Data/Word.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE Trustworthy #-}
 module Data.Word ( module Exports) where
 
 import "base" Data.Word as Exports
--- a/liquid-base/src/Prelude.hs
+++ b/liquid-base/src/Prelude.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE Trustworthy #-}
 module Prelude (module Exports) where
 
 import Data.Foldable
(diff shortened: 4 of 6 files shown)


ranjitjhala:
Thanks @oquechy !

I think the cabal-810 test is failing for some random reason; restarting and hopefully will pass and we can merge. (May do so otherwise too...)

adinapoli:
> I think the cabal-810 test is failing for some random reason; restarting and hopefully will pass and we can merge. (May do so otherwise too...)

Yes, I agree the test failure for `cabal-810` is unrelated to this PR -- it seems like #1828 also failed due to the same issue. In particular, it seems like the failing test is the golden one, which checks that the `--json` output we get from `liquidhaskell` is the same as the one we have stored in a file.

I don't think the failure should hold this PR, but I guess we should investigate _why_ the test is failing? Do we have an issue for that?

liquid-base/src/Data/Bits.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE Trustworthy #-}
```
adinapoli, line 1: I am surprised we need to add the pragma for anything except `Prelude`. For this and the other modules GHC should just transitively infer. But it looks like that wasn't the case?

oquechy: First, I tried to insert pragma only in Prelude but GHC still complained about `safe import Data.Bits` and `safe import Data.Word` in the project I was building. It didn't compile until I inserted pragmas in all liquid modules that the project uses.

Should I also file an issue to GHC about not inferring safety?

adinapoli: Uhm, maybe this is actually expected after all? I was just reading the [Safe Haskell Inference](https://downloads.haskell.org/~ghc/7.8.4/docs/html/users_guide/safe-haskell.html#safe-inference) section in GHC's documentation, and it says:

> "[..]In the case where a module is compiled without one of -XSafe, -XTrustworthy or -XUnsafe being used, GHC will try to figure out itself if the module can be considered safe. This safety inference will never mark a module as trustworthy, only as either unsafe or as safe. GHC uses a simple method to determine this for a module M: If M would compile without error under the -XSafe flag, then M is marked as safe. If M would fail to compile under the -XSafe flag, then it is marked as unsafe.[...]"

So perhaps the extra pragmas were necessary, although it's a bit unsatisfactory. I am also noticing that the original `base` modules have different inferred degree of safety: `Data.Bits` and `Prelude` for example are marked `Trustworthy`, but `Data.Word` is marked as `Safe`, so perhaps this has something to do with the chain of pragmas you had to add. Perhaps @kosmikus can shed some extra light here?

nikivazou:
The golden error seems very minor (to me). The `json` fields have the same information but different orders. I edited the field order in my PR and waiting to see what happens.

PR #1845, opened 2021-06-28
Hacking notes and minimal nix configuration

facundominguez:
This PR explains how to use a minimal `nix` configuration with `stack`. The advantage of this configuration with respect to the previous nix setup is that we reuse all of the `stack` configuration for Haskell dependencies. `nix` is only used to provision `stack` itself and the smt solver.

Also added some notes explaining how to tests changes to LH without rebuilding all of the liquid-* packages. This is a common use case when inserting print statements during debugging or when profiling.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -8,6 +8,8 @@ To run `liquid` you need to install:
 
 ## Step 1: Install SMT Solver
 
+You can skip this if you will be building LiquidHaskell with [Nix][nix].
+
 Download and install *at least one* of
 
 + [Z3](https://github.com/Z3Prover/z3/releases) or [Microsoft official binary](https://www.microsoft.com/en-us/download/details.aspx?id=52270)
@@ -36,6 +38,15 @@ GHC version is in the environment. As an example,
 If you want the most recent version, you can build from source as follows,
 either using `stack` (recommended) or `cabal`. In either case: *recursively*
 clone the repo and then build:
+
+### Build with `stack` and `Nix`
+
+This doesn't require to install `stack` or `z3` in advance. Though it will require
+installing [Nix][nix].
+
+    git clone --recursive git@github.com:ucsd-progsys/liquidhaskell.git
+    cd liquidhaskell
+    nix-shell --pure --run "stack install liquid-platform"
 
 ### Build with `stack` (recommended)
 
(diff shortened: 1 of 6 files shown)


facundominguez:
I'm happy to split the PR or make other changes if necessary.

ranjitjhala:
Oops looks like CI is not happy -- any idea? (Also I don't know much about nix maybe @kosmikus has an idea?)

facundominguez:
> Oops looks like CI is not happy -- any idea?

Most jobs fail when installing systems dependencies with `apt-get`. There is also a couple of jobs in which a golden test fails. I'm afraid I don't see how any of these failures is related to the changes contributed here.

facundominguez:
The failing golden test is presenting the json fields in a different order than expected, but their values match the golden output. Perhaps it is worth editing the test so reordered fields are accepted.

ranjitjhala:
@nikivazou i thought you had fixed that golden JSON test?

facundominguez:
hm, I just rebased the `develop` branch, in case I was missing fixes to the CI jobs.

ranjitjhala:
Yay, looks like the tests are working much better now!

@facundominguez re: the `hacking.md` did you see this document/approach?

http://ucsd-progsys.github.io/liquidhaskell/develop/

http://ucsd-progsys.github.io/liquidhaskell/develop/#fast-recompilation

these docs live in

 https://github.com/ucsd-progsys/liquidhaskell/tree/develop/docs/mkDocs/docs

Can you comment on how whether the the above is attempting to solve a similar problem as in the `hacking.md`?

facundominguez:
Ah, indeed I've missed the developer's guide. That LIQUID_DEV_MODE setting achieves the same effect. I think it would be fine to retire hacking.md.

Also, this makes me ponder whether documenting the nix setup in `INSTALL.md` is enough, or it should be explained or linked in the website too, say, in http://ucsd-progsys.github.io/liquidhaskell/#quick-start.

ranjitjhala:
Hi @facundominguez sure I think it would be great to update the PR with a link in the install and without the hacking.md? (We need to make the latter easier to spot - suggestions?)

facundominguez:
I've dropped the notes in HACKING.md. If people find the nix setup helpful I could eventually contribute the CI configuration to prevent accidental breakage.

> We need to make the latter easier to spot - suggestions?

INSTALL.md and the developer's guide seem to be quite related. If they should get more visibility, maybe INSTALL.md and the developer's guide could cross-reference each other, so the next Facundo that comes across the code base doesn't miss them :see_no_evil:

facundominguez:
Hello! Is there anything else I can do on my side to resolve this PR?

ranjitjhala:
oops, so sorry I missed this, thanks @facundominguez !!!

Issue #1850, opened 2021-07-02
Deprecated (?) Syntax for Abstract Refinements


yanhasu:
If I try and replicate the abstract refinements example from [a more recent version of the tutorial](https://goto.ucsd.edu/~nvazou/cufp15_liquidHaskell/05-abstract-refinements.html) ([and the blog](https://ucsd-progsys.github.io/liquidhaskell-blog/2013/06/03/abstracting-over-refinements.lhs/)):

```
module ARTestBad where
import Prelude hiding (max)

{-@ max :: forall <p :: Int -> Prop>. 
            Int<p> -> Int<p> -> Int<p> @-}
max     :: Int -> Int -> Int
max x y = if x <= y then y else x 
```
LH doesn't want to parse it:

```
Prelude> :l ARTestBad
[1 of 1] Compiling ARTestBad        ( src/ARTestBad.hs, interpreted )

src/ARTestBad.hs:4:36: error:
    * Cannot parse specification:
    unexpected ">"
    expecting predicatesP, bareTyArgP, mmonoPredicateP, white space, "->", "~>" or "=>"
    Proposition type with non-Bool output: lq_tmp$db##0:Int -> Prop
    * 
  |
4 | {-@ max :: forall <p :: Int -> Prop>. 
  |      
```

On the other hand, if I replace `Prop` with `Bool`, it parses OK:
```
module ARTestGood where
import Prelude hiding (max)

{-@ max :: forall <p :: Int -> Bool>. 
            Int<p> -> Int<p> -> Int<p> @-}
max     :: Int -> Int -> Int
max x y = if x <= y then y else x 
```
```
Prelude> :l ARTestGood
[1 of 1] Compiling ARTestGood       ( src/ARTestGood.hs, interpreted )

**** LIQUID: SAFE (2 constraints checked) **************************************
Ok, one module loaded.

```
And passes basic sanity tests:
```
{-@ type Odd = Int<{\x -> x mod 2 == 1}> @-}

{-@ isOdd :: Odd @-}
isOdd = max 3 9   -- passes as expected

{-@ notOdd :: Odd @-}
notOdd = max 8 4  -- fails as expected

```

This might be due to (e.g.) type aliases only defined in the tutorial, but it's hard to dig further since I can't find the source code for the [more recent tutorial](https://goto.ucsd.edu/~nvazou/cufp15_liquidHaskell/05-abstract-refinements.html) on GitHub yet.
[...]

PR #1851, opened 2021-07-05
Add self-updating syntax cheatsheet

yanhasu:
A last resort for when the main documentation (e.g. tutorials, papers) is out of date.

Suggested in #1850 

Some notes on curation:
* The list should keep itself up to date as syntax changes, since it *points to* continuous integration tests (rather than *copying* examples)
* Where possible, the examples containing unrelated mysterious syntax (e.g. the `TT` type) or deprecated/unsound features (e.g. `--no-termination`, `inline`) are avoided.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@
 
 * [LH Documentation](http://ucsd-progsys.github.io/liquidhaskell/) 
 * [Edit here](/docs/mkDocs)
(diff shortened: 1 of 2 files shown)


ranjitjhala:
This is fantastic! Thank you very much!!!

Issue #1773, opened 2020-10-05
Odd GHC panic `getFunName`


ranjitjhala:
GHC panic 

When using LH in plugin mode with either `voltron` or `disco` we get this error (if we disable `--compile-spec`) on the `Auth.hs` module

https://github.com/binah-framework/voltron/tree/package

```
ghc: panic! (the 'impossible' happened)
  (GHC version 8.10.2:
	Uh oh.
    getFunName on 	Control.Monad.Time.$fMonadTimet
  @ (Control.Monad.Trans.Reader.ReaderT
       Database.Persist.Sql.Types.Internal.SqlBackend
       (Frankie.Config.ConfigT
          Controllers.Config
          (Frankie.Controller.ControllerT Binah.Infrastructure.TIO)))
  @ (Binah.Infrastructure.TaggedT
       (Database.Persist.Class.PersistEntity.Entity Model.User))
  $dMonadTime_atSa
  lq_anf$##7205759403792912202
```

Also when checking the `Server.hs` file we get this:

```
[13 of 13] Compiling Server
ghc: panic! (the 'impossible' happened)
  (GHC version 8.10.2:
	error-strictResolveSymUnknown type constructor `Update`
    ofBDataDecl-2
CallStack (from HasCallStack):
  error, called at src/Language/Fixpoint/Misc.hs:152:14 in liquid-fixpoint-0.8.10.2.1-2YB6ojANCz25qTCkqBw6u8:Language.Fixpoint.Misc
```

unless we add the bogus imports ...

```
import           Binah.SMTP             -- TODO: DUMMY RECURSIVE IMPORTS for LH
import           Binah.Updates          -- TODO: DUMMY RECURSIVE IMPORTS for LH 
```

plredmond:
I've experienced a similar bug a couple of times recently, though it's very hard to reproduce and seems sometimes to relate to cached compilation files. Here are the situations that I've found the error and how I worked around it.

---
**First example:**
```
<no source information>:1:1: error:
    • Unknown type constructor `Data.Map.Internal.Map`
    matchTyCon: Data.Map.Internal.Map
    •
```
* This error occurs (I think) in a module `A` which defines a `newtype`. Its possible that the bug is actually triggered in another module which imports `A`. It's not clear.
* I've worked-around this error by adding a noop data refinement in module `A` to the `newtype` in question. I don't know why this resolves the issue.
    * `{-# LANGUAGE QuasiQuotes #-}`
[...]

PR #1857, opened 2021-07-20
Refactor name resolution to use `Either` / `Diagnostics` 

ranjitjhala:
A partial step towards fixing #1773 

We want to at least error-out gracefully instead of a GHC panic due to a random exception.


ranjitjhala:
@nikivazou lmk if you want to go over this -- it is a super tedious PR that refactors a bunch of "pure" code to use `Either` (aka `Bare.Lookup`) to not crash in the middle of name resolution.

ranjitjhala:
Hi @yiyunliu and @jprider63 -- can you also take a quick look at this? In particular, is it going to mess  up stuff with the `class` PR?

yiyunliu:
I'll take a look at around 4pm today. I don't think it would break anything
because I didn't really modify the existing resolution code (except for the
extra allowTypeclass boolean flag). I'll check to make sure.

jprider63:
Didn't we modify name resolution in refinements so that GHC will desugar typeclasses for us? 

FYI we're meeting tomorrow morning to discuss merging the typeclass updates.

yiyunliu:
The ghc elaboration is a separate pass. It could have been used to replace
the original LH resolution but I kept the old one and used ghc to type
check the refinements with symbols already resolved by LH. This should make
the type class extension compatible with the edits to the symbol resolution
code.

Issue #1687, opened 2020-06-11
SMT2LIB crash with non-eta-expanded Constructors


michaelborkowski:
I have the following two small files which have boiled down from research code:

https://github.com/michaelborkowski/liquidmeta/blob/master/Lambda2/SMTTest.hs
https://github.com/michaelborkowski/liquidmeta/blob/master/Lambda2/SMTTest2.hs

The first file defines the data types for exprs and types and contains the types for Small Step proof trees and its transitive closure. The first file typechecks as SAFE. The second file imports the first and contains just a trivial Typing judgement type. Running LH on the second file fails with this error

```
:1:1-1:1: Error
  crash: SMTLIB2 respSat = Error "line 1025 column 4065: Sort mismatch at argument #2 for function (declare-fun apply$35$$35$0 (Int Int) Int) supplied sort is (Array SMTTest.Expr SMTTest.Expr SMTTest.StepP)"
```

But Expr and StepP appear nowhere in the file except for this line at the top to import them into scope for LH:

```
semantics = (Step, EvalsTo)
```

If I remove this line and attempt to run `liquid SMTTest2hs` I get several errors that refer to the imported SMTTest.hs (which already checked as SAFE) along the lines of:

```
 SMTTest.hs:80:5-49: Error: Illegal type specification for `SMTTest.EApp1`

 80 |     EApp1 :: Expr -> Expr -> Step -> Expr -> Step
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

     SMTTest.EApp1 :: e:SMTTest.Expr -> e':SMTTest.Expr -> lqdc##$select##SMTTest.EApp1:{VV : SMTTest.Step | propOf VV == Step e e'} -> e1:SMTTest.Expr -> {VV : SMTTest.Step | EApp1 VV == e1
                                                                                                                                                                                && EApp1##lqdc##$select VV == lqdc##$select##SMTTest.EApp1
                                                                                                                                                                                && EApp1 VV == e'
                                                                                                                                                                                && EApp1 VV == e
                                                                                                                                                                                && propOf VV == Step (App e e1) (App e' e1)}
     Sort Error in Refinement: {VV : SMTTest.Step | propOf VV == Step e##SMTTest.EApp1 e'##SMTTest.EApp1}
     Unbound symbol Step --- perhaps you meant: len, prop ?


[...]

Issue #1682, opened 2020-05-29
crash: SMTLIB2  on ple


mystaroll:
Hi guys, PLE fails on this for some reason:
```haskell
{-@ LIQUID "--reflection"    @-}
{-@ LIQUID "--ple-local"    @-}

-- anyM is just Prelude's `any` but lifted to LH
{-@ reflect anyM @-}
anyM :: (a -> Bool) -> [a] -> Bool
anyM _ []        = False
anyM p (x:xs)    = p x || anyM p xs 

{-@ reflect asd @-}
asd :: Int -> [Int] -> Bool
asd a b = anyM (\x -> a == x) b

{-@ ple pr @-}
{-@ pr :: { asd 1 [1] } @-}
pr = ()
```
LH says: 
```haskell
:1:1-1:1: Error
  crash: SMTLIB2 respSat = Error "line 4415 column 23: unknown constant x$35$$35$auns"
```

mystaroll:
Modifying `asd` to this:
```haskell
{-@ reflect asd @-}
asd :: Int -> [Int] -> Bool
asd a b = anyM (a ==) b
```
takes away the crash, however it cannot prove this triviality:
```haskell
{-@ ple pr @-}
{-@ pr :: { asd 1 [1] } @-}
pr = asd 1 [1]
  ===  anyM (1 ==) [1] --asd definition
    ***QED
```

Any help would be appreciated 🙏

ranjitjhala:
Support for lambdas is super flaky; can you hoist out to a top level def and see if that helps?

mystaroll:
It works for the example I posted above, however my real use case is something like this :disappointed: :
```haskell
import Language.Haskell.Liquid.ProofCombinators


{-@ LIQUID "--ple-local" @-}
{-@ LIQUID "--reflection" @-}

data CList a = Empty
             | CList [a] a [a]


{-@ reflect singleton @-}
singleton :: a -> CList a
singleton e = CList [] e [] 

{-@ reflect toList @-}
toList :: CList a -> [a]
toList Empty = []
toList (CList l f r) = f : (r ++ (reverse l))


{-@ reflect eqToList @-}
eqToList ::  CList Int -> CList Int -> Bool
eqToList a b = toList a == toList b

{-@ reflect =*= @-}
{-@ infix 4 =*= @-}
(=*=) :: CList Int -> CList Int -> Bool
x =*= y = (any (eqToList x) (toList (singleton y)))

[...]

Issue #1673, opened 2020-05-14
LH is unable to prove that two elements are equal when using type families


yiyunliu:
LH commit: `develop@14c6367f46139ba3370b6e4dfe4133d14595cd9d (Tue May 12 13:12:44 2020 +0200)`
Here's an example that demonstrates the issue. `F Thing` is the same as `PNat`, but a theorem for the associativity of `PNat` (`natAssoc`) can't be invoked to prove the associativity of `F Thing` (`natAssoc'`). `coercEq` reveals that under the presence of coercions, LH is unable to prove the trivial fact that `Z == Z`.
```haskell
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE TypeFamilyDependencies #-}
{-# LANGUAGE ScopedTypeVariables #-}

{-@ LIQUID "--reflection" @-}
{-@ LIQUID "--ple" @-}

module N where

data Thing = T

data PNat = Z | S PNat

type family F t = o | o -> t

type instance F Thing = PNat

{-@ asserteq :: x:a -> {y:a | x == y} -> () @-}
asserteq :: a -> a -> ()
asserteq _ _  = ()

{-@ reflect natAdd @-}
natAdd :: PNat -> PNat -> PNat
natAdd Z y = y
natAdd (S x) y = S (natAdd x y)

{-@ reflect natAdd' @-}
natAdd' :: F Thing -> F Thing -> F Thing
natAdd' x y = natAdd x y


{-@ natAssoc :: x:PNat -> y:PNat -> z:PNat -> {natAdd x (natAdd y z) == natAdd (natAdd x y) z} @-}
natAssoc :: PNat -> PNat -> PNat -> ()
natAssoc Z y z = ()
natAssoc (S x) y z = natAssoc x y z

-- unsafe, despite being equal
{-@ natAssoc' :: x: F Thing -> y:F Thing -> z:F Thing -> {natAdd' x (natAdd' y z) == natAdd' (natAdd' x y) z} @-}
natAssoc' :: F Thing -> F Thing -> F Thing -> ()
natAssoc' x y z = natAssoc x y z

-- unsafe, despite being equal
coercEq :: ()
coercEq = asserteq (Z :: PNat) (Z :: F Thing)

```
@nikivazou  @jprider63

nikivazou:
@ranjitjhala actually this is a good example of coercions loosing information. 
I promised to make such an example a while back, but never made it :)

You do not need everything, just 

```
[...]

Issue #1150, opened 2017-11-01
exact-data-cons does not work with embedded functions?


edsko:
Consider

```
{-@ LIQUID "--exact-data-cons" @-}

{-@ data Proc a b = Proc { next :: a -> Proc a b , done :: b } @-}
data Proc a b = Proc (a -> Proc a b) b

{-@ reflect run @-}
run :: [a] -> Proc a b -> b
run []     (Proc f b) = b
run (a:as) (Proc f b) = run as (f a)
```

this yields

```
 /home/edsko/wt/clients/iohk/wt-ouroboros-spec/praos/lh/T2.hs:6:13: Error: Illegal type specification for `Main.run`
  
 6 | {-@ reflect run @-}
                 ^
  
     Main.run :: lq1:[a] -> lq2:(Proc a b) -> {VV : b | VV == Main.run lq1 lq2
                                                        && VV == (if isNull lq1 then Main.done lq2 else Main.run (tail lq1) (Main.next lq2 (head lq1)))}
     Sort Error in Refinement: {VV : b_aSa | (VV == Main.run lq1 lq2
                                              && VV == (if isNull lq1 then Main.done lq2 else Main.run (tail lq1) (Main.next lq2 (head lq1))))}
     Unbound Symbol Main.next
 Perhaps you meant: Main.done, Main.Proc, Main.run
```

It is as if the `next` constructor somehow get skipped by `--exact-data-cons`? Not sure if this is a bug, a known property, or whether I'm doing something wrong (still very new to Liquid Haskell).

edsko:
Ah, I just realized that this can be fixed (worked around?) by enabling `--higherorder`.

edsko:
Hmmm but if I don't also enable `--notermination` or mark `run` as `lazy` then then I get

```
:1:1-1:1: Error
  elaborate solver elabBE 43 "lq_anf$##7205759403792797027##dTh" {lq_tmp$x##330 : (T2.Proc a_aS9 b_aSa) | [(lq_tmp$x##330 = ds_dT3);
                                          ((T2.done lq_tmp$x##330) = b##aRp);
                                          ((T2.next lq_tmp$x##330) = f##aRo);
                                          ((lqdc##$select##T2.Proc##2 lq_tmp$x##330) = b##aRp);
                                          ((lqdc##$select##T2.Proc##1 lq_tmp$x##330) = f##aRo);
                                          ((is$T2.Proc lq_tmp$x##330) <=> true);
[...]

PR #1858, opened 2021-07-24
Limit the data types natively encoded in SMT

ranjitjhala:
In short we shouldn’t have to write “—no-adt” matching https://github.com/ucsd-progsys/liquid-fixpoint/pull/476


ranjitjhala:
Fixes #1687
Fixes #1682 
Fixes #1673 (crash)
Fixes #1150

Issue #1852, opened 2021-07-05
Why separate the documentation and blog?


yanhasu:
This project has both a [documentation website](https://ucsd-progsys.github.io/liquidhaskell/) and a [blog website](https://ucsd-progsys.github.io/liquidhaskell-blog/).

Both are run out of the `ucsd-progsys.github.io` domain, and contain some shared information (e.g. links to this repo and the online repl).

However, the sites are styled differently, and the documentation links to the blog but not the other way round.

The documentation also seems a bit more complete than the blog, except for the fact that the blog contains the blog posts as well as an animated splash screen (which are not to be found in the documentation).

I suggest merging the two sites so that newcomers can more easily find their way to relevant information.

ranjitjhala:
hmm interesting idea ... I think the main challenge is that 

1. The `docs` are mostly just plain markdown links generated by `mkdocs` but
2. The `blog` is mostly .lhs files with the type annotations etc.

Not clear how to combine? But maybe the main point is to make them clearly linked so from the one it is totally obvious how to get to the other?

yanhasu:
I see three issues:
1. **Linking** is poor, so newbies miss out on half the information
2. **Styles** are different, suggesting that one of the sites is abandoned or unofficial
3. **Duplication** of common information (tutorial, links, basic info) makes it harder to keep the sites up-to-date

I suspect that all of 1 and most of 2 could be solved by giving both sites a common nav-bar (containing all the links).

ranjitjhala:
Good points, thanks! the tricky thing is that the sites are generated quite
differently.

The blog by a custom script / hakyll and the documentation using mkDocs.
Each has unique pluses.

Maybe we can somehow use the styles for the one in the other … or perhaps
move the blog stuff into mkDocs …

Issue #1849, opened 2021-07-02
Fragmented Tutorial


yanhasu:
There are (at least) three versions of the 'Programming with Refinement Types' tutorial floating around the LH documentation:

1. https://ucsd-progsys.github.io/lh-workshop/
    (linked to from README.md as a '120 minute workshop')
2. https://ucsd-progsys.github.io/liquidhaskell-tutorial/
    (linked to from README.md as a 'long ish tutorial', also linked to as the 'PDF version' of tutorial 3)
3. https://goto.ucsd.edu/~nvazou/cufp15_liquidHaskell/
    (on a different domain, seems to contain the most up-to-date information, but isn't prominently advertised)

There are also a bunch of (similar?) documents in the `docs/slides` directory.

As a beginner, it's quite hard to find up-to-date information.
It's also not obvious which file to make pull requests on when the tutorial needs updating.

ranjitjhala:
Hmm good point! Sorry about that! Do you have any suggestions for how to improve matters?

yanhasu:
I suggest:
* Finding the source code for [the most up to date tutorial](https://goto.ucsd.edu/~nvazou/cufp15_liquidHaskell/) (maybe @nikivazou knows where it is?) and putting it somewhere prominent (e.g. this repo, its own repo) so that it's easily updated.
* Updating all the links to point to this most up-to-date tutorial (the most obvious ones are in `README.md`, [the website](https://ucsd-progsys.github.io/liquidhaskell/), [the blog's navbar](https://ucsd-progsys.github.io/liquidhaskell-blog/), and [the blog's about page](https://ucsd-progsys.github.io/liquidhaskell-blog/about.html)).

yanhasu:
Here's an even more complete version: https://github.com/ucsd-progsys/intro-refinement-types

ranjitjhala:
Great idea -- I think at the very least, as we make other different
tutorials,
we can add them to this central list with a date / duration -- so we have
them
in one place!

mstarodub:
Merging these and updating the website to point to one tutorial would be incredibly helpful!

The website currently points to both 

https://ucsd-progsys.github.io/liquidhaskell-tutorial/

and 

https://ucsd-progsys.github.io/intro-refinement-types/120/ (a fourth version?)

edit: https://nikivazou.github.io/lh-course/ seems to be another (recent) version of the course

PR #1859, opened 2021-07-24
Update README with current information

yanhasu:
i.e. point to most complete tutorial out there, and recommended installation.

Ideally the [documentation](https://ucsd-progsys.github.io/liquidhaskell/) and [blog](https://ucsd-progsys.github.io/liquidhaskell-blog/) should also be updated, but that requires susbtantially more effort.

This is related to #1849 and #1852.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,10 @@
 
 ## Quick Links
 
-* See [INSTALL.md](INSTALL.md) for building instructions.
 * [Try LiquidHaskell in your browser](http://goto.ucsd.edu:8090/index.html)
+* [Use LiquidHaskell in your stack/cabal project](https://ucsd-progsys.github.io/liquidhaskell/plugin/)
 * [Splash page with examples and link to blog](https://ucsd-progsys.github.io/liquidhaskell-blog/)
-* [120 minute workshop with more examples](http://ucsd-progsys.github.io/lh-workshop/01-index.html)
-* [Long ish Tutorial](http://ucsd-progsys.github.io/liquidhaskell-tutorial/)
+* [120 minute tutorial](https://ucsd-progsys.github.io/intro-refinement-types/120/)
 
 ## Documentation
 


facundominguez:
Hello! I just tried navigating the tutorial in https://ucsd-progsys.github.io/intro-refinement-types/120/ and the the table of contents only seems to work if clicking on the links leading to section titles. For instance, when going to https://ucsd-progsys.github.io/intro-refinement-types/120/01-index.html#/whats-this, neither Chrome 19.0 nor Firefox 76.0 can move through the slides.

README.md
```diff
@@ -7,11 +7,10 @@
 
 ## Quick Links
 
-* See [INSTALL.md](INSTALL.md) for building instructions.
 * [Try LiquidHaskell in your browser](http://goto.ucsd.edu:8090/index.html)
+* [Use LiquidHaskell in your stack/cabal project](https://ucsd-progsys.github.io/liquidhaskell/plugin/)
```
facundominguez, line 11: This documentation doesn't explain about adding `liquid-base`, `liquid-containers`, etc, to package dependencies. And it only tells you to add `liquidhaskell` to dependencies if you care to read the section titled "Benefits"!

I don't oppose making these changes, but I think I'd like the README to contain the up-to-date information one day, without sending the user through links.

yanhasu:
Good spot! I've made a new issue in the relevant repo so that the bad links can be fixed separately.

README.md
yanhasu: How do you suggest we avoid staleness without using links?

This project already has 2 websites and 4+ tutorials, with a lot of duplicated and out of date information.
Adding non-link information to the README seems like it will aggravate this situation in the long run.

yanhasu: Perhaps we should move in the other direction and cut out even more of the README links to make it really clear to newbies that they should go to [the docs website](https://ucsd-progsys.github.io/liquidhaskell/).

ranjitjhala: Yes I agree with this - otherwise the README will again get too big.

facundominguez: > Perhaps we should move in the other direction and cut out even more of the README links to make it really clear to newbies that they should go to the docs website.

This could work + improving the documentation in the website :+1:

ranjitjhala:
overriding the random CI failure...

PR #1866, opened 2021-08-07
Fix stack/configs for CI

ranjitjhala:


facundominguez:
This gets me further in fixing `ghc-9.0.1` builds.
```
diff --git a/cabal.ghc9.project b/cabal.ghc9.project
index 405c16e18..8071d2e03 100644
--- a/cabal.ghc9.project
+++ b/cabal.ghc9.project
@@ -35,7 +35,7 @@ constraints:
   FPretty ==1.1,
   HTTP ==4000.3.15,
   ListLike ==4.7.2,
-  QuickCheck ==2.13.2 || ==2.14.1,
+  QuickCheck ==2.14.2,
   active ==0.2.0.14,
   aivika ==5.9,
   aivika-transformers ==5.9,
@@ -120,8 +120,8 @@ constraints:
   text-show ==3.8.5,
   th-abstraction ==0.4.2.0,
   th-desugar ==1.10 || ==1.11,
-  th-expand-syns ==0.4.6.0,
-  th-lift-instances ==0.1.17,
+  th-expand-syns ==0.4.8.0,
+  th-lift-instances ==0.1.18,
   th-utilities ==0.2.4.3,
   tpdb ==2.2.0,
   trivial-constraint ==0.6.0.0,
```
However, `store-0.7.11` fails to build because it needs an instance of `Store Pico` which I think is generated via Template Haskell, and for some reason, the instance is not generated for ghc 9.

This might fix the crash in stack_884:
```
diff --git a/.circleci/config.yml b/.circleci/config.yml
index 86e72054b..8fdd48d47 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,7 +101,7 @@ commands:
           name: Dependencies
           command: |
             stack --stack-yaml << parameters.stack_yaml_file >> setup
-            stack --stack-yaml << parameters.stack_yaml_file >> build --only-dependencies --test --no-run-tests
+            stack --stack-yaml << parameters.stack_yaml_file >> build -j2 --only-dependencies --test --no-run-tests
       - save_cache:
           key: stack-{{ checksum "<< parameters.stack_yaml_file >>" }}
           paths:
```

ranjitjhala:
Thanks @facundominguez -- let me try your edits!

facundominguez:
The QuickCheck update is missing:
```
-  QuickCheck ==2.13.2 || ==2.14.1,
+  QuickCheck ==2.14.2,
```
Then you should be able to see many build failures on `store` of the form:
```
src/Data/Store/Internal.hs:199:12: error:
    • Could not deduce (Store Int) arising from a use of ‘peek’
      from the context: (IsSequence t, Store (Element t), Index t ~ Int)
        bound by the type signature for:
                   peekSequence :: forall t.
                                   (IsSequence t, Store (Element t), Index t ~ Int) =>
                                   Peek t
        at src/Data/Store/Internal.hs:197:1-74
    • In a stmt of a 'do' block: len <- peek
      In the expression:
        do len <- peek
           replicateM len peek
      In an equation for ‘peekSequence’:
          peekSequence
            = do len <- peek
                 replicateM len peek
    |
199 |     len <- peek
    |            ^^^^
```
There is one such failure for many basic types.

ranjitjhala:
Thanks @facundominguez ! At least the 884 thing worked… is store not updated to work for ghc9?

facundominguez:
> is store not updated to work for ghc9?

I don't think it is. Its CI tests up to ghc-8.8.

ranjitjhala:
Hmm wonder what the options are.

1. Disable `ghc9` from the CI _until_ the `store` team updates
2. Disable `ghc9` from the CI _until_ I can make a patch to update `store`
3. Disable `store` (and revert to `binary`)
4. Ping the `store` folks to ask what their plans are (last commit on April 2021)

Any suggestions @facundominguez  or @adinapoli ?

facundominguez:
Asking the `store` maintainers would inform a decision.

Instead of (3), maybe consider using some of the other alternatives for serialization.

ranjitjhala:
Have asked a question here https://github.com/mgsloan/store/issues/160

facundominguez:
Hello! What's the plan after the exchange in https://github.com/mgsloan/store/issues/160?

ranjitjhala:
Hmm I thought it may be good to try to get a sense of how easy it is to
write a patch?

facundominguez:
When I tried it, it required some non-obvious changes involving Template Haskell and I didn't get to the bottom of it.

The main question I have is whether I should pester you to merge other work before this is resolved or I should wait :)

ranjitjhala:
Let's merge the other work (the ANF-flag, right?) - and kick the can of the
broken ghc9 build down the road :-)

ranjitjhala:
Hmm... if the TH changes were non obvious to you, they are certainly going
to be beyond me ...

adinapoli:
Sorry folks, I was on holiday and I have missed all of this 🙈 

Out of those 4 options, looks like `4` was eventually attempted and I agree that `1` and `2` are not ideal as GHC 9.2 RC1 was _literally_ out last night, so we are slowly moving towards into a more broader-support GHC 9 era 😉 

On `3. Disable store (and revert to binary)`, what @facundominguez says (i.e. drop `store`, use a different serialisation library) makes sense to me. There are lots of different serialisation libraries out there which performance should be adequate enough for LiquidHaskell, and they might already support GHC 9 out of the box.

I haven't looked at the internal implementation of LiquidHaskell's support for serialisation, but it looks like it might be a good time now to abstract over the _concrete_ serialisation library being used, so that in the future changes like these should be less traumatic. But perhaps supporting one or the other library is easy enough, so maybe it doesn't matter 😉

PR #1867, opened 2021-08-10
Update liquid-fixpoint to get environment reduction

facundominguez:
`liquid-fixpoint` now includes the changes of ucsd-progsys/liquid-fixpoint#481 and ucsd-progsys/liquid-fixpoint#473.

This PR includes the CI fixes from #1866.


facundominguez:
I'm working on making the tests pass in CI jobs other than cabal_900.

facundominguez:
Environment reduction is causing a failure in `Tests.Benchmarks.icfp_pos.IfM2.hs`, but only when built with `cabal-install`. I'm going to debug that.

facundominguez:
@ranjitjhala, this is ready for review. I don't think I can fix the cabal_900 job without messing with `store`.

There was a bug in environment reduction whose fix I'm delivering in https://github.com/ucsd-progsys/liquid-fixpoint/pull/481 together with the flag for inlining.

facundominguez:
@ranjitjhala, cabal_900 is green, but there is a failed benchmark in there.

Do you know why the returned code of the `cabal` invocation is ignored in the CI configuration?
```
(... cabal v2-test ...) || true
```

`stack_810_legacy_executable` failed for a transient failure and it was green before. The only change since then was e004836 which doesn't affect that job.

facundominguez:
Apparently, environment-reduction isn't breaking that one :disappointed_relieved:

ranjitjhala:
Hmm that’s odd - I think maybe @adinapoli might recall…

facundominguez:
I just checked that the benchmark fails with cfc334a9, which is a commit old enough that it doesn't include store or any of my modifications related to environment reduction.

Still, is this good to merge as is?

ranjitjhala:
Wow so this has been failing “silently” since then?

Yes let’s merge!

ranjitjhala:
Btw, @facundominguez -- this PR subsumes #1866, yes?

facundominguez:
> Btw, @facundominguez -- this PR subsumes #1866, yes?

That's correct.

adinapoli:
> Do you know why the returned code of the cabal invocation is ignored in the CI configuration?

Sorry folks, I was on holiday and I have missed this one until now 😉 To give you some historical context, @ranjitjhala might recall that we decided to provisionally ignore the exit code of the test run for `cabal_900` because we did have 3 tests failing (one of which exploded with an out-of-stack error) but we still wanted to ensure LH would build with GHC9, so this was our interim measure. Great to see things are moving in this area!

> stack_810_legacy_executable failed for a transient failure and it was green before. The only change since then was e004836 which doesn't affect that job.

Weird, probably some caching issue in the Docker container? Either way this PR is now merged, which means that whichever it was, it was indeed transient 😉

PR #1868, opened 2021-08-15
Add GHC.Num.abs spec

renanroberto:
`abs` now returns a non-negative number

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ You can directly extend and run the tests by modifying the files in
 
     tests/harness/
 
+
 ### Parallelism in Tests
 
 Most tests run in parallel, with a few module dependencies built sequentially in
--- a/liquid-base/src/GHC/Num.spec
+++ b/liquid-base/src/GHC/Num.spec
@@ -9,7 +9,7 @@ GHC.Num.negate :: (GHC.Num.Num a)
                => x:a
                -> {v:a | v = -x}
 
-GHC.Num.abs :: (GHC.Num.Num a) => a -> {x:a | x >= 0}
+GHC.Num.abs :: (GHC.Num.Num a) => x:a -> {y:a | (x >= 0) ==> (y = x) && (x < 0) ==> y = -x}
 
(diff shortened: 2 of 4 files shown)


nikivazou:
Thanks!!! Any idea why your tests fails on ghc-9?

renanroberto:
@nikivazou to be honest I have no idea =( I was hoping to fix this PR and then use the lessons learned to fix #1869. But without success until the moment. If you have any direction you can give me I would love to jump back into this PR!

andrewthad:
The CircleCI logs say:

```
          > tests/pos/abs.hs:5:1: error:
          >     Liquid Type Mismatch
          >     .
          >     The inferred type
          >       VV : GHC.Types.Int
          >     .
          >     is not a subtype of the required type
          >       VV : {VV : GHC.Types.Int | VV >= 0}
          >     .
          >     Constraint id 1
          >   |
          > 5 | f x = abs x
          >   | ^^^^^^^^^^^
```

That doesn't really make sense though. Also, I think that instead of

     GHC.Num.abs :: (GHC.Num.Num a) => a -> {x:a | x >= 0}

It would be better to have

     GHC.Num.abs :: (GHC.Num.Num a) => x:a -> {y:a | (x >= 0) ==> (y = x) && (x < 0) ==> y = -x}

Since the solver will deduce that `y >=0` even if nothing about x is known.

facundominguez:
Hello! Is anyone willing to test this in the newest LH?

renanroberto:
> Also, I think that instead of
> 
> ```
>  GHC.Num.abs :: (GHC.Num.Num a) => a -> {x:a | x >= 0}
> ```
> 
> It would be better to have
> 
> ```
>  GHC.Num.abs :: (GHC.Num.Num a) => x:a -> {y:a | (x >= 0) ==> (y = x) && (x < 0) ==> y = -x}
> ```
> 
> Since the solver will deduce that `y >=0` even if nothing about x is known.

Great idea! Added this improvement

> Hello! Is anyone willing to test this in the newest LH?

Not sure how to do this. I've rebased this branch with develop. Should this be enough? Also, I'm having some problem with CircleCI, but I think it is unrelated to the PR

PR #1871, opened 2021-08-19
Show constraint ids in error messages for easier correlation with fq dumps

facundominguez:
This PR adds constraint ids to error messages about unsafe subtyping constraints. Now errors read

```

Type.hs:19:1: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : {v : () | (isAdmit Language.Haskell.Liquid.ProofCombinators.QED => false)
                     && (not (isAdmit Language.Haskell.Liquid.ProofCombinators.QED) => true)}
    .
      VV : {VV : () | ?a == Type.TFun (Type.funArgTy ?a) (Type.funResTy ?a)}
    .
    in the context
      ?a : Type.Ty
    Constraint id 5
```

Where the last line indicates the id of the constraint in `Type.hs.fq` and `Type.hs.fq.prettified`.

Constraint ids are more convenient than source code locations to identify failing constraints. Several constraints might be associated to a source code line and telling them apart is not always immediate. Also, I've seen sometimes errors reported at slightly different lines than their respective constraints, which also makes more challenging relying on source locations alone.

### DIFF
--- a/src/Language/Haskell/Liquid/Constraint/Split.hs
+++ b/src/Language/Haskell/Liquid/Constraint/Split.hs
@@ -365,7 +365,7 @@ bsplitC' γ t1 t2 tem isHO
     r2' = rTypeSortedReft' γ tem t2
     ci  = \sr -> Ci src (err sr) (cgVar γ)
     tag = getTag γ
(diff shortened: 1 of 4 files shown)


facundominguez:
Does the current formatting of the error work?

ranjitjhala:
This is great thanks! We can merge with CIs blessing …

facundominguez:
All green :tada:

PR #1874, opened 2021-08-30
treat measures as reflected too

nikivazou:
Measures are also PLE-unfolded (when `--reflection` flag is on). As a side effect of this PR, reflection of functions that case split on characters is also supported.

### DIFF
--- a/tests/pos/T1874.hs
+++ b/tests/pos/T1874.hs
@@ -1,4 +1,4 @@
-module TypeFail where
+module T1874 where
 
 {-@ LIQUID "--exact-data-cons" @-}
 {-@ LIQUID "--reflection"      @-}
--- a/tests/pos/T1874.hs
+++ b/tests/pos/T1874.hs
@@ -1,0 +1,68 @@
+module TypeFail where
+
+{-@ LIQUID "--ple" @-}
+{-@ LIQUID "--exact-data-cons" @-}
+
+import Language.Haskell.Liquid.ProofCombinators
+
+data Ty = TInt | TFun Ty Ty
+
+{-@
+type FunTy = { t : Ty | isFunTy t }
+data Ty = TInt | TFun Ty Ty
+@-}
+
+{-@ measure isFunTy @-}
+isFunTy :: Ty -> Bool
+isFunTy (TFun _ _) = True
+isFunTy _ = False
+
+{- reflect funResTy @-}
+{-@ measure funResTy @-}
+{-@ funResTy :: FunTy -> Ty @-}
+funResTy :: Ty -> Ty
+funResTy (TFun _ b) = b
+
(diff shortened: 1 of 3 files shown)


tests/pos/T1874.hs
```diff
@@ -0,0 +1,69 @@
+module TypeFail where
```
facundominguez: Maybe
```suggestion
module T1874 where
```

nikivazou:
Any idea why `cabal_810` fails to install th-orphans-0.13.11? In this branch I did not edit anything related to template Haskell.... 
I get exactly the same issue as mgsloan/th-orphans/issues/29 (which seems to be addressed long ago).

facundominguez:
I tried building with cabal locally with the same docker image used in CI, and it succeeded. Maybe try to ignore the circleci cache.
```
diff --git a/.circleci/config.yml b/.circleci/config.yml
index 0870715c7..011b12a62 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,7 +51,7 @@ commands:
       - setup_project
       - restore_cache:
           keys:
-            - cabal-{{ checksum "liquidhaskell.cabal" }}
+            - cabal-{{ checksum "liquidhaskell.cabal" }}-2
       - run:
           name: Dependencies
           command: |
```
But I'm puzzled as well.

facundominguez:
Oh, well, cabal_810 succeeded at installing dependencies in the last attempt. :thinking:

facundominguez:
I'm not confident enough with PLE yet, but there is some setup that seems to involve measures here:
https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/src/Language/Fixpoint/Solver/PLE.hs#L101

Is this better resolved in Liquid Haskell instead?

nikivazou:
> I'm not confident enough with PLE yet, but there is some setup that seems to involve measures here:
> https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/src/Language/Fixpoint/Solver/PLE.hs#L101
> 
> Is this better resolved in Liquid Haskell instead?

The edit in Liquid Haskell is basically to tell it to treat measures as if they are also defined to be reflected, so it seemed simpler to me. 
The only failing test after that, had to do with Characters (we could not reflect Chars because GHC is constructing them using a Primitive Data constructors; but I wired this constructor in Liquid Haskell as well). 

Everything is fixed now, other than circleci cannot download `fpco/stack-build:lts-16.8`...

facundominguez:
That sounds fine to me. It raises a design question that can be addressed elsewhere: should liquid-fixpoint treat measures as reflected or should the user of liquid-fixpoint be explicit on what to treat as reflected?

Either answer will likely require an issue in LF.

PR #1876, opened 2021-09-02
Shorten ci times by avoiding rebuilds and uploading caches

facundominguez:


facundominguez:
Jobs are now half-an-hour shorter. They could be shrunk another half hour by running the tests in parallel, but unfortunately that will require refactoring the tests. When concurrent invocations of ghc use files which depend on each other, say
```
ghc Imported.hs &
ghc Importer.hs
```
interference between the calls could cause an intermittent build failure. To run reliably in parallel, both calls need to work on disjoint sets of modules.

Update: Maybe it wouldn't be too difficult to tell ghc/liquidhaskell to place artifacts in per test folders.

facundominguez:
I did try parallelism in tests. Though it does speed up running the testsuite locally, it has the opposite effect in CI jobs. [Here's](https://github.com/ucsd-progsys/liquidhaskell/compare/develop...facundominguez:fd/parallel-tests) a branch with the changes.

nikivazou:
Thanks a lot! 
So, we leave the parallelism out now!

PR #1877, opened 2021-09-10
Plugin reexport

nikivazou:
Addresses re-export of refined data types bug observed by @plredmond. 
This was failing before with the plugin: https://github.com/ucsd-progsys/liquidhaskell/commit/26ce9c9bbaed8c4730fd798425354da5bd753aee


plredmond:
Thanks for this!

nikivazou:
Strengthening does not work.
Alternative solution: tell the final module to use the type of the data constructor from the specs instead of datadecl.

PR #1862, opened 2021-07-28
Revamp Docs

yanhasu:
As discussed in #1852, the `docs` site could do with an update.

TODO:
- [x] Upgrade nav & theme
- [x] Update existing content from `docs`
- [x] Port non-blogpost content from `blog` to `docs`
- [x] Tidy up

PR #1880, opened 2021-09-21
Some improvements to chart generation

facundominguez:
With this patch we don't need imagemagick anymore. Also, chart_perf.sh can generate charts when invoked from any directory. Also, the rendering of the labels is correct. ImageMagick would be confused by some characters when converting from svg to png.

Here's an [example](https://github.com/ucsd-progsys/liquid-fixpoint/pull/493#issuecomment-924046602).

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,24 @@ You can directly extend and run the tests by modifying
 
     tests/test.hs
 
+## How to create performance comparison charts
+
+Everytime `liquidhaskell` tests are run, a report of the time taken by
+each test is written to a file `tests/logs/<host>-<time>/summary.csv`.
+
+There is a script `scripts/plot-performance/chart_perf.sh` that can be
+used to generate comparison charts in `svg` and `png` formats. It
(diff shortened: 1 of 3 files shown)


ranjitjhala:
Super! Can you link this from the docs/develop.md so we don’t lose it?
Thanks so much!!!

facundominguez:
I can't find docs/develop.md after some reorganization in #1862.

Some of the contents of develop seem to have gone to the README. Should I add a section on running chart_perf after
https://github.com/ucsd-progsys/liquidhaskell#how-to-run-regression-tests
?

ranjitjhala:
Yes that’s great!

facundominguez:
Done. CI failures seem unrelated.

facundominguez:
Hey, and how about merging this? Is it worth retrying the failed jobs? The retry buttons are disabled for me in the CI interface.

ranjitjhala:
oops my bad! merged!

PR #1881, opened 2021-10-04
Update liquid-fixpoint after performance improvements to serialization

facundominguez:


facundominguez:
There are many different failures here. 
* `store` doesn't build anymore in cabal_900
* `liquidhaskell` doesn't build anymore with cabal_810
* Various stack jobs fail on a test that takes to long to produce output
* stack_884_legacy_executable fails when downloading and extracting the docker image with ghc and stack.

I guess all of these will be fixed eventually. But I wanted to ask: is it worth running all of these jobs? For instance, does someone depend on `stack_865_legacy_executable` and `stack_884_legacy_executable` passing? What about appveyor?

ranjitjhala:
Hmm, good point.

I thought Facundo updated `store`?

Also, yes, lets ditch 884 for sure!

Why is LH not building with cabal_810? That's rather unfortunate :-(

facundominguez:
> Why is LH not building with cabal_810? That's rather unfortunate :-(

aeson released a new major version which changes the type used for keys.

ranjitjhala:
Will a constraint to use the earlier version fix it?

facundominguez:
I think there is a fix that works with both versions of aeson. I'm sending fixes here: https://github.com/ucsd-progsys/liquidhaskell/pull/1884

facundominguez:
All green!

ranjitjhala:
Thank you!!!

PR #1884, opened 2021-10-14
CI fixes

facundominguez:


facundominguez:
`Lists.hs` continues to timeout every now and then. Should we skip this test?

```
Prover
    foundations
      Lists.hs:                               TIMEOUT (1200.10s)
        Timed out after 1200s
```

PR #1885, opened 2021-10-15
Shorten CI times by using VMs instead of docker images

facundominguez:
VMs offer 7.5GB of memory, while the docker executor only has 4GB. Tests and builds do seem to run in half of the time with VMs.


facundominguez:
This is ready to merge!

PR #1890, opened 2021-10-24
Disable environment reduction by default

facundominguez:
Environment reduction can break the tree-like invariant necessary by PLE, which can cause LH to decide incorrectly on the validity of constraints.

Therefore, this PR makes not running environment reduction the default.

We can revert this after https://github.com/ucsd-progsys/liquid-fixpoint/issues/496 is fixed.

### DIFF
--- a/src/Language/Haskell/Liquid/UX/CmdLine.hs
+++ b/src/Language/Haskell/Liquid/UX/CmdLine.hs
@@ -434,6 +434,11 @@ config = cmdArgsMode $ Config {
     = Nothing 
         &= help "Maximum fuel (per-function unfoldings) for PLE"
 
+  , environmentReduction
+    = def
+        &= explicit
+        &= name "environment-reduction"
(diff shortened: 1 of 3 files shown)


ranjitjhala:
Thanks for catching this ! Shall I merge?

ranjitjhala (approved):

src/Language/Haskell/Liquid/UX/CmdLine.hs
```diff
@@ -702,6 +707,7 @@ defConfig = Config
   , skipModule               = False
   , noLazyPLE                = False
   , fuel                     = Nothing
+  , environmentReduction     = True
```
facundominguez:
```suggestion
  , environmentReduction     = False
```

facundominguez:
Now this should be ready to merge. (Had a flipped boolean before)

facundominguez:
@ranjitjhala, @nikivazou, Ping! :)

PR #1893, opened 2021-11-02
Update CI to build with ghc 8.10.7

facundominguez:
Companion PR of https://github.com/ucsd-progsys/liquid-fixpoint/pull/501


facundominguez:
All green!

nikivazou:
Owowow that is great! 
So, LH is not compatible with ghc-8.10.7 and we can just make a new release?
We generally follow the same version numbers as ghc (https://github.com/ucsd-progsys/liquidhaskell/blob/develop/liquidhaskell.cabal#L3)

facundominguez:
> So, LH is not compatible with ghc-8.10.7 and we can just make a new release?
> So, LH is **now** compatible with ghc-8.10.7 and we can just make a new release?

I have no objections. :)

ranjitjhala:
😅 the “not” had me puzzled for a bit

PR #1895, opened 2021-11-04
change isExportedId to isExternalId

yiyunliu:
This is a draft at the moment. According to SPJ, isExternalName is more semantically correct than isExportedId. Need to make sure the change doesn't break any of the existing tests.

@jprider63

### DIFF
--- a/src/Language/Haskell/Liquid/Bare/Axiom.hs
+++ b/src/Language/Haskell/Liquid/Bare/Axiom.hs
@@ -54,7 +54,7 @@ findVarDefType :: [Ghc.CoreBind] -> [(Ghc.Var, LocSpecType)] -> LocSymbol
 findVarDefType cbs sigs x = case findVarDefMethod (val x) cbs of
   -- YL: probably ok even without checking typeclass flag since user cannot
   -- manually reflect internal names
-  Just (v, e) -> if Ghc.isExportedId v || isMethod (F.symbol x) || isDictionary (F.symbol x)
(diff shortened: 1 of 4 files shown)


yiyunliu:
@jprider63 looks like the all tests are passing but the AppVeyor build failed for some reason.

jprider63:
Looks like one of the dependencies failed to download. @nikivazou or @ranjitjhala, can you restart the AppVeyor CI build?

jprider63:
@yiyunliu, can you add the test with everything in the same module? That should trigger the CI as well.

yiyunliu:
> @yiyunliu, can you add the test with everything in the same module? That should trigger the CI as well.

Done. Hopefully, the download will succeed this time

jprider63:
Whoo! Looks like that worked! 

@nikivazou and @ranjitjhala, does this change make sense? After chatting with SPJ, it sounds like we might actually want to use `isExternalName` instead of `isExportedId`. IIRC, `isExportedId` returns `False` for top level definitions in the current module (in plugin mode), while `isExternalName` returns `True`. Does that match what you remember @yiyunliu? SPJ also asked us to update the GHC documentation so this is clearer for others. 

We should be able to publish the blog post once this is fixed.

ranjitjhala:
Sounds great! (Hopefully tests don’t break?)

jprider63:
Yup! All the tests pass in CI.

ranjitjhala:
hey, should we merge?

jprider63:
I'm in favor of merging.

PR #1897, opened 2021-11-05
Undo #1874

facundominguez:
This is the companion PR of https://github.com/ucsd-progsys/liquid-fixpoint/pull/504


facundominguez:
@nikivazou, please let me know if we need to undo more of #1874. If only going by performance, the first commit here seems to be sufficient.

ranjitjhala:
Awesome let’s merge !

nikivazou:
Thanks!!!! There is another point that I believe should be reverted, so created a PR for that.

PR #1898, opened 2021-11-08
revert more #1874

nikivazou:


facundominguez:
Thanks @nikivazou! Works for me. I just tested that MergeSort.hs continues to verify fast with this change.

PR #1911, opened 2021-11-24
Interpreter for fixpoint (LH part of the code)

michaelborkowski:
This one adds the --no-interpreter flag to liquid haskell, applies it to the fuel-related tests.

Not sure why it can't merge automatically -- I think something is off with the submodule pointer?


ranjitjhala:
I suspect your FP is stale make sure it’s upto date with develop?

nikivazou:
Yes, it seems you have linked it with an older version of fixpoint that fails to complile because of warnings. To link to the develop of fixpoint you can do the following: 

```
cd liquid-fixpoint/
git pull origin develop
cd .. 
git add liquid-fixpoint
git commit -am "update FP"
```

facundominguez:
I gave a try at rebasing `develop` with `git rebase origin/develop` and then force-pushing.

facundominguez:
I'm finding something odd when building `stitch-lh` with these changes. Here is the branch in `stitch-lh`: https://github.com/facundominguez/stitch-lh/compare/fd/interpreter

Building `stitch-lh` with 
```
stack --nix test
```
blocks indefinitely when building the module `Language.Stitch.LH.Check` with one cpu busy.
Sometimes hitting `Ctrl-C` and then running again the same command, makes compilation of the same module resume and succeed.

`liquid-fixpoint` seems to validate fine the produced `Check.hs.bfq` file.

I've bisected the failure to the commits introducing the interpreter. I hope to look more at it during next week, but I wanted to mention it here in case it inspires a diagnosis.

ranjitjhala:
Very odd. Does the standalone run of the bfq file run with the same flags
(Ie with the interpreter on?)

facundominguez:
> I've bisected the failure to the commits introducing the interpreter.

Actually, this appears to be true only when building with `ghc-8.10.2`. When building with `ghc-8.10.7` I can reproduce the failure in earlier commits. So I wouldn't hold this PR because of it.

facundominguez:
> Does the standalone run of the bfq file run with the same flags (Ie with the interpreter on?)

I think so.

facundominguez:
I created an issue https://github.com/ucsd-progsys/liquidhaskell/issues/1912 to investigate the failure, after more investigation it looks more related to REST than to the interpreter.

michaelborkowski:
What I'm seeing is the same as Facundo. The compilation of Language.Stitch.LH.Check hangs most of the time with the PLEunfold branch on ghc 8.10.2/8.10.7. On 8.10.7 it works consistently on a commit from late June (that happens to be pre-REST).

ranjitjhala:
Did you see this update From @facundominguez https://github.com/ucsd-progsys/liquidhaskell/issues/1912

ranjitjhala:
@michaelborkowski are you looking into this? Thanks!

ranjitjhala:
Ok this looks good to go, thanks @facundominguez and @michaelborkowski !!!

Issue #1914, opened 2021-11-30
Bare/Elaborate.hs: "unused top binds" warnings for `RTypeF` fields


curiousleo:
There is an error of the following form for each (?) of `RTypeF`'s fields:

```
liquidhaskell    > /.../liquidhaskell/src/Language/Haskell/Liquid/Bare/Elaborate.hs:151:7: error: [-Wunused-top-binds, -Werror=unused-top-binds]
liquidhaskell    >     Defined but not used: ‘rtf_var’
liquidhaskell    >     |
liquidhaskell    > 151 |       rtf_var    :: !tv
```

Issue #1913, opened 2021-11-30
Bare/Elaborate.hs: "dodgy import" of Data.Functor.Foldable.Fix


curiousleo:
```
liquidhaskell    > /.../liquidhaskell/src/Language/Haskell/Liquid/Bare/Elaborate.hs:41:1: error: [-Wdodgy-imports, -Werror=dodgy-imports]
liquidhaskell    >     Module ‘Data.Functor.Foldable’ does not export ‘Fix’
liquidhaskell    >    |
liquidhaskell    > 41 | import           Data.Functor.Foldable         hiding (Fix)
liquidhaskell    >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

PR #1915, opened 2021-11-30
Add "devel" flag to fail the build if there are warnings

curiousleo:
Add a `devel` flag to optionally build `liquidhaskell` such that warnings fail the build.

# Motivation

I was building https://github.com/facundominguez/stitch-lh and noticed lots of GHC warnings as it was building `liquidhaskell`. That's not a great user experience. So I wanted to see if I could fix all compiler warnings.

# Changes

This PR adds a flag called `devel` that changes the GHC flags to:

- `-Wall` to enable all warnings by default, with [these exceptions](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/using-warnings.html#ghc-flag--Wall).
- `-Wno-name-shadowing` because name shadowing is widely used in the project and this seems like a valid stylistic choice.
- `-Werror` to fail the build if there is a warning. This is useful primarily for CI.

In addition, this PR disables some warnings in specific files:

- `-Wno-orphans` wherever orphan instances are defined. This is done in lots of files. I still think it is useful to have this warning enabled by default so new code in new files is discouraged from using orphan instances.
- `-Wno-incomplete-patterns` to disable misleading warnings about inexhaustive patterns, see #1918.
- `-Wno-dodgy-imports`: see https://github.com/ucsd-progsys/liquidhaskell/issues/1913.
- `-Wno-unused-top-binds`: see https://github.com/ucsd-progsys/liquidhaskell/issues/1914.

I have made appropriate changes to the source code so there are no more GHC warnings when `liquidhaskell` is built (see #1913, #1914, #1915, #1916 for exceptions / further work on this front).

# Alternatives

The alternative to adding a new flag would have been to set the stricter GHC options by default. This is brittle: new GHC versions sometimes add new warnings to `-Wall`, which in combination with `-Werror` could lead to build failures in projects that depend on `liquidhaskell` that we as `liquidhaskell` developers are not even aware of.

# To do

- [x] Document how to set the `devel` flag when developing locally (it is enabled by default in `stack.yaml`)

### DIFF
--- a/liquid-base/src/Foreign/ForeignPtr.spec
+++ b/liquid-base/src/Foreign/ForeignPtr.spec
@@ -3,9 +3,10 @@ module spec Foreign.ForeignPtr where
 import GHC.ForeignPtr
 import Foreign.Ptr
 
-Foreign.ForeignPtr.withForeignPtr :: forall a b. fp:(GHC.ForeignPtr.ForeignPtr a) 
-  -> ((PtrN a (fplen fp)) -> GHC.Types.IO b) 
-  -> (GHC.Types.IO b)
+// `withForeignPtr` is missing from the `base` version shipped with GHC 9.0.1.
+// Foreign.ForeignPtr.withForeignPtr :: forall a b. fp:(GHC.ForeignPtr.ForeignPtr a)
+//   -> ((PtrN a (fplen fp)) -> GHC.Types.IO b)
+//   -> (GHC.Types.IO b)
 
 GHC.ForeignPtr.newForeignPtr_     :: p:(GHC.Ptr.Ptr a) -> (GHC.Types.IO (ForeignPtrN a (plen p)))
 Foreign.Concurrent.newForeignPtr  :: p:(PtrV a) -> GHC.Types.IO () -> (GHC.Types.IO (ForeignPtrN a (plen p)))
--- a/liquid-base/src/Foreign/ForeignPtr.spec
+++ b/liquid-base/src/Foreign/ForeignPtr.spec
@@ -3,8 +3,8 @@ module spec Foreign.ForeignPtr where
 import GHC.ForeignPtr
 import Foreign.Ptr
 
-GHC.ForeignPtr.withForeignPtr :: forall a b. fp:(GHC.ForeignPtr.ForeignPtr a)
-  -> ((PtrN a (fplen fp)) -> GHC.Types.IO b)
+Foreign.ForeignPtr.withForeignPtr :: forall a b. fp:(GHC.ForeignPtr.ForeignPtr a) 
+  -> ((PtrN a (fplen fp)) -> GHC.Types.IO b) 
   -> (GHC.Types.IO b)
 
 GHC.ForeignPtr.newForeignPtr_     :: p:(GHC.Ptr.Ptr a) -> (GHC.Types.IO (ForeignPtrN a (plen p)))
--- a/liquid-base/src/Foreign/ForeignPtr.spec
+++ b/liquid-base/src/Foreign/ForeignPtr.spec
@@ -3,10 +3,9 @@ module spec Foreign.ForeignPtr where
 import GHC.ForeignPtr
 import Foreign.Ptr
 
-// `withForeignPtr` is missing from the `base` version shipped with GHC 9.0.1.
-// Foreign.ForeignPtr.withForeignPtr :: forall a b. fp:(GHC.ForeignPtr.ForeignPtr a)
-//   -> ((PtrN a (fplen fp)) -> GHC.Types.IO b)
-//   -> (GHC.Types.IO b)
+GHC.ForeignPtr.withForeignPtr :: forall a b. fp:(GHC.ForeignPtr.ForeignPtr a)
+  -> ((PtrN a (fplen fp)) -> GHC.Types.IO b)
+  -> (GHC.Types.IO b)
 
 GHC.ForeignPtr.newForeignPtr_     :: p:(GHC.Ptr.Ptr a) -> (GHC.Types.IO (ForeignPtrN a (plen p)))
 Foreign.Concurrent.newForeignPtr  :: p:(PtrV a) -> GHC.Types.IO () -> (GHC.Types.IO (ForeignPtrN a (plen p)))
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -65,6 +65,11 @@ source-repository head
   type:     git
   location: https://github.com/ucsd-progsys/liquidhaskell/
 
+flag strict
+  default:     False
+  description: Enable more warnings and fail compilation when warnings occur.
+               Turn this flag on in CI.
+
 flag include
   default:     False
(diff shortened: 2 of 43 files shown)


liquidhaskell.cabal
```diff
@@ -65,6 +65,11 @@ source-repository head
   type:     git
   location: https://github.com/ucsd-progsys/liquidhaskell/
 
+flag strict
```
facundominguez: `liquid-platform` and `liquid-fixpoint` use a `devel` flag for the same purpose. Probably it would be better to chose one name to use everywhere, and probably using `devel` will require changing the fewer scripts.

facundominguez (commented):
Looks good to me.

I think we might want to change the name of either the `strict` or the `devel` flag.

`-Wno-incomplete-patterns` probably deserves an issue to get rid of it, and tweak the definitions that are triggering it.

nikivazou:
Thanks! But `cabal_900` build is failing. To be honest we had this on and (if I remember correctly) removed it because it was too difficult to maintain and support many versions. 

If `cabal_900` build gets fixed REALLY HAPPY TO MERGE :)

liquidhaskell.cabal
curiousleo: Thanks for the pointer! Renamed `strict` to `devel`.

curiousleo:
@facundominguez wrote:
> I think we might want to change the name of either the `strict` or the `devel` flag.
> 
> `-Wno-incomplete-patterns` probably deserves an issue to get rid of it, and tweak the definitions that are triggering it.

I've renamed `strict` to `devel` and opened (and referenced) #1918 to track the removal of `-Wno-incomplete-patterns`.

curiousleo:
@nikivazou wrote:
> Thanks! But `cabal_900` build is failing. To be honest we had this on and (if I remember correctly) removed it because it was too difficult to maintain and support many versions.

I managed to tweak things so a local build with `stack` against GHC 9.0.1 succeeded, so there's hope that `cabal_900` will pass -- if not I will investigate further.

> If `cabal_900` build gets fixed REALLY HAPPY TO MERGE :)

Fantastic, thanks for the encouragement! :)

liquid-base/src/Foreign/ForeignPtr.spec
```diff
@@ -3,9 +3,10 @@ module spec Foreign.ForeignPtr where
 import GHC.ForeignPtr
 import Foreign.Ptr
 
-Foreign.ForeignPtr.withForeignPtr :: forall a b. fp:(GHC.ForeignPtr.ForeignPtr a) 
-  -> ((PtrN a (fplen fp)) -> GHC.Types.IO b) 
-  -> (GHC.Types.IO b)
+// `withForeignPtr` is missing from the `base` version shipped with GHC 9.0.1.
+// Foreign.ForeignPtr.withForeignPtr :: forall a b. fp:(GHC.ForeignPtr.ForeignPtr a)
+//   -> ((PtrN a (fplen fp)) -> GHC.Types.IO b)
+//   -> (GHC.Types.IO b)
```
curiousleo: I believe that [the CI failures](https://app.circleci.com/pipelines/github/ucsd-progsys/liquidhaskell/1148/workflows/b46da860-8689-4faf-8233-a8ebcd957b01) for `stack_810` and `cabal_810` are caused by this change.

To do: investigate how this spec can be used for GHC 8.10 only.

curiousleo: Without this change, on GHC 9.0.1 but not on GHC 8.10.7, I get:

```
liquid-base      > /.../liquidhaskell/liquid-base/src/Foreign/ForeignPtr.spec:6:1: error:
liquid-base      >     • Unknown variable `Foreign.ForeignPtr.withForeignPtr`
liquid-base      >     rawTySigs
liquid-base      >     • 
liquid-base      >   |   
liquid-base      > 6 | Foreign.ForeignPtr.withForeignPtr :: forall a b. fp:(GHC.ForeignPtr.ForeignPtr a)
liquid-base      >   |
```

facundominguez: Given that `liquid-base` is building with both compilers in the `develop` branch, do you know if there is a change in this PR that started causing this error with ghc-9?

curiousleo: Hm, I reverted the change (so `ForeignPtr.spec` is now as in `develop`), and CI passes -- as expected, since I I did not (intentionally) change any behaviour in this PR.

curiousleo: This is puzzling. I created the following file, `stack-901.yaml`:

```yaml
flags:
  liquidhaskell:
    devel: true
  liquid-platform:
    devel: true
extra-package-dbs: []
ghc-options:
  hscolour: -w
allow-newer: true # 8.10.1
packages:
- liquid-fixpoint
- liquid-ghc-prim
- liquid-base
- liquid-bytestring
- liquid-prelude
- liquid-vector
- liquid-containers
- liquid-parallel
- liquid-platform
- .
extra-deps:
- hashable-1.3.0.0
- rest-rewrite-0.1.1

resolver: nightly-2021-11-28
compiler: ghc-9.0.1

nix:
  shell-file: shell-stack.nix
  path: ["nixpkgs=./nixpkgs.nix"]
```

When I build using this file, I get:

```
$ stack build --stack-yaml stack-901.yaml
[...]
liquid-base      >       
liquid-base      > **** LIQUID: SAFE (0 constraints checked) **************************************
liquid-base      > [159 of 214] Compiling Foreign.ForeignPtr
liquid-base      >       
liquid-base      > /.../liquidhaskell/liquid-base/src/Foreign/ForeignPtr.spec:6:1: error:
liquid-base      >     • Unknown variable `Foreign.ForeignPtr.withForeignPtr`
liquid-base      >     rawTySigs
liquid-base      >     • 
liquid-base      >   |   
liquid-base      > 6 | Foreign.ForeignPtr.withForeignPtr :: forall a b. fp:(GHC.ForeignPtr.ForeignPtr a) 
liquid-base      >   |
```

I now realise that there is no Stack + GHC 9 setup in CI, only Cabal + GHC 9. Perhaps the issue is with the Stack setup I created locally to test this ...

curiousleo: When I build with Stack using the YAML above (sans the `liquidhaskell:devel` flag) on `develop`, it fails too, with the same error. I don't understand what's going on there, but this points to the issue being unrelated to this PR. So I think we can go ahead and merge this PR, and deal with this Stack + GHC 9 build failure separately.

curiousleo:
All comments addressed, CI is green. Ready for review / merging.

facundominguez (approved):
LGTM
Thanks @curiousleo!

liquid-base/src/Foreign/ForeignPtr.spec
ranjitjhala: Awesome thanks for looking into this!

ranjitjhala:
Thanks!!!

PR #1920, opened 2021-12-10
Add a test with examples trimmed down from stitch-lh

facundominguez:


facundominguez:
I'm preparing a blog post containing these examples and I thought that perhaps it was good idea to include them in the test-suite.

Issue #1906, opened 2021-11-16
Record syntax rejected in GADTs


facundominguez:
The following program
```
{-# LANGUAGE GADTs #-}

{-@
data T where
  T :: { getT :: Int } -> T
 @-}
data T where
  T :: { getT :: Int } -> T

main :: IO ()
main = print ()
```
fails with
```
Test.hs:5:16: error:
    • Cannot parse specification:
    unexpected ':'
    • 
  |
5 |   T :: { getT :: Int } -> T
  |                ^

```
Should this be supported?

ranjitjhala:
I didn’t realize that was valid Haskell syntax! :-) we could support it I
guess? Seems related to the other issue (having to duplicate selector
names) so maybe can be done together?

Fizzixnerd:
I have started work on this issue.  Please see the linked PR once it's up.

ranjitjhala:
Thanks Matt!

PR #1922, opened 2022-01-10
Fix for #1906: adding GADT fields syntax

Fizzixnerd:
This is for issue #1906.  I have made the majority of the changes, but need to write more tests as things can get quite tricky.

Notably, GADTs can have shared field names only when the result type is the same, so the following is illegal, but if both were `-> T Int` it would be fine:

```haskell
data T a where
  T :: { getT :: Int } -> T Int
  S :: { getT :: Int } -> T Float
```

I want to make sure the names are resolved properly, so any ideas for additional tests are appreciated.

Depends on PR 1921.

### DIFF
--- a/src/Language/Haskell/Liquid/Bare/DataType.hs
+++ b/src/Language/Haskell/Liquid/Bare/DataType.hs
@@ -73,11 +73,15 @@ makeDataConChecker = F.testSymbol . F.symbol
 --   equivalent to `head` and `tail`.
 --------------------------------------------------------------------------------
 makeDataConSelector :: Maybe Bare.DataConMap -> Ghc.DataCon -> Int -> F.Symbol
-makeDataConSelector dmMb d i = M.lookupDefault def (F.symbol d, i) dm
-  where 
-    dm                       = Mb.fromMaybe M.empty dmMb 
+makeDataConSelector dmMb d i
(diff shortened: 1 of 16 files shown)


Fizzixnerd:
tests/basic/pos/GADTFields02.hs currently fails with:

```text
liquidhaskell    >           > **** LIQUID: UNSAFE ************************************************************
liquidhaskell    >           > 
liquidhaskell    >           > tests/basic/pos/GADTFields02.hs:21:1: error:
liquidhaskell    >           >     Liquid Type Mismatch
liquidhaskell    >           >     .
liquidhaskell    >           >     The inferred type
liquidhaskell    >           >       VV : GHC.Types.Int
liquidhaskell    >           >     .
liquidhaskell    >           >     is not a subtype of the required type
liquidhaskell    >           >       VV : {VV : GHC.Types.Int | VV >= 0}
liquidhaskell    >           >     .
liquidhaskell    >           >     Constraint id 22
liquidhaskell    >           >    |
liquidhaskell    >           > 21 | f = getT
liquidhaskell    >           >    | ^^^^^^^^
liquidhaskell    >           > 
liquidhaskell    >           
```

I also just realized I put the tests under `basic` so they would be run earlier, but I should probably move them to `datacons`.

Issue #1924, opened 2022-01-17
Tests are built and run sequentially


Fizzixnerd:
Although built with the `-threaded` option, the RTS is not provided with the `-N` option for the tests, which means they are run sequentially instead of in parallel.

Some dependency declarations are likely necessary to make this work.  For example, `xLib.hs` files are libraries for `x.hs` files, and need to be run before them.

Fizzixnerd:
I have started work on this issue.  See incoming PR

PR #1872, opened 2021-08-19
Skip failing benchmark in cabal_900

facundominguez:
This PR has CI fail if any test fails in the cabal_900 job. The CI configuration is changed, so the tests that currently fail are skipped.

This should prevent subsequent PRs from breaking more tests than those currently failing.

Issue #1491, opened 2019-05-14
Express Foreign.Marshal.Alloc.alloca property


googleson78:
`alloca` has the following signature:
```alloca :: Storable a => (Ptr a -> IO b) -> IO b```

It is expected for the `p : Ptr a` that you get in the callback has `plen p == sizeOf (undefined :: a)`.
I am having difficulties expressing this. I don't know if it's okay to ask here, as I think I am just missing something, so feel free to close this if it's misplaced.

I tried to add 
```
measure Foreign.Storable.sizeOf :: (Foreign.Storable.Storable a) => Foreign.Ptr.Ptr a -> {n : GHC.Types.Int | 0 <= n}

using (Foreign.Ptr.Ptr a) as {p : Foreign.Ptr.Ptr a | plen p == Foreign.Storable.sizeOf (deref p)}
````
to `Foreign.Storable.spec`
and 
```
import Foreign.Storable
import Foreign.Ptr

Foreign.Marshal.Alloca.alloca :: (Storable a)
                              => (({p : Ptr a | plen p == Foreign.Storable.sizeOf (deref p)}) -> GHC.Types.IO b) -> GHC.Types.IO b
```
to `Foreign.Marshal.Alloc.spec`,
but `liquid` doesn't seem to care at all.
What am I doing wrong here?

I am testing with the following program:
```
import Foreign.Marshal.Alloc (alloca)
import Foreign.Storable (peek)
import Foreign.Ptr (Ptr)

main :: IO ()
main = alloca $ \pint -> print =<< peek (pint :: Ptr Int)
```
```
 /tmp/liquid/ptrs/asdf.hs:6:41-57: Error: Liquid Type Mismatch

 6 | main = alloca $ \pint -> print =<< peek (pint :: Ptr Int)
                                             ^^^^^^^^^^^^^^^^^

   Inferred type
     VV : {v : (GHC.Ptr.Ptr GHC.Types.Int) | v == pint}

   not a subtype of Required type
     VV : {VV : (GHC.Ptr.Ptr GHC.Types.Int) | 0 < plen VV}

   In Context
     pint : (GHC.Ptr.Ptr GHC.Types.Int)
```

googleson78:
In fact it seems to be silently ignoring whatever I've written in the specs, as long as it's syntactically correct (it will complain if I leave out a `:` but not if I write `GHC.Types.BInt` instead of `GHC.Types.Int`) (??)

[...]

PR #1925, opened 2022-01-17
Parallelize tests

Fizzixnerd:
Requires adding dependency annotations still.

Performance impact is relatively significant: tests run ~2x faster on my machine with 8 logical cores.  Where a single test took ~2.5 seconds to run, now 8 are run in parallel at roughly 6-8s per test.

### DIFF
--- a/tests/test.hs
+++ b/tests/test.hs
@@ -310,7 +310,9 @@ microTests = group "Micro"
   -- RJ: disabling because broken by adt PR #1068
   -- , tesGoupsWithLibs "gradual/pos"    <$> dirTests "tests/gradual/pos"                    []                ExitSuccess
   -- , tesGoupsWithLibs "gradual/neg"    <$> dirTests "tests/gradual/neg"                    []                (ExitFailure 1)
-  , mkMicroPos "typeclass-pos"  "tests/typeclasses/pos"
+#if !MIN_VERSION_GLASGOW_HASKELL(9,0,0,0)
+  , mkMicroPos "typeclass-pos"  "tests/typeclasses/pos" -- breaks on GHC 9.0.1
+#endif
   ]
   where
     mkMicroPos name dir = testGroupsWithLibs name <$> dirTests dir [] ExitSuccess     (Just " SAFE ") (Just " UNSAFE ")
@@ -779,7 +781,9 @@ textIgnored :: [FilePath]
 textIgnored 
   = [ "Setup.lhs"
     -- , "Data/Text/Axioms.hs"
+#if MIN_VERSION_GLASGOW_HASKELL(9,0,0,0)
     , "Data/Text/Foreign.hs"                -- Breaks on GHC 9.0.1
+#endif
     , "Data/Text/Encoding/Error.hs"
     , "Data/Text/Encoding/Fusion.hs"        -- has nothing in int (compile-spec) but triggers 18 import-builds
     , "Data/Text/Encoding/Fusion/Common.hs"
--- a/tests/test.hs
+++ b/tests/test.hs
@@ -439,17 +439,23 @@ testGroupsWithLibs name (DependentTests libTests nonlibTests) =
   , after AllFinish (name <> "-libs") $ testGroup name nonlibTests
   ]
 
+
+
 -- | Creates a [TestTree] that runs without parallelism
 testSequentially :: String -> SequentialTests -> [TestTree]
 testSequentially name (SequentialTests indexedTests) =
-  let grouped = reverse $ (\(t, n) -> (testGroup (name <> show n) [t], n)) <$> indexedTests
+  let grouped = (\(t, n) -> (testGroup (mkName n) [t], n)) <$> indexedTests
       -- turns i.e. [a, b, c, d] into [(a, b), (b, c), (c, d)]
       pairs = zip <*> tail $ grouped
-      deps = (\((_, n1), (t2, _)) -> after AllFinish (name <> show n1) t2) <$> pairs
+      deps = (\((_, n1), (t2, _)) -> after AllFinish (mkName n1) t2) <$> pairs
   in
     case grouped of
       [] -> []
       (t:_) -> fst t : deps
+  where
+    -- Supports up to 1000 tests -- anymore, and you'll get a dependency cycle
+    -- issue because "X-100" matches, say"X-1000"
+    mkName n = printf "%s-%03d" name n
 
 --------------------------------------------------------------------------------
 -- | For each file in `root` check, that we get the given exit `code.`
--- a/tests/test.hs
+++ b/tests/test.hs
@@ -54,7 +54,7 @@ data DependentTests = DependentTests
   }
 
 newtype SequentialTests = SequentialTests
-  { getTests :: [(TestTree, Int)] }
+  { getTests :: [TestTree] }
 
 testRunner :: Ingredient
 testRunner = rerunningTests
@@ -443,19 +443,16 @@ testGroupsWithLibs name (DependentTests libTests nonlibTests) =
 
 -- | Creates a [TestTree] that runs without parallelism
 testSequentially :: String -> SequentialTests -> [TestTree]
-testSequentially name (SequentialTests indexedTests) =
-  let grouped = (\(t, n) -> (testGroup (mkName n) [t], n)) <$> indexedTests
-      -- turns i.e. [a, b, c, d] into [(a, b), (b, c), (c, d)]
-      pairs = zip <*> tail $ grouped
+testSequentially name (SequentialTests tests) =
+  let grouped = (\(t, n) -> (testGroup (mkName n) [t], n)) <$> (zip tests [0 :: Int ..])
+      pairs = zip grouped (tail grouped)
       deps = (\((_, n1), (t2, _)) -> after AllFinish (mkName n1) t2) <$> pairs
   in
     case grouped of
       [] -> []
       (t:_) -> fst t : deps
   where
-    -- Supports up to 1000 tests -- anymore, and you'll get a dependency cycle
-    -- issue because "X-100" matches, say"X-1000"
-    mkName n = printf "%s-%03d" name n
+    mkName n = printf "%s-sequential-%d." name n
 
 --------------------------------------------------------------------------------
 -- | For each file in `root` check, that we get the given exit `code.`
@@ -470,7 +467,7 @@ dirTests root ignored ecode yesLog noLog = odirTests root ignored Nothing ecode 
 sequentialOdirTests :: FilePath -> [FilePath] -> SequentialFileOrder -> ExitCode -> Maybe T.Text -> Maybe T.Text -> IO SequentialTests
 sequentialOdirTests root ignored fo ecode yesLog noLog = do
   DependentTests libs nonlibs <- odirTests root ignored (Just (getFileOrder fo)) ecode yesLog noLog
-  pure $ SequentialTests $ zip (libs <> nonlibs) [0..]
+  pure $ SequentialTests (libs <> nonlibs)
 
 -- | Allow parallelism for these tests, but run any tests with `Lib` in its name before the others.
 --------------------------------------------------------------------------------
--- a/tests/test.hs
+++ b/tests/test.hs
@@ -53,6 +53,9 @@ data DependentTests = DependentTests
   , toplevel :: [TestTree]
   }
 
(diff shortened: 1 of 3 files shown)


Fizzixnerd:
Forgot to mention #1924

ranjitjhala:
this is awesome, thank you!!!

Fizzixnerd:
This is ready for review, (assuming the tests pass!) but I just wanted to comment on what I changed.  I partitioned the test for each group into two sets: `Lib` tests and non-`Lib` tests.  This assumes that any test that needs to be compiled before another one is named `xLiby.hs` where `xx` and `yy` are arbitrary.  (This seems to be the convention in the tests from my view, but I wanted to make sure I wasn't missing any "dependency tests" that do not follow this naming convention)  The non-`Lib` tests are run after the `Lib` tests always, so this should ensure there is no race conditions -- at least for libs that do not depend on other libs!

This is kind of a quick and dirty solution, but it's automatic and seems to work.  As always with parallel programs however, it may be that things will mysteriously break in the future or on someone else's machine (test wise) because of races, so I thought I'd bring it up, so everyone knows what to expect

tests/test.hs
```diff
@@ -534,12 +540,17 @@ mkTest ec dir file
     test = dir </> file
 
 
+logFilePath :: FilePath -> FilePath -> FilePath
+logFilePath dir file = "tests/logs/cur" </> testFileFullPath <.> "log"
+  where
+    testFileFullPath = dir </> file
```
facundominguez: This function doesn't appear to be used anywhere.

tests/test.hs
```diff
@@ -414,29 +411,35 @@ textOrder = Just . mkOrder $
...
+testGroupsWithLibs name (libTests, nonlibTests) =
+  [ testGroup (name <> "Lib") libTests
+  , after AllFinish (name <> "Lib") $ testGroup name nonlibTests
   ]
 
 --------------------------------------------------------------------------------
 -- | For each file in `root` check, that we get the given exit `code.`
 --------------------------------------------------------------------------------
-dirTests :: FilePath -> [FilePath] -> ExitCode -> Maybe T.Text -> Maybe T.Text -> IO [TestTree]
+dirTests :: FilePath -> [FilePath] -> ExitCode -> Maybe T.Text -> Maybe T.Text -> IO ([TestTree], [TestTree])
```
facundominguez: How about making a record type to stand for the libs and the tests?
```haskell
data DependentTests = DependentTests
  { dependencies :: [TestTree]
  , toplevel :: [TestTree]
  }
```
It would help documenting the meaning of the output and other functions when it appears in the parameters.

facundominguez (commented):
Thank you, Matt!

> This assumes that any test that needs to be compiled before another one is named xLiby.hs where xx and yy are arbitrary.

How about mentioning this at the place in the code where the names are introduced?

The change looks good to me, though the times in CI do not appear to have reduced as dramatically. @Fizzixnerd, can you have a look? Perhaps some long tests simply don't parallelize well.

ranjitjhala:
The xLib idea sounds good to me!

tests/test.hs
Fizzixnerd: Removed.

tests/test.hs
Fizzixnerd: Added.

Fizzixnerd:
I noticed on my machine some tests sometimes failing in the "big library projects" (ByteString, Text, etc), so I changed anything that has a `FileOrder` to a `SequentialFileOrder` (just a newtype) and added a function that runs `SequentialFileOrder`ed tests in the order they appear in the order.

To summarize, use `testGroupsWithLibs` with `[o]dirTests` when you want to have parallel tests, and use `testSequentially` with `sequentialOdirTests` when you want to force sequential execution of tests.

Fizzixnerd:
Building on CI takes 13 minutes for stack_810, which would not change with the -N, but cabal_810 is still building dependencies as of 15 minutes in.  I think this is probably the reason for the less than-impressive speed up -- much of the time in CI is not in testing.  I'll continue to keep an eye on it.

Edit: I also see by inspection that with -N2 the tests are definitely being run 2 at a time.

Edit 2:  It seems that older pipelines run the tests at double speed, but only half as many at once.  There is still a local speed up depending on the number of cores your machine has.  But this is overall disappointing...

Edit 3: tests are mysteriously failing on cabal 900 and not cabal 810!

tests/test.hs
```diff
@@ -53,6 +53,9 @@ data DependentTests = DependentTests
   , toplevel :: [TestTree]
   }
 
+newtype SequentialTests = SequentialTests
+  { getTests :: [(TestTree, Int)] }
```
facundominguez: This could be a good place to document the meaning/purpose of the `Int` values stored in `getTests`.

tests/test.hs
```diff
@@ -432,17 +435,37 @@ selfTests
...
+  [ testGroup (name <> "-libs") libTests
+  , after AllFinish (name <> "-libs") $ testGroup name nonlibTests
   ]
 
+-- | Creates a [TestTree] that runs without parallelism
+testSequentially :: String -> SequentialTests -> [TestTree]
+testSequentially name (SequentialTests indexedTests) =
+  let grouped = (\(t, n) -> (testGroup (name <> show n) [t], n)) <$> indexedTests
+      -- turns i.e. [a, b, c, d] into [(a, b), (b, c), (c, d)]
+      pairs = zip <*> tail $ grouped
```
facundominguez: I did struggle to figure this one.  How about
```
pairs = zip grouped (tail grouped)
```
?

tests/test.hs
```diff
@@ -432,17 +435,37 @@ selfTests
...
+      deps = (\((_, n1), (t2, _)) -> after AllFinish (name <> show n1) t2) <$> pairs
+  in
+    case grouped of
+      [] -> []
+      (t:_) -> fst t : deps
```
facundominguez, line 457: Is this function similar in effect to
```
testSequentially name (SequentialTests indexedTests) =
    localOption (NumThreads 1) (testGroup "sequential" $ map fst indexedTests)
```
?

tests/test.hs
```diff
@@ -439,17 +439,23 @@ testGroupsWithLibs name (DependentTests libTests nonlibTests) =
...
-      deps = (\((_, n1), (t2, _)) -> after AllFinish (name <> show n1) t2) <$> pairs
+      deps = (\((_, n1), (t2, _)) -> after AllFinish (mkName n1) t2) <$> pairs
   in
     case grouped of
       [] -> []
       (t:_) -> fst t : deps
+  where
+    -- Supports up to 1000 tests -- anymore, and you'll get a dependency cycle
+    -- issue because "X-100" matches, say"X-1000"
+    mkName n = printf "%s-%03d" name n
```
facundominguez: I'd say you want to throw an error if `n` exceeds this limit instead of leaving it to tasty to fail.

That said, is it possible to write the pattern string so the match requires the full name?

tests/test.hs
```diff
@@ -414,29 +411,35 @@ textOrder = Just . mkOrder $
...
 selfTests
   = group "Self" [
-      testGroup "liquid"      <$> dirTests "src"  [] ExitSuccess (Just " SAFE ") (Just " UNSAFE ")
+      testGroupsWithLibs "liquid" <$> dirTests "src"  [] ExitSuccess (Just " SAFE ") (Just " UNSAFE ")
+  ]
+
+testGroupsWithLibs :: String -> ([TestTree], [TestTree]) -> [TestTree]
+testGroupsWithLibs name (libTests, nonlibTests) =
+  [ testGroup (name <> "Lib") libTests
+  , after AllFinish (name <> "Lib") $ testGroup name nonlibTests
```
facundominguez: Would `AllSucceed` work better here?

tests/test.hs
Fizzixnerd: I can write the `mkName` so that it appends a dot, i.e. `printf "%s-sequential-%d."`.... this should fix this issue.  Not sure why I didn't think of it before!

tests/test.hs
Fizzixnerd: Personally, I think we should retain the ability to run all the tests, even if previous ones fail, since only certain tests in `nonLibTests` will rely on the `Lib` tests, and it would be inconvenient to have the failure of one of the lib tests shadow the failure of other nonlib ones.

tests/test.hs
Fizzixnerd: Oddly, no.  I tried this, and it ran on all my CPUs regardless.  However, doing `stack --nix test --ta '-j1 -p sequential'` DID work.  I do like the idea of adding `sequential` to the test name, however, so that you can run the sequential tests by name.

tests/test.hs
Fizzixnerd: Sorry, I always use this form because of https://stackoverflow.com/questions/3774247/what-do-we-call-this-new-higher-order-function/3775293#3775293.  I will change it!

tests/test.hs
Fizzixnerd: I refactored this to just not store the Int.

Fizzixnerd:
I've added the failing test to the text ignore list... quoting from my slack post:

> Having trouble with CI in https://github.com/ucsd-progsys/liquidhaskell/pull/1925 : not sure why, but cabal_900 is failing on Data/Text/Foreign.hs.  I look at a previous working CI job, https://app.circleci.com/pipelines/github/ucsd-progsys/liquidhaskell/1178/workflows/cdc139d2-ee49-45d9-9b90-7d3cd24d9f6c/jobs/7426?invite=true#step-112-1302, and noticed this test is not even run here, but the test is not in the text-ignored list. https://github.com/ucsd-progsys/liquidhaskell/issues/1491 points out a similar issue someone was dealing with?  Also, the test name is very strange when given to "rerun it", mentioning typeclass-pos ... I'm tempted to add this to the ignore list for test, but I don't understand why I have to when it wasn't being run before!  Anyone have any ideas?

I have ignored this test because it was only being run on the 810 CircleCIs, and fails on the 900 one.  I still do not know why the test wasn't picked up before (but is now), which is a bit worrying... I'm not sure a better way to deal with it though, since it was already being ignored (somehow).

tests/test.hs
```diff
@@ -443,19 +443,16 @@ testGroupsWithLibs name (DependentTests libTests nonlibTests) =
 
 -- | Creates a [TestTree] that runs without parallelism
 testSequentially :: String -> SequentialTests -> [TestTree]
-testSequentially name (SequentialTests indexedTests) =
-  let grouped = (\(t, n) -> (testGroup (mkName n) [t], n)) <$> indexedTests
-      -- turns i.e. [a, b, c, d] into [(a, b), (b, c), (c, d)]
-      pairs = zip <*> tail $ grouped
+testSequentially name (SequentialTests tests) =
+  let grouped = (\(t, n) -> (testGroup (mkName n) [t], n)) <$> (zip tests [0 :: Int ..])
```
facundominguez, line 451: The need to create a `testGroup` is that there is no way to get the name of the given `TestTree`, right?
If that is the case, a comment informing this could be helpful if someone wants to change it later on.

tests/test.hs
```diff
@@ -779,6 +779,7 @@ textIgnored :: [FilePath]
 textIgnored 
   = [ "Setup.lhs"
     -- , "Data/Text/Axioms.hs"
+    , "Data/Text/Foreign.hs"                -- Breaks on GHC 9.0.1
```
facundominguez, line 787: IIUC this disables the test in all configurations, not just GHC 9.0.1.

There is this PR that disabled this test in the CI configuration: https://github.com/ucsd-progsys/liquidhaskell/pull/1872

I guess the options are:
1. Use CPP macros to skip tests for ghc9, or
2. Update the CI configuration after the changes to the test name

In general, I prefer to avoid CPP, but you let me know if the current arrangement is too confusing.

Fizzixnerd: Because 1) `test.hs` already makes use of CPP; 2) the fact there are two ignore lists for tests is highly confusing (to me, at least); and 3) the name of the test in question is generated and depends on the number of tests run before it, and so cannot be reliably managed outside of `test.hs`; I propose that CPP is the preferable answer in this case.

facundominguez: Works for me.

tests/test.hs
Fizzixnerd: This is correct, and a comment has been added.

facundominguez (approved):
Thank you, Matt!

Issue #159, opened 2013-12-03
recursion via data-types or references makes termination checker unsound 


nikivazou:
It seems that we can encode the fixpoint or the Y combinator (ie, recursion) with data types[1]:

```
newtype Rec a = In { out :: Rec a -> a }

y :: (a -> a) -> a
y = \f -> (\x -> f (out x x)) (In (\x -> f (out x x))) 
```

Since `y` is not a recursive function liquidHaskell will trivially assume that y always terminates.

With `y` we can define our unsound example (see tests/todo/Recursion.hs):

```
{-@ foo :: n:Nat -> {v:Nat | v < n} @-}
foo :: Int -> Int
foo = y go
  where go f n = if n > 0 then n-1 else f n

prop = let x = 0 in
       liquidAssert ((\n -> 0==1) (foo 0))
```

So, maybe in the termination checker we should reason about data types like Rec, 
ie recursive types that appear in negative positions. Concretely, functions that open such types should be considered as diverging. 

[1] http://en.wikipedia.org/wiki/Fixed-point_combinator

ranjitjhala:
This seems related to the covariant/contravariant discussion from earlier today?

nomeata:
Is there a plan here? Does LH currently have a concept of unsafe types that it rejects? I guess you only want to reject these types in proofs, but not in all programs, or something...

BTW, higher rank polymorphism can also lead to recursion, so a plain positivity-checker is not sufficient.

nikivazou:
No current plan. I think Rustan has worked around this in Dafny, but as far as I know, has not published his technique. 

An idea is to always generate termination error when you have negative or higher rank types?

nomeata:
Yeah, that’d be the conservative approach. But it is a policy decision on your (i.e. the Liquid Haskell authors) to weigh soundness-in-all-cases vs, practicality. If you document that “Liquid Haskell is not sound in the presence of negative types and higher rank types”, then there is no bug at least, and people who shoot in their foot are welcome to do so :-)

ranjitjhala:
The conservative approach sounds good to me. We can have some mechanism to
“relax” the checks for such cases if the programmer so desires?

mgree:
We independently ran into this (per email); posting the file here for posterity. A negative occurrence check would rule out the following program:

```Haskell
module Omega where

import qualified Data.Map as Map
import Data.Map (Map, (!))

-- | Standard domain for LC + Bool
[...]

PR #1927, opened 2022-01-19
Add positivity checker

nikivazou:


src/Language/Haskell/Liquid/GHC/Play.hs
```diff
@@ -10,12 +10,98 @@ import Prelude hiding (error)
...
+-------------------------------------------------------------------------------
+
+getNonPositivesTyCon :: [TyCon] -> [(TyCon, [DataCon])]
+getNonPositivesTyCon tcs = Mb.catMaybes $ map go (M.toList $ makeAppearences tcs)
+  where 
+    go (tc,dcocs) = case filter (\(_,occ) -> elem tc (negA occ)) dcocs of  
+                      [] -> Nothing
+                      xs -> Just (tc, fst <$> xs) 
+
+data TyConAppearance = TyConA {posA :: [TyCon], negA :: [TyCon] }
```
facundominguez: How about documenting the meaning of the fields in this type?

src/Language/Haskell/Liquid/GHC/Play.hs
```diff
@@ -10,12 +10,98 @@ import Prelude hiding (error)
...
+import Language.Haskell.Liquid.GHC.API as Ghc hiding (substTysWith, panic,showPpr)
 import Language.Haskell.Liquid.GHC.Misc ()
 import Language.Haskell.Liquid.Types.Errors
 
 
+-------------------------------------------------------------------------------
+-- | Positivity Checker -------------------------------------------------------
+-------------------------------------------------------------------------------
+
+getNonPositivesTyCon :: [TyCon] -> [(TyCon, [DataCon])]
```
facundominguez, line 28: How about documenting the output of this function? Perhaps documenting an example of an input and an output would be telling.

src/Language/Haskell/Liquid/Types/Errors.hs
```diff
@@ -1043,6 +1049,10 @@ ppError' _ dCtx (ErrRewrite _ msg )
         $+$ dCtx
         $+$ nest 4 msg
 
+ppError' _ dCtx (ErrPosTyCon _ tc dc)
+  = text "Non strictly positive occurence of" <+> tc <+> "in" <+> dc 
+        $+$ dCtx
```
facundominguez, line 1054: If the error could reference a github issue or some other pointer discussing the purpose of the check the user might have a better time figuring it out.

Also, maybe the error message could mention the flag to disable the check and warn appropriately about using it.

facundominguez (commented):
I didn't try to understand the code, but made a pass trying to identify what could be interesting to have a bit more explained.

ranjitjhala:
These are excellent suggestions @facundominguez !

nikivazou:
Indeed, great suggestions, esp. the documenting what can go wrong with the non positive occurrences! 
I will do it tomorrow. 

I was happily surprised that only one test failed! (All these complaints for only one test :))

ranjitjhala:
Complaints? You are referring to the oopsla reviewers right? :-)

nikivazou:
Not only. 

This check addresses #159 and the known TT fact that negative occurrences are  bad (https://stackoverflow.com/questions/12651146/why-inductive-datatypes-forbid-types-like-data-bad-a-c-bad-a-a-where-th).

facundominguez:
The comments look great, thank you!

PR #1928, opened 2022-01-21
More edits on positivity Check

nikivazou:
- Correct error message: we only check positivity (not strictly)...
- Fix the check on type applications 
- Add example of Guarded Negative Case


nikivazou:
After struggling with circle and linking with base it is green, so I will merge! 

The critical edit is local in the positivity checker: it now treats properly type arguments of type constructors. 
But, after this edit liquid-vector and liquid-base have negative occurrences, so I had to turn off the positivity checker when checking them. 
For liquid-vector it was easy, because it is local, but for liquid-base I ended up linking to a specific commit (on the develop branch) which is the most dangerous edit of this PR: https://github.com/ucsd-progsys/liquidhaskell/pull/1928/files#diff-0dc22391de47454ca5291fe5d1e756176d28cceee149725c7e53147f665bd821R28

ranjitjhala:
Is there a way to disable the check on a "per-type" basis? So

```
{-@ allow-negative TYPENAME @-}
data TYPENAME = ...
```

nikivazou:
Nope. It is not that difficult to implement, but since negative types do not appear that often, it is not clear to me that it worths the effort...

Issue #1912, opened 2021-11-30
REST blocks indefinitely


facundominguez:
I've got a minimal reproducer for the strange behavior I was finding when verifying stitch-lh with the latest changes to Liquid Haskell.

Inserting print statements it looks as if the function `Language.Fixpoint.Solver.PLE.evalREST` from `liquid-fixpoint` never returns.

I'm sharing a branch with a new failing test [here](https://github.com/ucsd-progsys/liquidhaskell/compare/develop...facundominguez:fd/rest-bug). This version is from before the interpreter-related changes in `liquid-fixpoint`.

Trivial changes make the verification pass, so your best chance to reproduce it is using `nix` and `stack`. But perhaps you can make it work with `cabal-install` too.

I run the test with either
```
stack exec liquidhaskell -- tests/pos/Rest.hs
```
or
```
stack exec -- $(stack path --dist-dir)/build/test/test -p Tests.Macro.unit-pos.Rest.hs
```

As one would expect, I can reproduce the failure with the generated `Rest.hs.bfq` file
```
stack exec fixpoint -- --eliminate=some --rewriteaxioms tests/pos/.liquid/Rest.hs.bfq
```

The file `Rest.hs.fq` can't be fed to `fixpoint` because of parsing errors. Though maybe I could work on fixing the pretty printer if necessary.

zgrannan:
I will investigate this, but it will probably be next week.

facundominguez:
Thanks!

facundominguez:
I was debugging this a bit today. Turns out that `deANF` sometimes fails to terminate. The algorithm tries to apply a substitution on an expression until a fixpoint is achieved:
```Haskell
deANF ctx = subst'
  where
    subst' ee =
      let ee' = subst su ee
      in if ee == ee'
          then ee
          else subst' ee'
```
This strategy doesn't terminate because the substitutions that are produced sometimes contain mutually referencing substitutions like
```
[lq_anf$##7205759403792805575##d379:=e##a1Qi]
[lq_anf$##7205759403792805583##d37h:=
   Language.Stitch.LH.Check.Cons##lqdc##$select##Language.Stitch.LH.Check.Cons##1
      lq_anf$##7205759403792805584##d37i
]
[lq_anf$##7205759403792805584##d37i:=
   Language.Stitch.LH.Check.Cons lq_anf$##7205759403792805583##d37h ctx0##a1Qg
]
```
[...]

PR #1929, opened 2022-02-04
Update rest-rewrite

facundominguez:


facundominguez:
Updating `rest-rewrite` is necessary if we are to debug #1912 with the latest changes to rest-rewrite.

I had to [relax an upper bound in rest-rewrite](https://github.com/zgrannan/rest/pull/4) and created a patch. That's why the current PR is pointing to my fork instead of Zack's.

nikivazou:
Perfect! I am merging and when @zgrannan updates REST, we can link it back to the main branch!
Thanks!

PR #1946, opened 2022-02-28
Add stitch-lh directly

### LINKED ISSUE #1933: Add stitch-lh to the suite of benchmarks
Now that stich-lh seems to be building again with the fixes for #1912, perhaps it is a good time to consider adding it to the benchmark (or maybe Macro?) suite. This would prevent accidental breakage of a moderately interesting use case of LH.

The code takes near 1 minute to verify, and the longest modules take between 20 and 30 seconds, which perhaps it is acceptable to run on every PR.

I'm asking here ahead of making the change in case someone has a better idea.

Fizzixnerd:
Closes #1933 (?)

Adds `stitch-lh` directly as a submodule, as per discussion in above issue.


Fizzixnerd:
I was speaking with @facundominguez today, and he was suggesting adding the stitch-lh stuff:
1. as a benchmark used in test.hs so that we have runtime metrics, and
2. not as a submodule, to reduce the overhead of making breaking changes.

I will do this.

Fizzixnerd:
I messed up my history pretty badly on this one, so I just committed the final changes in the end.

benchmarks/stitch-lh/LICENSE
```diff
@@ -0,0 +1,28 @@
+Copyright (c) 2020, Richard Eisenberg
+Copyright (c) 2021, Facundo Domínguez
+All rights reserved.
```
facundominguez, line 3: @ranjitjhala, is it ok to copy this license notice as is?

ranjitjhala:
Sure thing!

Fizzixnerd:
Note that stitch-lh doesn't compile properly on GHC 9 right now.  I have removed the offending files using an ignore list and CPP.

Fizzixnerd:
Going to open an issue to track the ignores in [c805dca](https://github.com/ucsd-progsys/liquidhaskell/pull/1946/commits/c805dca951ab5fb4f427d6fb39625367e2ce93ce) and [05cf0f6](https://github.com/ucsd-progsys/liquidhaskell/pull/1946/commits/05cf0f6e60e27ec5d685787061f7aca58934b409)

facundominguez:
Bravo @Fizzixnerd!

PR #1949, opened 2022-03-03
Add `Lib` prefix to some test filenames

Fizzixnerd:
This changes some `todo` tests so that they will be correct if/when they are updated.


Fizzixnerd:
It seems there is an ssh server failure?

facundominguez:
I think you need to apply this fix to the LH repo: https://github.com/ucsd-progsys/liquid-fixpoint/commit/4826df85fee0aa3ffee426e139d1481cd8e75560

facundominguez:
The `stack_810` job failed. The error message is talking about ghc 8.10.4 which `stack.yaml` is trying to override to 8.10.7. Maybe it is worth trying to upgrade the stackage snapshot to use ghc 8.10.7 without overrides.

```
T1649MeasuresDef.hs:                    FAIL (0.23s)
liquidhaskell    >         tests/test.hs:556:
liquidhaskell    >         Wrong exit code; output:
liquidhaskell    >           > stack: internal error: timerfd_settime: Invalid argument
liquidhaskell    >           >     (GHC version 8.10.4 for x86_64_unknown_linux)
liquidhaskell    >           >     Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
liquidhaskell    >           > Aborted (core dumped)
liquidhaskell    >         
liquidhaskell    >         expected: ExitSuccess
liquidhaskell    >          but got: ExitFailure 134
liquidhaskell    >         Use -p '/T1649MeasuresDef.hs/' to rerun this test only.
```

Fizzixnerd:
There are now two level of dependencies allowed.  Let me know if I should just generalize to `n` once and forall, as two seems a bit awkward.

stack.yaml
```diff
@@ -24,7 +24,7 @@ extra-deps:
 - git: https://github.com/facundominguez/stitch-lh
   commit: 8e130a40376cdf9afaeb6e3e4322e1069339bb51
 
-resolver: lts-18.14
+resolver: lts-18.27
 compiler: ghc-8.10.7
```
facundominguez: I think we can drop this `compiler:` line now.

facundominguez:
We had a conversation with @Fizzixnerd where he proposed simplifying a bit these tests by building/verifying all Lib modules sequentially ahead of running the tests concurrently, so we don't have to introduce multiple levels of concurrency as this PR was exploring so far. This would affect only unit tests, and not the benchmarks which would always run sequentially.

One other thing to add in this PR is some documentation explaining the concurrency scheme being implemented, there would be some documentation in the README for people wishing to run tests, and some documentation in the source code for people wishing to modify the tests.

Fizzixnerd:
Let me know if the documentation is unclear or needs to be expanded.

stack.yaml
Fizzixnerd: Changed.

README.md
```diff
@@ -128,6 +128,53 @@ You can directly extend and run the tests by modifying
 
     tests/test.hs
 
+### Parallelism in Tests
```
facundominguez, line 131: The documentation looks very good. I propose we keep the details closer to the code by moving this section to a note in `test.hs`. But here we could say

```
### Parallelism in Tests

Most tests run in parallel, with a few module dependencies built sequentially in advance.
Benchmarks run sequentially after all other tests have finished. For detail on adding tests,
see note [bla] in `tests/test.hs`.
```

tests/test.hs
```diff
@@ -48,11 +48,20 @@ import Paths_liquidhaskell
...
+-- tests are run in *parallel* after all dependencies have been run. For more
+-- information on specifying the order of the dependencies, see the
+-- `SequentialFileOrder` sections below; by default they are run in *reverse*
+-- alphabetical order.
 data DependentTests = DependentTests
   { dependencies :: [TestTree]
   , toplevel :: [TestTree]
   }
 
+-- | Tests which are purely sequential (like most benchmarks) can use this
```
facundominguez: Is there any reason to run some benchmarks in parallel?
I'm thinking it could help to make all benchmarks sequential to simplify explanations. Also, the measurements would be more comparable across runs if they are built sequentially, though I still would expect VMs to introduce a lot of variability.

Fizzixnerd: This is a good point, I will change it.

EDIT: Changed.

README.md
Fizzixnerd: Changed.

README.md
```diff
@@ -128,6 +128,12 @@ You can directly extend and run the tests by modifying
...
+
+Most tests run in parallel, with a few module dependencies built sequentially in
+advance. Benchmarks are run sequentially after all other tests have finished.
+For details on adding tests, see note [Parallel_Tests] in `tests/test.hs`.
```
facundominguez, line 135: Note [Parallel_Tests] doesn't exist, but there's a similarly named note. Worth renaming to make it easier to grep.

Also, the windows failure confuses me. I don't know what command the error message refers to.

facundominguez:
The windows job is failing the same in other PRs like #1953, so I'm merging this. We can investigate that separately.

Thanks @Fizzixnerd for getting this one through!

PR #1951, opened 2022-03-15
Drop dependency on optparse-simple

### LINKED ISSUE #1751: optparse-applicative-0.16.0.0 works just fine
The Cabal file excludes optparse-applicative-0.16.0.0 from being used as a dependency. When I ignore that constraint, however, the build works just fine with optparse-applicative-0.16.0.0. Would it be possible to please lift that constraint in the Cabal file?

googleson78:
We were only using it to generate some version info. Instead, we inlined
the simpleVersion function from optparse-simple.

Fixes #1751

### DIFF
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -177,6 +177,7 @@ library
                       Language.Haskell.Liquid.UX.DiffCheck
                       Language.Haskell.Liquid.UX.Errors
                       Language.Haskell.Liquid.UX.QuasiQuoter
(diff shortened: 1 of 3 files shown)


googleson78:
Whoops, my "auto-clear trailing whitespace" has kicked in. I'll remove those changes if you don't like this.

ranjitjhala:
No worries !

ranjitjhala (approved):
Looks great thanks!!!

PR #1953, opened 2022-03-24
Update LF after implementing --no-interpreter there

facundominguez:
Accompanying PR for https://github.com/ucsd-progsys/liquid-fixpoint/pull/537.


facundominguez:
Ah, I just discovered this syntax in one of the tests: `--useinterpreter=False`. I think it is still better to have `--no-interpreter` for consistency with LH.

PR #1965, opened 2022-04-08
Fix calculation of eqRec when renaming Equations

facundominguez:


facundominguez:
Aparently the `eqRec` field is not used much. I discovered it was wrongly set when I tried to use it in PLE. Perhaps some refactoring is possible in LH, so the name of equations does not need to be changed after the equation is created, but I haven't analyzed that far.

Issue #1962, opened 2022-04-07
Remove -Wno-name-shadowing from liquidhaskell.cabal


facundominguez:
In a similar spirit to https://github.com/ucsd-progsys/liquid-fixpoint/issues/538, enabling this warning would help preventing some mistakes. But please, let me know if there are reasons to keep it disabled.

tbidne:
If no one else is interested, I'd be happy to get started on this.

ranjitjhala:
That would be fantastic, thank you so much!

PR #1967, opened 2022-04-11
Remove -Wno-name-shadowing from liquidhaskell

tbidne:
Addresses https://github.com/ucsd-progsys/liquidhaskell/issues/1962.

I cannot speak for the possible downsides of disallowing name shadowing (other than some annoyance), but after going through this, there were a number of instances like:

```haskell
foo x = ...
  where
    bar = ... x ...   -- _should_ refer to the below binding
    y = ...           -- renamed from x
```

where everything type-checked after renaming the `x` in the `where` clause because a variable with the same name+type was accidentally captured. Thus the issue is well-motivated, as shadowing bugs seem likely.

I have very little familiarity with the codebase, so my primary strategy was appending a "prime" `'` to duplicate names, or expanding the name based on its type. But please offer improvements where there is a better name.

I ran `./scripts/test/test_810.sh` and it passed, though I wasn't able to run other test suites at the moment due to errors with my environment (I'm using nix).

Thanks!


tbidne:
Looks like I have tests to fix! `cabal_900` appears to merely be some spots I missed behind CPP. Once I get my dev env better suited to swapping ghc versions I can fix those. The others I don't understand, as the reported error appears at the end of a bunch of success messages:

```
liquidhaskell    >       Language/Stitch/LH/Repl.hs:             OK (12.98s)
liquidhaskell    > (skipping as apparently already compiled ...) (skipping as apparently already compiled ...) (skipping as apparently already compiled ...) (skipping as apparently already compiled ...) (skipping as apparently already compiled ...) (skipping as apparently already compiled ...) (skipping as apparently already compiled ...) (skipping as apparently already compiled ...) (skipping as apparently already compiled ...) (skipping as apparently already compiled ...) 
liquidhaskell    > Test suite test failed
Completed 10 action(s).
Test suite failure for package liquidhaskell-0.8.10.7.1
    test:  exited with: ExitFailure 1
Logs printed to console
```

nikivazou:
Thanks! 
You have to scroll up to see the failing tests. 
For example, in `stack_810` the test `LocalHole.hs` of `Error-Messages` fails.

tbidne:
Hi @nikivazou! Thanks for the help. Indeed I see you are correct; I should have scrolled up.

Anyway, at the moment it appears CI is passing. I am a bit mystified by this as earlier runs had some failures, and I did not change anything other than rebasing the latest `develop`. I have had some trouble running tests reliably locally, so I have been relying on CI for guidance. I'm not sure what to make of this, so any guidance would be greatly appreciated.

src/Language/Haskell/Liquid/Bare/Axiom.hs
```diff
@@ -181,10 +181,10 @@ data AxiomType = AT { aty :: SpecType, aargs :: [(F.Symbol, SpecType)], ares ::
 
 -- | Specification for Haskell function
 axiomType :: Bool -> LocSymbol -> SpecType -> AxiomType
-axiomType allowTC s t = AT to (reverse xts) res  
+axiomType allowTC s st = AT to (reverse xts) rytpe
```
facundominguez, line 184: rtype?

tests/test.hs
```diff
@@ -0,0 +1,1086 @@
+{-# LANGUAGE CPP  #-}
```
facundominguez: This file was removed from `develop`. Likely, it needs to be removed from this PR as well.

facundominguez (commented):
@tbidne this is a lot of work! I have 20 more files to read. But so far it is looking good.

tbidne:
Thanks! Yes it is quite a lot, so please take as much time as you need. No rush on my account. I can also split this up on a per-package basis, if that is easier.

I will wait until you have finished this round of comments before pushing up any of the requested changes, in either case.

facundominguez (approved):
LGTM! Thanks @tbidne. I'm merging now. Any other PR will conflict with this otherwise. There would be only a comment to address, which can be done in another PR.

tbidne:
@facundominguez Thanks for merging! Regarding the `rtype` term: That variable has type `RType RTyCon RTyVar (UReft F.Reft)))`, so I renamed it `res -> rtype` (for lack of a better name), to disambiguate it from the `res` further down on line 192. If you would like to me to make any changes in a subsequent PR, please let me know.

facundominguez:
> Regarding the rtype term: That variable has type ...

That's alright. My earlier comment refers to a typo. It says "rytpe" where I think you meant **rtype**.

tbidne:
Oh! Thank you, you are exactly right.

Issue #1955, opened 2022-03-27
Use `cabal-install` instead of the machinery in `tests/test.hs`


Fizzixnerd:
I have noticed that `tests/test.hs` is basically reproducing a build system, complete with manual dependency tracking and all the fun that entails.  I propose that we instead switch to using cabal.  I have a proof of concept on my for this with `tests/pos/`, which I will link to below.

## benefits of using cabal:

- automatic dependency tracking
- easier forced test suite running with forced rebuilds
- correct parallelism without hacks
- automatic tooling support! (editors will highlight liquid errors in test files, reducing the need to rerun the test suite after making changes and making development more interactive)
- more accurate and more detailed profiling via https://github.com/qnikst/ghc-timings-report (with `-ddump-timings` and `-ddump-to-file`)
- theoretically faster test running because of the lack of needing to startup ghc for each test file
- easier to add benchmarks (since they are just another cabal project)
- still supports partial test runs, but correctly connects dependencies (i.e., `cabal build MyFooTest` will build `MyFooLib` too (if needed!))
- much smaller footprint of `test.hs`, since it doesn't double as a build system; it only needs to parse the output of one

## complications:

- negative tests require a rework -- test.hs is still necessary to parse the output of cabal (or to use cabal as a library, whichever is easier!) for detecting specific error messages and the like
- need to adjust profiling infrastructure to new profile data -- see however ghc-timings-report/tmp/ in the below branch for the output of the new profiling tools.  Notice especially `tests-0.1.0.0.html` for a summary of how long the compiler took on each stage of compilation.
- Renames need to happen, but see `lh-renamer` in below branch for something that does _most_ of the work.
- legacy executable...

Fizzixnerd:
https://github.com/Fizzixnerd/liquidhaskell/tree/test-hs-refactor2

Fizzixnerd:
For simplicity's sake, here are screenshots of tests-0.1.0.0.htm
![Screenshot from 2022-03-27 13-40-50](https://user-images.githubusercontent.com/5892651/160293688-a296cf2d-de75-4d62-8fef-223c2fbca98a.png)
![Screenshot from 2022-03-27 13-40-54](https://user-images.githubusercontent.com/5892651/160293690-9092863d-5753-4cd9-a836-1412d8ec6f6b.png)
l:

Fizzixnerd:
I did some preliminary profiling of the cabal build, and these are the results:

For the tests.cabal run,
```bash
[...]

PR #1968, opened 2022-04-11
Catch various panics

Fizzixnerd:
As noted in #1955 , panics are thrown by GHC at various times when they should probably be regular error messages.  I don't think this was an intentional change, and was likely introduced by cd2640cbe98b9ee2ea98222bd204dc1360c1f64e.  However, since the tests essentially grep for an error message, and that error message was contained in the panic, they didn't fail.

This catches all of the panics I could find so far.


ranjitjhala:
Looks like some weird failure with the QQTySig! If it works locally let’s just retry the test? Thanks so much for this PR I’ve wanted this for a while!!!

Fizzixnerd:
Looks like it's good now.

ranjitjhala:
Awesome thank you!!!

Issue #1980, opened 2022-05-06
Add a plugin flag to signal expected failures


facundominguez:
Tests of verification failures currently require to parse the output of compilation failures to check that the errors are the expected ones. If the plugin offered a flag to specify which failures are expected, then the output wouldn't have to be parsed. This would be another step to simplify the testsuite. 

I'm proposing that we add flags
* --expectErrorIncluding "some string" 
* --expectAnyError

The following has verification fail only if no verification failures are found.
```
{-@ LIQUID "--expectAnyError" @-}
```

The following has verification fail only if an error message does not include either letters `a` or `b`.
```
{-@ LIQUID "--expectErrorIncluding \"a\"" @-}
{-@ LIQUID "--expectErrorIncluding \"b\"" @-}
```

In principle, all expected errors are suppressed in the output. This will help the testsuite printing in the output only the errors that aren't expected. We could add later on a flag `--noSuppressExpectedErrors` to have them printed to stderr if needed.

Fizzixnerd:
This looks good!  One thing that might simplify things is if we only allow the flag once, at least at first.  I don't expect the use-case of "fails with X *or* fails with Y" to come up very often, but I could be wrong.

facundominguez:
> I don't expect the use-case of "fails with X or fails with Y" to come up very often, but I could be wrong.

Ok, but then we have to change the meaning of `--expectErrorIncluding "some string"` to `--expectAtLeastOneErrorIncluding "some string"`. It must be common for a test to fail with multiple verification errors.

Update: We discarded this option in later conversations since we do want to make sure that no new errors appear.

Fizzixnerd:
I have been working on this issue.  So far I have the filtering logic down, I think.  I have a bunch of tests; not sure if there is a place already for unit tests in `liquidhaskell` that aren't source files checked by the plugin.  If not, I'll make a new suite.

I changed the code that folds over the pragmas to do them all at once.  Before it would iteratively do them, and this would break list-like arguments that can be supplied more than once.  I also spent a bunch of time chasing down a _second_ reason the flag wasn't showing up, and that's because I didn't realize Liquid has two configs: a module config, and a global config.  The global config is formed from (I believe) `LIQUID_OPTS` and command line flags, while the module config includes the pragmas specific to that file.  I was using the global config in a number of places, but was using pragmas to control the option, so it wasn't working.

[...]

PR #1960, opened 2022-04-04
Testing Refactor to Use Cabal/Stack

Fizzixnerd:
Related to #1955 

Things are ugly at the moment as I iterate on the design, but I wanted to have a place where people could see the changes being proposed, and comment if they wanted to.

The history will be rewritten before becoming non-draft, and I need to rebase, but here are some interesting things to look at:

- `tests/tests.cabal` for the cabal file
- the contents of `tests/harness` for new test driver
  - `tests/harness/Test/Parse.hs` for the cabal/stack output parser
  - `tests/harness/Test/Summary.hs` for the summarizer which will be used to generate test reports in xml
- `lh-renamer` for the renamer thing that will not be a part of the final merge
- `ghc-timings-report` which will likely be spun off into a submodule in the final merge, but is included verbatim here for now

PR #1983, opened 2022-05-11
Translate crashes to OutputResult with a message

facundominguez:
If PLE crashes, sometimes LH would not print the message in the console.

Turns out that in some cases only the `orMessages` field of `OutputResult` would be printed when the interesting information is in `orHeader`.

This PR strives to only have non-empty `orMessages` produced from crashes.

### DIFF
--- a/src/Language/Haskell/Liquid/UX/CmdLine.hs
+++ b/src/Language/Haskell/Liquid/UX/CmdLine.hs
@@ -811,6 +811,11 @@ resDocs _ (F.Safe  stats) =
     orHeader   = text $ "LIQUID: SAFE (" <> show (Solver.numChck stats) <> " constraints checked)"
   , orMessages = mempty
   }
+resDocs _k (F.Crash [] s)  =
+  OutputResult {
+    orHeader = text "LIQUID: ERROR"
+  , orMessages = [(GHC.noSrcSpan, text s)]
+  }
 resDocs k (F.Crash xs s)  =
   OutputResult {
     orHeader = text "LIQUID: ERROR" <+> text s


facundominguez:
@Fizzixnerd, let me know what you think.

Fizzixnerd:
This looks good to me.  To be clear, the problem was that I stopped printing the header in #1960 without fully understanding the consequences in the case of an `LIQUID ERROR`.

src/Language/Haskell/Liquid/UX/CmdLine.hs
```diff
@@ -811,6 +811,11 @@ resDocs _ (F.Safe  stats) =
...
   , orMessages = mempty
   }
+resDocs _k (F.Crash [] s)  =
+  OutputResult {
+    orHeader = text "LIQUID: ERROR"
+  , orMessages = [(GHC.noSrcSpan, text s)]
+  }
 resDocs k (F.Crash xs s)  =
   OutputResult {
     orHeader = text "LIQUID: ERROR" <+> text s
```
Fizzixnerd, line 821: I notice that this case adds the `text s` to the output of the `messages`, so perhaps it would be good to add that to the `orMessages` too (as like the first message)?  That way we don't get inconsistent messages when there are `xs` or not but `s` is the same.

facundominguez:
Hm, let's merge this as a short term remedy. But we should undo it, and undo the suppression of messages in #1960 after implementing #1980.

src/Language/Haskell/Liquid/UX/CmdLine.hs
facundominguez: Ah, missed this before merging. Feel free to amend, though we might just go for #1980 now.

PR #1987, opened 2022-05-27
Prevent name collisions in proofs using Prop

plredmond:
change the name of the variable used in the liquid-fixpoint ProofCombinators.Prop type alias to prevent collisions

PR #1991, opened 2022-05-30
Revert "Prevent name collisions in proofs using Prop"

facundominguez:
Reverts ucsd-progsys/liquidhaskell#1987

Apparently, the fix in #1987 would address the same issue fixed in https://github.com/ucsd-progsys/liquidhaskell/pull/1943


facundominguez:
@plredmond, please let me know if this doesn't work for you.

PR #1992, opened 2022-05-30
Fail tests when there is a non-parsing failure too

Fizzixnerd:
Renames `flag` to more descriptive `isBadFlag`.

Fixes issue where only _parse_ errors caused the test suite to fail.  As of right now, there are failing tests that aren't being caught properly.  This should fail in CI, until I disable those specific tests.


Fizzixnerd:
Closes #1995.

facundominguez (approved):
LGTM!

PR #1993, opened 2022-05-30
Explain how to collect measures with the new test framework

facundominguez:
Also does some cleanups:
* Puts ghc-timing-reports in the stack and cabal configuration
* Removes the ghc-timing-reports submodule
* Has modules compiled with -fforce-recomp

cabal sometimes invokes the compiler twice on modules. When the second time the code is found up-to-date, the `.dump-timing` files are overwritten with no useful timings. To workaround this, files are compiled with `-fforce-recomp` so `.dump-timing` files are overwritten with the timings of the second compilation.


facundominguez:
@Fizzixnerd, does this look good to you?

tests/tests.cabal
```diff
@@ -94,11 +100,10 @@ executable benchmark-stitch-lh
                       -Wno-unticked-promoted-constructors
                       -fno-warn-name-shadowing
                       -fkeep-going
-                      -fforce-recomp
-                      -ddump-timings
-                      -ddump-to-file
                       -O0
 
+    if flag(measure-timings)
+      ghc-options:    -fforce-recomp -ddump-timings -ddump-to-file
```
Fizzixnerd, line 106: In my experience with playing with cabal, `-fforce-recomp` only works the first time you add it as a ghc flag (because the ghc-flags have changed) and doesn't have an effect after that.  Have you tried running the test-suite twice after this change to make sure the dumps are still picked up, even after cabal has a -fforce-recomp build in the cache?

tests/harness/Test/Parse.hs
```diff
@@ -227,9 +227,9 @@ results :: TestGroupData -> Parser (Either ResultsException [ModuleInfo])
 results TestGroupData {..} =
   (Left (UpToDateException tgdName) <$ upToDate)
   <|> (do
-          skipStartupTrash
-          xs <- P.many moduleInfo
-          skipEndingTrash
+          xs <- skipStartupTrash *> P.many moduleInfo <* skipEndingTrash
+          -- Benchmarks might run twice
+          _ys <- P.option [] $ P.many moduleInfo <* skipEndingTrash
```
Fizzixnerd, line 232: I am confused as to what this does and why it's necessary.  Why are the benchmarks run twice?  When are they run twice?  Why can we discard the second run?

tests/tests.cabal
facundominguez: The purpose of the flag is not to cause cabal to always rebuild the file. The purpose is to rebuild the file if cabal invokes ghc on it.

tests/harness/Test/Parse.hs
facundominguez: I think `cabal` is calling `ghc` twice sometimes. Thus, if I use `-fforce-recomp` ghc produces sometimes twice the output of compilation, which is what I'm parsing here.

PR #1994, opened 2022-05-30
Revert "Remove -Wno-name-shadowing from liquidhaskell"

facundominguez:
Reverts ucsd-progsys/liquidhaskell#1967

CI was green, but we discovered since then a bug in the test framework and this PR introduced some failures inadvertently.

PR #2004, opened 2022-06-06
Initial commit removing -Wno-name-shadowing

tbidne:
Follow-up to #1967 and #1994.

In order to split this up, I've only removed the flag from the `liquidhaskell` library. That still includes about 95% of the original PR, so I have additionally added `{-# OPTIONS_GHC -Wno-name-shadowing #-}` to most of the modules. They can be removed and properly addressed in subsequent PRs.


tbidne:
It appears there is still some flakiness in the tests. `ci/circleci: stack_810` passed in the previous runs and only just now failed on the latest (I was addressing the ghc 9 build).

facundominguez (approved):
Thanks @tbidne. Please, could we have new changes in separate commits? Say, one commit per file. In that way it is going to be simpler to identify which commit breaks a test.

Anyway, I think this is good to merge without further changes if we indeed have a flaky test here.

tbidne:
@facundominguez Excellent idea. If you want I can split this PR into individual commits. Otherwise I will do so in the next round.

PR #2003, opened 2022-06-03
Parse filenames for `benchmark-timings` instead of module names

Fizzixnerd:
Closes #1996.

PR #2010, opened 2022-06-16
Prettify benchmark names in reports

### LINKED ISSUE #1996: `benchmark-timings` uses module names instead of filenames
This breaks the charting code because there can be aliases between modules which share a name in different suites.

facundominguez:
Closes #1996
Continues #2003

### DIFF
--- a/benchmark-timings/app/Main.hs
+++ b/benchmark-timings/app/Main.hs
@@ -75,14 +75,14 @@ opts = info (options <**> helper)
 -- | Parse the original filename from the .dump-timings filename
 dumpFilenameParser :: MP.Parsec Void String FilePath
 dumpFilenameParser = do
-  _arch <- element
-  _ghcVersion <- element
-  _pkg <- element
-  pathPieces <- MP.manyTill element "dump-timings.json"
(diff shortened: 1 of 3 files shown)


benchmark-timings/app/Main.hs
```diff
@@ -66,14 +67,33 @@ opts = info (options <**> helper)
...
 program :: Options -> IO ()
 program Options {..} = do
-  csvFields <- for optsFilesToParse $ \fp -> do
-    Just (phases :: [Phase]) <- decodeFileStrict' fp
-    let (modName, time) = foldr (\Phase {..} (_, acc) -> (phaseModule,
-                                                          if init phaseName `elem` optsPhasesToCount then acc + phaseTime else acc)) ("", 0) phases
-     -- convert milliseconds -> seconds
-    if modName == "" then pure Nothing else pure $ Just $ PhasesSummary modName (time / 1000) True
+  csvFields <- for optsFilesToParse $ \fp ->
+    -- irrefutably get the filename and fail if we can't!
```
facundominguez:
```suggestion
```

PR #2016, opened 2022-06-23
Implement --ple-with-undecided-guards

### LINKED ISSUE #602: Disable expansion of reflected functions with unevaluated guards

facundominguez:
Closes ucsd-progsys/liquid-fixpoint#602

Companion PR of ucsd-progsys/liquid-fixpoint#603


facundominguez:
It is worth paying some attention to the modifications in the tests.

The functions `NegNormalForm.not` and `WhyLH.max` both need to be unfolded to save the user quite some trouble. In the case of `NegNormalForm.not` the SMT solver can use some reasoning on booleans, otherwise the user has to write lemmas to prove that `not` is an inverse of itself, and the de Morgan laws.

In these cases, I'm using a function `pleUnfold` that behaves like `id`, but it additionally fools the PLE algorithm into expanding the equation in which it is used. If we are to adopt this function, it should be moved somewhere like to `liquid-prelude` I guess (?).

Alternatively, we could try to come up with some heuristic to unfold functions with undecided guards. Unfolding non(mutually)-recursive functions by default could work in all the cases I've seen. But PLE doesn't have readily available a graph analysis of function definitions.

ranjitjhala:
Hmm interesting - I don’t quite follow this clever trick you are using with
pleUnfold …

facundominguez:
PLE won't unfold

```Haskell
not b = if b then False else True
```

if nothing is known about `b`. The problem for PLE is that `not` starts with an if-statement whose condition `b` is undecided. Therefore, I can give it such a decided condition by rewriting `not` to

```Haskell
not b = if True then if b then False else True else True
```

This almost works, but some simplification in LH is turning that to the original definition. Therefore, I make it a bit harder to simplify by hiding the `True` condition in an auxiliary definition.

```Haskell
not b = if bTrue then if b then False else True else True
  where
    bTrue = True
```

Now PLE unfolds `not` but the definition is ugly to read for the user. So instead, I factor out the surrounding code to `pleUnfold`, obtaining

```Haskell
not b = pleUnfold (if b then False else True)

-- | Forces the PLE algorithm to unfold the definitions that use @pleUnfold@
{-@ reflect pleUnfold @-}
pleUnfold :: a -> a
pleUnfold a = if bTrue then a else a
 where
   bTrue = True
```

ranjitjhala:
aha, got it, thanks :-)

ranjitjhala:
Would be great to have a more permanent note about this, wherever we
explain the `--ple-with-undecided-guards` flag.

ranjitjhala:
By, which I mean -- maybe we can just copy paste your above explanations
into the relevant place?

facundominguez:
> By, which I mean -- maybe we can just copy paste your above explanations into the relevant place?

I think it is good idea to document the existence of `pleUnfold`. I wasn't sure if it was good enough to keep, but in that case, we can make it less tricky by just hardcoding it in PLE. Whenever PLE finds a body that starts with an application of `Language.Haskell.Liquid.ProofCombinators.pleUnfold`, it proceeds with the unfolding.

The heuristics to decide when to unfold equations will be adjusted over time, I presume, and keeping `pleUnfold` as it is now would be a bit fragile.

facundominguez:
I documented `pleUnfold`, and put it in `Language.Haskell.Liquid.ProofCombinators`, and hardcoded it in PLE.

facundominguez:
I rebased to update LF with another fix. On my side this looks good to merge.

facundominguez:
@ranjitjhala, is this good to merge?

ranjitjhala:
oy I thought we had merged it -- lets do so! thanks @facundominguez !!!

ranjitjhala:
(and any matching PR in fixpoint too of course)

ranjitjhala:
oops looks like the `STLC.hs` test is broken in the `cabal_900` build -- any idea why?

facundominguez:
CI is passing in the `develop` branch. Do you have a link to the failing job?

ranjitjhala:
https://app.circleci.com/pipelines/github/ucsd-progsys/liquidhaskell/1353/workflows/69ecdfe8-de27-48ec-b5cc-63e79bc198f6/jobs/7954

facundominguez:
Unfortunately, rerunning the job doesn't cause it to fail. So we have a new little project in our hands: track and fix the sources of non-determinism in LH :)

facundominguez:
I created https://github.com/ucsd-progsys/liquidhaskell/issues/2020 which could help investigating these flaky failures.

michaelborkowski:
Hi Facundo, how would I try to use `pleUnfold` with a function defined with multiple equations instead of an explicit `if .. then ... else`? For instance

```
{-@ reflect isValue @-} -- meaning: is a syntactic value 
isValue :: Expr -> Bool
isValue (Prim _)        = True
isValue (FV _)          = True 
isValue (BV _)          = True 
isValue (Lambda   e)    = True 
isValue (LambdaT   k e) = True
isValue _               = False
```

facundominguez:
Hello! How about
```
{-@ reflect isValue @-}
isValue :: Expr -> Bool
isValue e0 = pleUnfold (case e0 of
  (Prim _)        -> True
  (FV _)          -> True 
  (BV _)          -> True 
  (Lambda   e)    -> True 
  (LambdaT   k e) -> True
  _               -> False)
```
?

ranjitjhala:
Awesome, lets merge!

PR #2021, opened 2022-06-29
add fancy termination

nikivazou:
Accepts the following as terminating: 

```
{-@ data ITree a where 
      ILeaf :: a -> {t:ITree a  | itsize t == 0 } 
      INode :: f:(Int -> ({tt:ITree a | itsize tt < itsize self})) 
            -> ITree a  @-}

{-@ imapTr :: (a -> a) -> t:ITree a -> o:ITree a / [itsize t] @-}
imapTr :: (a -> a) -> ITree a -> ITree a 
imapTr f (ILeaf x) = ILeaf $ f x 
imapTr f (INode n) = INode (\x -> imapTr f (n x)) 
```

* Caveat* `self` is treated as a special symbol now in the logic (currently should be defined as a `measure self :: a`) so code that uses `self` might break.


ranjitjhala:
Awesome, thanks @nikivazou !!! Btw, can you see if your method works for mutually recursive functions too -- maybe @michaelborkowski can craft some that are inspired from his proofs, but as a starting point suppose we split your example out into:

```haskell
imapTr :: (a -> a) -> ITree a -> ITree a 
imapTr f (ILeaf x) = ILeaf (f x) 
imapTr f (INode n) = INode (imapN f n) 

imapN :: (a -> a) -> (Int -> ITree a) -> (Int -> ITree a)
imapN f n = \x -> imapTr f (n x)
```

nikivazou:
@ranjitjhala you still need to provide termination metrics (doesn't solve halting problem :))
Note sure what are the metrics in the function you defined, unless we give ghost size arguments...

ranjitjhala:
Yes, I assumed that you still need the metrics -- but can never quite remember how that works for mutual-rec so left it to you :-) 

Also, sure I think its fine to have to pass in a ghost e.g. something like

```haskell
imapTr :: (a -> a) -> ITree a -> ITree a 
imapTr f (ILeaf x) = ILeaf (f x) 
imapTr f (INode n) = INode (imapN f t n) 

imapN :: (a -> a) -> ITree a -> (Int -> ITree a) -> (Int -> ITree a)
imapN f _ n = \x -> imapTr f (n x)
```

(again I left the metrics out, but if there is some way to add metrics to make the above work I think that's great!)

nikivazou:
@ranjitjhala with the ghost and metrics it works: 
https://github.com/ucsd-progsys/liquidhaskell/blob/fe56b43f9aed1503a712365a02eff46791f04c62/tests/pos/FancyMutualTerm.hs

nikivazou:
Let's merge?

ranjitjhala:
Nice! Lets merge! I think it would be good to create an issue to think about how to automatically

* detecting recursive, positive occurrence under an `->` in a data type, and then
* adding all the `this` and `tsize` strengthening using the metric for the data type
* adding the non-negative size `invariant` 

Maybe instead of automatic, it can be if the user writes something like

```haskell
{-@ data ITree a [tsize] @-} 
{-@ measure tsize :: ITree a -> Nat @-} 
```

?

facundominguez:
Just as an additional check, does any user documentation need to be updated or written after this change?

nikivazou:
@facundominguez I prefer to test it a little more before we "expose" it to the user.

PR #2023, opened 2022-07-05
Use pleUnfold when requiring more guards to unfold in PLE

facundominguez:
Companion PR of https://github.com/ucsd-progsys/liquid-fixpoint/pull/605


facundominguez:
Merging since the corresponding PR in LF was merged.

PR #2025, opened 2022-07-05
remove lowercased-files which duplicate other module-cased files

plredmond:
ref: https://liquidhaskell.slack.com/archives/C54QAL9RR/p1657047490172549

I checked out LH on a mac, with a case-insensitive filesystem as is the norm on macs, and got this error:
```
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'tests/names/neg/Local00.hs'
  'tests/names/neg/local00.hs'
  'tests/names/neg/Vector0.hs'
  'tests/names/neg/vector0.hs'
  'tests/names/neg/Vector1.hs'
  'tests/names/neg/vector1.hs'
```

The lowercased ones contain the same content as the uppercased ones, but the lowercased ones are not correctly named modules as a result.

@facundominguez


facundominguez (approved):
Thanks @plredmond!

@ranjitjhala, it looks to me like the tests in `names-neg` need a closer look. Right now they use
```
{-@ LIQUID "--expect-any-error" @-}
```
and the errors that they produce perhaps aren't the expected ones.
```
tests/names/neg/vector0.hs:26:23: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : {v : GHC.Types.Bool | (v <=> vlen Vector0.vs > (3 : int))
                                 && v == (vlen Vector0.vs > (3 : int))}
    .
      VV : {VV : GHC.Types.Bool | VV}
    .
    in the context
      Vector0.vs : {Vector0.vs : (Data.Vector.Vector GHC.Types.Int) | vlen Vector0.vs == len Vector0.xs
                                                                      && 0 <= vlen Vector0.vs}
       
      Vector0.xs : {Vector0.xs : [GHC.Types.Int] | len Vector0.xs > 0
                                                   && len Vector0.xs >= 0}
    Constraint id 10
   |
26 | prop3 = liquidAssertB (Data.Vector.length vs > 3)
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Does this have anything to do with names? Moreover, the files do not provide a clue on what they are trying to test.

Issue #2029, opened 2022-07-08
Hints suggested by HLint.


philderbeast:
Similar to https://github.com/ucsd-progsys/liquid-fixpoint/issues/552, could we chip away at the hints found in `src/` maintaining an hlint gh-action and `.hlint.yaml` as we go?

# Warnings currently triggered by your code
- [ ] Avoid lambda, 4 hints
- [ ] Avoid lambda using \`infix\`, 4 hints
- [ ] Eta reduce, 51 hints
- [ ] Functor law, 8 hints
- [x] Fuse concatMap/map, 1 hint
- [x] Fuse foldMap/fmap, 1 hint
- [x] Fuse traverse/map, 1 hint
- [x] Hoist not, 2 hints
- [x] Move brackets to avoid $, 52 hints
- [ ] Move guards forward, 2 hints
- [x] Move reverse out, 1 hint
- [ ] Parse error: on input `,', 1 hint
- [ ] Parse error: possibly incorrect indentation or mismatched brackets, 3 hints
- [x] Redundant $, 41 hints
- [x] Redundant <$>, 5 hints
- [x] Redundant bracket, 232 hints
- [x] Redundant fmap, 1 hint
- [x] Redundant id, 1 hint
- [ ] Redundant lambda, 7 hints
- [x] Redundant multi-way if, 1 hint
- [x] Redundant pure, 1 hint
- [x] Redundant return, 3 hints
- [x] Redundant uncurry, 1 hint
- [ ] Replace case with fromMaybe, 2 hints
- [ ] Unused LANGUAGE pragma, 9 hints
- [x] Use ++, 1 hint
- [ ] Use /=, 1 hint
- [x] Use :, 1 hint
- [x] Use <$>, 10 hints
- [ ] Use <&>, 3 hints
- [x] Use =<<, 1 hint
- [ ] Use camelCase, 61 hints
- [x] Use catMaybes, 1 hint
- [x] Use concatMap, 4 hints
- [ ] Use const, 5 hints
- [x] Use elemIndex, 2 hints
- [ ] Use fewer LANGUAGE pragmas, 2 hints
- [ ] Use first, 1 hint
- [ ] Use fmap, 65 hints
- [x] Use forM_, 1 hint
- [x] Use gets, 26 hints
- [x] Use guards, 1 hint
- [ ] Use if, 2 hints
- [ ] Use infix, 3 hints
- [x] Use list comprehension, 4 hints
[...]

PR #2031, opened 2022-07-09
Follow hlint suggestion, redundant brackets.

philderbeast:
See #2029.

There's a `Test suite synthesis: FAIL` with `cabal test all` that existed before this PR.


facundominguez (approved):
LGTM. Thanks @philderbeast!

PR #2032, opened 2022-07-11
Follow hlint suggestion to move brackets to avoid $.

philderbeast:
See #2029. I used HLS to apply the hints but unfortunately at the same time it removed trailing whitespace. If that's bothersome to review then please get back to me and I'll redo those files.


philderbeast:
There's a setting that hides whitespace when reviewing the diffs.

Screen Shot 2022-07-11 at 12 00 13 PM

facundominguez (approved):
LGTM!

PR #2049, opened 2022-07-16
Hlint use null and other minor hints.

philderbeast:
Minor HLint suggestions, use null, use /=, use if, use tuple section and replace case with fromMaybe, see #2029.


facundominguez (approved):
LGTM!

PR #2055, opened 2022-07-18
Follow hlint suggestion: avoid lambda.

philderbeast:
See #2029.


facundominguez:
This one is ignored in `liquid-fixpoint`. It would make sense to adopt the same list of ignores in both repos.
https://github.com/ucsd-progsys/liquid-fixpoint/pull/585

philderbeast:
I've moved this ignore to the keeper section.

PR #2059, opened 2022-07-19
Move ignores to a keeper section.

philderbeast:
See #2029. Takes the ignores triggered in liquidhaskell that we are keeping in liquid-fixpoint and moves them to a different section. This includes the ignore from #2055.


facundominguez (approved):
LGTM!

PR #2060, opened 2022-07-19
Follow hlint suggestion, move guard forward.

philderbeast:
See #2029. One of these hints may be a false positive from HLint. One is a guard but the other is just a let binding that I moved anyway to discharge the hint.


facundominguez (approved):
LGTM Thanks!

Issue #2065, opened 2022-07-22
NEVER depend on ANY module inside the tree.


philderbeast:
The module comment in `Language.Haskell.Liquid.GHC.TypeRep` is truncated:

https://github.com/ucsd-progsys/liquidhaskell/blob/c37ca0017f20070483ad0787e7f88f0223ac169a/src/Language/Haskell/Liquid/GHC/TypeRep.hs#L13-L22

It looks like the comment is a copy of a similar comment in `Language.Haskell.Liquid.GHC.Misc` but even that module doesn't heed the warning.

https://github.com/ucsd-progsys/liquidhaskell/blob/c37ca0017f20070483ad0787e7f88f0223ac169a/src/Language/Haskell/Liquid/GHC/Misc.hs#L17-L19

Could we either heed the warning or remove the warning?

facundominguez:
Looks like the warnings make sense. The modules in `GHC/*` seem to be an interface that abstracts GHC versions.

I guess hlint could be used to enforce that modules in `GHC/*` do not import other modules from LH, and other modules in LH do not import from the ghc package directly.

philderbeast:
The `GHC/*` modules are themselves within the `Language.Haskell.Liquid.*` tree so the warning is impossible to follow, as written when `GHC/*` modules depend on others within that subtree.

```
src> tree -L 4 -d
.
└── Language
    └── Haskell
        └── Liquid
            ├── Bare
            ├── Constraint
            ├── GHC
            ├── Interactive
            ├── Synthesize
            ├── Termination
            ├── Transforms
            ├── Types
            └── UX
```

philderbeast:
Spinning off a `ghc-liquid` package (#1970) for modules in `Language.Haskell.Liquid.GHC` would prevent imports from going the wrong way.

facundominguez:
> Spinning off a ghc-liquid package (#1970) for modules in Language.Haskell.Liquid.GHC would prevent imports from going the wrong way.

... or moving `Language.Haskell.Liquid.GHC` to `Language.Haskell.LiquidGHC`. I have no strong preference.

facundominguez:
Seems to be addressed by #2067. Further steps could be discussed in a new issue.

PR #2067, opened 2022-07-23
Move GHC modules to src-ghc tree.

philderbeast:
See #2065. Drops the `Language.Haskell` prefix for modules wrapping or shimming GHC. Just a small step for now. We may move these modules into their own package later on. We're not using automatic code formatting and with the search and replace I did, I've not re-aligned columns in the imports.


facundominguez (approved):
LGTM!

Issue #2068, opened 2022-07-27
`reflect` disabling termination checker?


ranjitjhala:
[Due to @facundominguez ]

Oddly, the following code _passes_

```haskell
{-@ reflect und @-}
und :: a
und = und
```

but **deleting** the `reflect` makes it properly fail with

```haskell
/Users/rjhala/research/lh-demo/src/Demo/Test.hs:9:1: error: Termination Error
Demo.Test.und
No decreasing parameter
  |
9 | und = und
  | ^^^
```

???

ranjitjhala:
@nikivazou any chance you can dig into this?

nikivazou:
Indeed very strange, yes, I will take a look!

PR #2069, opened 2022-07-30
Check Termination of Assumed Functions 

nikivazou:
Reflects are currently encoded internally as assumed. So a direct fix for #2068 is to check termination of assumed functions (and explicitly use `lazy` to disactivate the termination checker). Another fix would be to have a different internal representation of reflected functions.

### DIFF
--- a/src/Language/Haskell/Liquid/Constraint/Generate.hs
+++ b/src/Language/Haskell/Liquid/Constraint/Generate.hs
@@ -120,7 +120,7 @@ makeDecrIndex :: (Var, Template SpecType, [Var]) -> CG [Int]
 makeDecrIndex (x, Assumed t, args)
   = do dindex <- makeDecrIndexTy x t args
        case dindex of
-         Left _  -> return []
+         Left msg -> addWarning msg >> return []
          Right i -> return i
 makeDecrIndex (x, Asserted t, args)
   = do dindex <- makeDecrIndexTy x t args
--- a/tests/pos/AssumedRecursive.hs
+++ b/tests/pos/AssumedRecursive.hs
@@ -1,5 +1,6 @@
 module AssumedRecursive where
 
+{-@ lazy foo @-}
 {-@ assume foo :: a -> a @-}
 foo :: a -> a
 foo f = foo f
--- a/tests/benchmarks/stitch-lh/src/Language/Stitch/LH/Data/Map.hs
+++ b/tests/benchmarks/stitch-lh/src/Language/Stitch/LH/Data/Map.hs
@@ -78,6 +78,7 @@ empty = goEmpty ()
 -- XXX: Making goEmpty local to empty, causes LH to crash when
 -- building Check.hs.
 {-@ reflect goEmpty @-}
+{-@ lazy    goEmpty @-} -- With PR 2069 a termination error appears 
 -- XXX: For some reason, GHC aggrees to fire emptyImpl only if
 -- empty appears in an auxiliar definition like this one.
 goEmpty :: () -> Map a b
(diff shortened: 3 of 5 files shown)


src/Language/Haskell/Liquid/Constraint/Generate.hs
```diff
@@ -120,7 +120,7 @@ makeDecrIndex :: (Var, Template SpecType, [Var]) -> CG [Int]
 makeDecrIndex (x, Assumed t, args)
   = do dindex <- makeDecrIndexTy x t args
        case dindex of
-         Left _  -> return []
+         Left msg -> addWarning msg >> return []
```
facundominguez, line 123: If I had to understand this change, I'd have to figure out how `makeDecrIndex` is used and what it does. Perhaps it is a good opportunity to add some comment to `makeDecrIndex` explaining some of it. Or adding a pointer to some other place that explains it might do.

facundominguez (commented):
Thanks @nikivazou!

I would have to dive deeper in the code to understand the effect of the changes. Nonetheless, I suggested some documentation to add.

> Reflects are currently encoded internally as assumed.

I'm guessing this is only true when the reflected function lacks an accompanying specification. (?) Termination seemed to be checked otherwise.

facundominguez:
It may be worth to add a test like
```
-- Check that using assume still disables the termination check for reflected functions
{-@ reflect undefined @-}
{-@ assume undefined :: a  @-}
undefined :: a
undefined = undefined
```

I have the question really. Is using `assume` supposed to disable the termination check?

tests/pos/AssumedRecursive.hs
```diff
@@ -1,5 +1,6 @@
 module AssumedRecursive where
 
+{-@ lazy foo @-}
 {-@ assume foo :: a -> a @-}
```
facundominguez, line 4: Ah, so `assume` doesn't disable the termination check anymore. Is this ideal?

ranjitjhala:
Niki can we disable termination only if assume and not reflected?

facundominguez:
It seems 
```Haskell
{-@ reflect undefined @-}
undefined :: a
undefined = undefined
```

currently means

```Haskell
{-@ reflect undefined @-}
{-@ assume undefined :: a @-}
undefined :: a
undefined = undefined
```

How about changing it to mean instead

```Haskell
{-@ reflect undefined @-}
{-@ undefined :: a @-}
undefined :: a
undefined = undefined
```
?

This would have the termination check enabled on reflected functions without affecting the behavior of `assume`.

nikivazou:
@facundominguez to be more precise, 

```
{-@ reflect undefined @-}
undefined :: a
undefined = undefined
```

Currently means 
```
{-@ assume undefined :: {v:a | v = undefined} @-}
undefined :: a
undefined = undefined
```

i.e., puts the definition of the function inside the type. This cannot be checked (because it relates the uninterpreted function `undefined` with the Haskell function) and it has to be assumed. 

Indeed we can change it so that reflection is using a different kind of assume (that checks termination) or add by default a non assumed type for reflected (so that termination is checked). 

But I like this pull request because it separates the termination checking from the assume: if you want to disable termination checking you have to explicitly write the `lazy`. So I suggest to merge that (even though it might break some old tests).

tests/pos/AssumedRecursive.hs
nikivazou: Exaclty

src/Language/Haskell/Liquid/Constraint/Generate.hs
nikivazou: `makeDecrIndex` is a termination heuristic that tries to prove that the first argument of the function that LH knows how to turn to a Nat is decreasing and creates an error if no such an argument exists.

nikivazou: So, termination is checked as follows: 
1. structural termination,
2. user specified termination metrics (`/ [len x]`), 
3. if none of the above hold `makeDecrIndex` will be attempted

facundominguez:
> So I suggest to merge that (even though it might break some old tests).

Don't have strong feelings about it. So it would be fine with me.
@ranjitjhala?

PR #2071, opened 2022-07-31
Removing shadowing 1

tbidne:
Follow up to #2004.

Apologies for taking a while to get back to this. Each commit is restricted to a single file, as planned. I have attempted to strike a balance between number of PRs and manageable diffs, thus aiming for ~100 lines (there are 47 more files with `-Wno-name-shadowing`). But let me know if I should further split these PRs up to 1 commit each.

In general my strategy is:
* Keep changes local to the function themselves.
* _Within_ the function, favor changing the widest scope possible. Imo this is safest due to the following example, which is somewhat common (and presumably caused the original bug):

```haskell
-- want to change the 'x' shadowing.
silly x = go (x+1)
  where
    go x = x * x

sillyBad x = go (x+1)
  where
    -- missed one, semantics changed!
    go x' = x' * x

-- compilation errors instead, no silent errors
sillyGood x' = go (x+1)
  where
    go x = x * x
```

### DIFF
--- a/devel/Paths_liquidhaskell.hs
+++ b/devel/Paths_liquidhaskell.hs
@@ -1,7 +1,5 @@
 {-# LANGUAGE TemplateHaskell #-}
 
-{-# OPTIONS_GHC -Wno-name-shadowing #-}
-
 module Paths_liquidhaskell where
 
 import Language.Haskell.TH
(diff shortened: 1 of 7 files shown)


facundominguez (approved):
LGTM
Thanks @tbidne!

PR #2101, opened 2022-11-11
remove nix flake

plredmond:
As with https://github.com/ucsd-progsys/liquid-fixpoint/pull/629, I'm no longer able to keep up maintaining this flake for my personal use. I suggest removing it, because it's very complex, and most users are probably satisfied with cabal and stack.

I'm fairly certain that it's not used by CI. We might confirm that with @facundominguez.


facundominguez (approved):
LGTM

PR #2104, opened 2022-11-16
Refined Vec Fixes

nikivazou:


tests/pos/T2091.hs
```diff
@@ -0,0 +1,21 @@
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE GADTs #-}
+
+import Prelude (Bool(..))
```
facundominguez, line 6: @nikivazou, should these tests run in CI? They would need to be added to `tests.cabal` if so.

facundominguez: ... and they may have to start with `module T2091 where` for compilation to succeed.

nikivazou: Nope, just did it here: https://github.com/ucsd-progsys/liquidhaskell/pull/2107

PR #2105, opened 2022-11-16
Add a pass that removes GHC.Base.$

nikivazou:
This works assuming `GHC.Base.$` will not change...

### DIFF
--- a/src/Language/Haskell/Liquid/Transforms/Rewrite.hs
+++ b/src/Language/Haskell/Liquid/Transforms/Rewrite.hs
@@ -100,7 +100,7 @@ undollar = go
  
 
 untick :: CoreExpr -> CoreExpr 
-untick (Tick _ e) = e 
+untick (Tick _ e) = untick e 
 untick e          = e 
 
 tidyTuples :: RewriteRule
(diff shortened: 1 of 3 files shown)


src/Language/Haskell/Liquid/Transforms/Rewrite.hs
```diff
@@ -45,13 +45,64 @@ import qualified Data.HashMap.Strict as M
...
+     | App e1 a  <- untick e
+     , App e2 f  <- untick e1
+     , App e3 t3 <- untick e2 
+     , Type _    <- untick t3 
+     , App e4 t2 <- untick e3 
+     , Type _    <- untick t2 
+     , App d t1  <- untick e4 
+     , Type _    <- untick t1
+     , Var v     <- untick d 
+     , show v == "GHC.Base.$"
```
facundominguez: Here's how GHC seems to do it:
https://gitlab.haskell.org/ghc/ghc/-/blob/3c0e379322965aa87b14923f6d8e1ef5cd677925/compiler/GHC/Rename/Expr.hs#L2448

I'm guessing that must be faster than calling show on every variable.

nikivazou: Thanks!!!! Changed it!

facundominguez:
Added some edits after playing with this. Feel free to discard them. But I think you do want the new test run in CI, for which it needs to be added to the `tests.cabal` file.

nikivazou:
Looks great, thanks! (I did not know I need to add the cabal file.)

PR #2110, opened 2022-11-29
Add a test verifying the laziness requirement of Data.List.permutations

facundominguez:


ranjitjhala:
nice! lets merge as soon as CI says ok!

PR #2111, opened 2022-11-29
Fix some typos / dead link in blog post

gruhn:
The blog posts on the website are great. Interactive and very well explained :+1: I fixed some typos in [Refinement Types 101](https://ucsd-progsys.github.io/liquidhaskell/blogposts/2013-01-01-refinement-types-101.lhs/) and [Refinements 101 (contd.)](https://ucsd-progsys.github.io/liquidhaskell/blogposts/2013-01-27-refinements101-reax.lhs/).

### DIFF
--- a/docs/blog/2013-01-27-refinements101-reax.lhs
+++ b/docs/blog/2013-01-27-refinements101-reax.lhs
@@ -155,7 +155,7 @@ non-negative `n` guarantee holds trivially.
 **Reason 2: The Specification is a Fib**
 
 If you run the above in the demo, you will see that LiquidHaskell still
-doth protest loudly, and frankly, one might start getting a little
+does protest loudly, and frankly, one might start getting a little
 frustrated at the stubbornness and petulance of the checker.
 
 \begin{code} However, if you stare at the implementation, you will see that it in fact, *does not* meet the specification, as
@@ -194,7 +194,7 @@ recursive calls --- we get the above by plugging the parameters
 \end{code}
 
 \begin{code} Finally, to check the output guarantee is met, LiquidHaskell asks the SMT solver to prove that
-(b >= 2n - 2)  =>  (b >= n)
+(b >= 2n - 3)  =>  (b >= n)
 \end{code}
 
 The SMT solver will refuse, of course, since the above implication is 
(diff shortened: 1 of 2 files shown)


docs/blog/2013-01-27-refinements101-reax.lhs
```diff
@@ -194,7 +194,7 @@ recursive calls --- we get the above by plugging the parameters
 \end{code}
 
 \begin{code} Finally, to check the output guarantee is met, LiquidHaskell asks the SMT solver to prove that
-(b >= 2n - 2)  =>  (b >= n)
+(b >= 2n - 3)  =>  (b >= n)
```
gruhn, line 197: This should be 3 instead of 2 right? In line 193 it was written with a 3 and also the paragraph below isn't true anymore, if we write a 2 here:

> The SMT solver will refuse, of course, since the above implication is 
> *not valid* (e.g. when `n` is `2`) Thus, via SMT, LiquidHaskell proclaims
> that the function `fib'` does not implement the advertised type and hence
> marks it *unsafe*.

docs/blog/2013-01-27-refinements101-reax.lhs
```diff
@@ -258,5 +258,5 @@ There are several things to take away.
 [concolic]: http://en.wikipedia.org/wiki/Concolic_testing
 [icse04]: http://goto.ucsd.edu/~rjhala/papers/generating_tests_from_counterexamples.html
 [dsd]: http://dl.acm.org/citation.cfm?doid=1348250.1348254
-[mlton]: http://www.cs.purdue.edu/homes/zhu103/pubs/vmcai13.pdf
+[mlton]: https://www.cs.purdue.edu/homes/suresh/papers/vmcai13.pdf
```
gruhn, line 261: The link is dead:

http://www.cs.purdue.edu/homes/zhu103/pubs/vmcai13.pdf

I googled and tried to find an active link. I found this:

https://www.cs.purdue.edu/homes/suresh/papers/vmcai13.pdf

Is that the right paper? Otherwise let me know and I'll adjust this commit.

docs/blog/2013-01-27-refinements101-reax.lhs
ranjitjhala: Thanks for catching that!!!

docs/blog/2013-01-27-refinements101-reax.lhs
ranjitjhala: perfect, thank you!

ranjitjhala:
Thank you very much @gruhn !

PR #2115, opened 2022-12-05
Reproducible Nixpkgs

### LINKED ISSUE #2114: Nixpkgs version is not pinned
When building with Stack, users can optionally ask GHC to be sourced from Nixpkgs. The problem is that the version of Nixpkgs is not pinned. Depending on the version of Nixpkgs used by default in the user's environment, the build may fail as a result, say because the desired GHC version is not available in the user's particular version of Nixpkgs.

The solution is to pin Nixpkgs to a specific version. This makes the build more reproducible.

mboes:
Pin the Nixpkgs version when using `stack --nix`.

Closes #2114

### DIFF
--- a/nixpkgs.nix
+++ b/nixpkgs.nix
@@ -1,0 +1,9 @@
+let
+  # NixOS/Nixpkgs master 2022-11-27
+  rev = "a115bb9bd56831941be3776c8a94005867f316a7";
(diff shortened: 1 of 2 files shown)


ranjitjhala:
Thanks! I’ll merge asap!

PR #2118, opened 2022-12-08
Move to LTS-19.33

mboes:
- benchmark-timings: Loosen bound on aeson
- Loosen bounds for compatibility with GHC 9.0.2
- Move to LTS-19.33

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -11,7 +11,6 @@ ghc-options:
 packages:
 - liquid-fixpoint
 - liquid-ghc-prim
-- liquid-base
 - liquid-bytestring
 - liquid-prelude
 - liquid-vector
@@ -31,6 +30,8 @@ extra-deps:
   commit: 45ef3498e35897712bde8e002ce18df6d55f8b15
 # for tests
 - strip-ansi-escape-0.1.0.0@sha256:08f2ed93b16086a837ec46eab7ce8d27cf39d47783caaeb818878ea33c2ff75f,1628
+- git: https://github.com/facundominguez/liquid-base
+  commit: 8ad2378cee5ccf7937d9e08aacd5c5b7128318e8
 
 resolver: lts-19.33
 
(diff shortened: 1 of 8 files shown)


stack.yaml
```diff
@@ -31,6 +30,8 @@ extra-deps:
   commit: 45ef3498e35897712bde8e002ce18df6d55f8b15
 # for tests
 - strip-ansi-escape-0.1.0.0@sha256:08f2ed93b16086a837ec46eab7ce8d27cf39d47783caaeb818878ea33c2ff75f,1628
+- git: https://github.com/facundominguez/liquid-base
+  commit: 8ad2378cee5ccf7937d9e08aacd5c5b7128318e8
```
facundominguez, line 34: This could be
```
- git: https://github.com/liquidhaskell/liquid-base
  commit: 3a006d906f47c64c5a4c33ee698434a56bf612e3
```
which is the result of merging the patches in https://github.com/facundominguez/liquid-base

stack.yaml
```diff
@@ -34,6 +34,7 @@ extra-deps:
   commit: 8ad2378cee5ccf7937d9e08aacd5c5b7128318e8
 
 resolver: lts-19.33
+allow-newer: true
```
facundominguez, line 37: Is this change to be merged?

nikivazou:
Shall we merge this? @mboes can you apply Facundo's suggestions?

stack.yaml
mboes: This particular commit hash was chosen to be the same as in `cabal.ghc9.project`.

stack.yaml
mboes: Yes. Temporary due to bounds in liquid-base, which we either patch there or we live with allow-newer until my next PR, which is just waiting for this one to be merged, and which reverts this particular line.

PR #2120, opened 2022-12-14
Make ExactDC default

nikivazou:
This can create errors when fancy types are defined (e.g., the worker and wrapper constructor do not match) in which case LH suggests using the `no-exact-data-con` flag.

### DIFF
--- a/docs/blog/2016-10-06-structural-induction.lhs
+++ b/docs/blog/2016-10-06-structural-induction.lhs
@@ -108,7 +108,7 @@ Reflecting Lists into the Logic
 To talk about lists in the logic, we use the annotation
 
 \begin{code}
-{-@ LIQUID "--exact-data-cons" @-}
+{- LIQUID "--exact-data-cons" @-}
 \end{code}
 
 which **automatically** derives measures for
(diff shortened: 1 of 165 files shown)


docs/blog/2016-10-06-structural-induction.lhs
```diff
@@ -108,7 +108,7 @@ Reflecting Lists into the Logic
 To talk about lists in the logic, we use the annotation
 
 \begin{code}
-{-@ LIQUID "--exact-data-cons" @-}
+{- LIQUID "--exact-data-cons" @-}
```
facundominguez, line 111: Now that this is the default, perhaps it needs not be mentioned anymore in the post.

facundominguez (commented):
It looks good to me.

Eventually, we should have a bit of an explanation of what `--no-exact-data-cons` is disabling in `docs/mkDocs/docs/options.md`, and when to use it.

facundominguez:
@nikivazou, would you like finishing this PR?

facundominguez:
Hello @nikivazou, do you have any context to share on the decision to close this one?

nikivazou:
Yes, I closed it because I do not believe I will any time soon find time to fix the failing tests...

PR #2128, opened 2022-12-23
Drop support for GHC <9

### LINKED ISSUE #2116: Consider only supporting a single GHC version
I didn't find a GHC support policy documented anywhere. I propose changing it as follows, and documenting it.

Liquid Haskell claims to support GHC versions 8.6 through to 9.0. The cost of this support is many pre-processor directives and much code duplication in the thin compatibility layer found in `src-ghc/Liquid/GHC/API.hs` but also throughout the codebase. This makes the code harder to read, harder to maintain, and complicates porting GHC to new versions. Adding support for newer GHC versions would be easier if there no longer was a need to adapt the code in ways that minimize the amount of code duplication necessary to continue supporting older versions.

_One way to fix this problem would be to support only one GHC version at any given time._ The upside is more maintainable code and easier porting, and for users marginally more features and big fixes shipped since Liquid Haskell developer time is a limited resource. The downside is Liquid Haskell users stuck on old GHC versions don't get the benefit of bug fixes and new features, unless these get backported to previous versions of Liquid Haskell.

Keep in mind that Liquid Haskell relies heavily on the GHC API throughout its codebase. The API itself provides no backwards compatibility guarantees, in practice changes fairly drastically from version to version, and comes with no release notes explaining the changes and how to adapt to them. In short, there is no release management around the GHC API. The API exposes the internal guts of GHC to any external consumer, who uses them the way any GHC-internal module would. In effect, heavy users of a large cross-section of the GHC API, like Liquid Haskell does, are effectively part of GHC itself. Right now the code happens to live in a repo separate to GHC's, with its own release cadence. If the repos were merged, then newer Liquid Haskell would only support newer GHC anyway - as in the above proposed policy.

mboes:
As argued in #2116, supporting multiple GHC versions is onerous and becoming increasingly so. This is because the extended GHC API, which LH is intimately tied to, is not a stable API suitable for public consumption (no stability guarantees, no deprecation periods, no changelog...). This PR introduces the one-ghc-version policy, documented in the README, and removes support for all GHC versions except 9.0, thereby eliminating nearly all uses of CPP.

With this PR, there is now less code, CI is simpler and is slightly faster. The upcoming GHC 9.2 support builds on this PR. Basing that work on this PR reduced the effort required, because we can now adapt LH to newer versions of the GHC API without worrying about backwards compatibility with older versions of GHC.

Closes #2116


ranjitjhala:
Amazing let’s do this, thank you @mboes !!!

README.md
```diff
@@ -300,10 +300,22 @@ Bash script. The script doesn't accept any argument and it tries to determine th
...
+
+LH supports only one version of GHC at any given time. This is because LH depends heavily on the `ghc` library
+and there is currently no distinction between public API's and API's internal to GHC. There are currently no
+release notes for the `ghc` library and breaking changes happen without notice and without deprecation
+periods. Supporting only one GHC version saves developer time because it obviates the need for `#ifdef`'s
+throughout the codebase, or for an compatibility layer that becomes increasingly difficult to write as we
+attempt to support more GHC versions. Porting to newer GHC versions takes less time, the code is easier to
+read and there is less code duplication.
+
+Users of older versions of GHC can still use older versions of LH.
```
facundominguez, line 313: @ranjitjhala, @nikivazou, does this policy work for you?

ranjitjhala:
Oops sorry I jumped the gun and merged policy works for me, I agree we
should simplify! @niki I hope ok with you too? (Else we can revert?)

PR #2129, opened 2022-12-26
Move to GHC 9.2

mboes:
Move to LTS 20, which uses GHC 9.2. This required making a number of changes to liquid-ghc-prim, since upstream changed significantly.

Adapting the codebase of the liquidhaskell package was difficult. The following breaking changes happened in the GHC API:
- Some modules were further split up, after the module reorganization that occurred in 9.0.
- Some data types and functions moved to different modules or were removed altogether.
- The error datatypes changed definition.
- Locations and AST annotations work differently in GHC 9.2.
- The representation of case alternatives in Core has changed.
- Some state kept in `DynFlags` has moved to `HscEnv`.
- Logging requires a logger argument, which we thread through a number of functions.
- Field names are handled differently due to `-XDuplicateRecordFields`.
- `elabRnExpr` is based on on `tcRnExpr` upsteam, which has been significantly refactored as part of the deep subsumption work.
- `deepSplitProductType_maybe` was removed.
- `unsafeGlobalDynFlags` was removed.


mboes:
This PR is in draft stage for two reasons:
- spec checking in liquid-ghc-prim raises multiple errors and I don't know why or how to fix that.
- `unsafeGlobalDynFlags` was removed. The existence of that function was a hack. I was able to stop using that hack in most places, except in `src/Language/Haskell/Liquid/Synthesize/Check.hs`, where I replaced a call to that function with `undefined` for now. This needs a proper fix, but I don't know how.

nikivazou:
Thanks!!!! (Sorry, just got back from XMas holidays....). 
So, the replacement of `unsafeGlobalDynFlags` with `error` is what is cousing the circleci failures? 
I will take a look this week.

src-ghc/Liquid/GHC/GhcMonadLike.hs
```diff
@@ -294,14 +289,27 @@ lookupModule mod_name Nothing = do
...
+          throwOneError $ noModError hsc_env noSrcSpan mod_name err
+
+-- | Abstraction of 'EpaComment'.
+data ApiComment
+  = ApiLineComment String
+  | ApiBlockComment String
+
+-- | Extract top-level comments from a module.
+apiComments :: ParsedModule -> [Ghc.Located ApiComment]
+apiComments pm =
```
facundominguez, line 309: The behavior of this function has changed with the port.

For a module like
```
module A where

f :: Int
f = 1

{-@ ignore g @-}
g :: Int
g =2
```
before it would yield `{-@ ignore g @-}` but now it skips the comment.

facundominguez: I figured this one. A way to find missing comments is to use `ghc-exactprint` to print the AST with [showAst :: (Data a) => a -> String](https://hackage.haskell.org/package/ghc-exactprint-1.6.1.1/docs/Language-Haskell-GHC-ExactPrint.html#g:6) (it might be possible to do this with the GHC API alone with GHC.Hs.Dump.showAstData as well).

This shows the whole tree, and in particular the node that holds the `{-@ ignore ... @-}` comment.

Now, I'm expecting this problem to come back when we depend on some comment attached to a function in a `where` clause. But I'm letting it come for now.

liquid-prelude/src/Language/Haskell/Liquid/Bag.hs
```diff
@@ -44,7 +44,7 @@ put k m = M.insert k (1 + get k m) m
 union :: (Ord k) => Bag k -> Bag k -> Bag k
 union m1 m2 = M.union m1 m2
 
-{-@ thm_emp :: x:k -> xs:Bag k ->  { Language.Haskell.Liquid.Bag.empty /= put x xs }  @-}
+{-@ thm_emp :: x:k -> xs:Bag k ->  { Map_select xs x >= 0 => Language.Haskell.Liquid.Bag.empty /= put x xs }  @-}
```
facundominguez, line 47: @ranjitjhala, I believe this extra hypothesis is needed, right? Otherwise, how could the SMT solver see that `put x xs` doesn't turn an entry from `-1` to `0`.

I would have expected the missing hypothesis to be noted before the port, though.

facundominguez:
In porting LH to ghc-9.2.5, there is at least one test that I don't know how to fix:  [tests/pos/FractionalInstance.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/pos/FractionalInstance.hs).

It produces a default method like the following for Fractional (Frac a) 
```$crecip = \ (eta :: Frac a) -> (/) FractionalInstance.$fFractionalFrac 1 eta```
where it used to be
```$crecip = GHC.Real.$dmrecip FractionalInstance.$fFractionalFrac```

The former now produces a verification failure
```
pos/FractionalInstance.hs:18:10: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : {v : (FractionalInstance.Frac a) | v == eta}
    .
      VV : {VV : (FractionalInstance.Frac a) | VV /= 0}
    .
    in the context
      eta : (FractionalInstance.Frac a)
    Constraint id 13
   |
18 | instance Fractional (Frac a) where
   |          ^^^^^^^^^^^^^^^^^^^
```
The issue seems to be that the default method doesn't get a refinement type establishing that `eta` is non-zero. Maybe because `--typeclass` is not in effect.

An option could be to drop derived bindings (those with names starting with $) when `--typeclass` is not in effect. Would this be acceptable? Note that this would disable verification in typeclass instances whose methods have been written explicitly. It doesn't look like GHC provides a way to distinguish between methods provided by default and those given by the programmer.

I'm happy to hear about other ideas.

**Comment by facundominguez** [comment]
I suppressed the above failure by adding Fractional to the list of classes whose instances aren't verified in f2554c7.
The instances of this class can be verified again by using the `--check-derived` flag.

**Comment by facundominguez** [comment]
This is ready to merge!

**Comment by ranjitjhala** [comment]
🎉

**Comment by nikivazou** [comment]
Amazing!!!! Thanks!!!
@facundominguez the idea is these edits only support GHC 9.2, so we should tag/release the current develop as the last GHC 8.* version before we merge?

File: liquid-prelude/src/Language/Haskell/Liquid/Bag.hs (hunk 2 above)
Comment (nikivazou) [reply liquid-prelude/src/Language/Haskell/Liquid/Bag.hs:47]: I do not see that... Put always returns a non empty bag and this is what the theorem says. 
The domain of put's entry could be express in a different theorem.

File: liquidhaskell.cabal [hunk 3]
```diff
@@ -195,7 +195,7 @@ library
                     , aeson
```
nikivazou, line 195: @facundominguez @ranjitjhala we should change the version to `version: 0.9.2.0.0`? The convention is that the first two numbers follow GHC's version

facundominguez:
> so we should tag/release the current develop as the last GHC 8.* version before we merge?

Up to you. The history of all commits stays available, so releasing is possible both before and after merging.

liquid-prelude/src/Language/Haskell/Liquid/Bag.hs
facundominguez: > Put always returns a non empty bag and this is what the theorem says.

In the representation, an empty Map has all entries set to 0.

Does `put 1 m` produce a map with some entry greater than 0? It depends on what `m` is.

What if I give you the map `Map_store (Map_default 0) 1 (-1)`? It looks like in this case `put 1 m` would produce a map where all entries are 0.

We know, this implementation has an invariant where all maps representing a bag only stores non-negative values. But this invariant is not expressed in this specification of bags.

Does that explain it?

liquidhaskell.cabal
facundominguez: In 754379903 I set the version to `0.9.2.5.0`.

facundominguez:
> so we should tag/release the current develop as the last GHC 8.* version before we merge?

Note that the `origin/develop` branch is already building only with `ghc-9.0.2` since https://github.com/ucsd-progsys/liquidhaskell/pull/2128.

PR #2136, opened 2023-01-16
Move to GHC 9.2 with more fixes

facundominguez:
Adds some fixes on top of #2129.

Except for the fix to the CI configuration, none of the fixes here address fundamental causes. I only offer them to point at things that maybe aren't working as they should after porting the ghc-9.2. In other words, I don't know why these changes are necessary after the port, but I can see that building liquid-* libraries fails in various ways without them.


facundominguez:
The next failure in CI is because LH is trying to verify functions maked as `{-@ ignore bla @-}`. e.g.

```Haskell
{-@ ignore liquidAssume @-}
{-@ assume liquidAssume :: b:Bool -> a -> {v: a | b}  @-}
{-# NOINLINE liquidAssume #-}
liquidAssume :: Bool -> a -> a
liquidAssume b x = if b then x else error "liquidAssume fails"
```

```
src/Language/Haskell/Liquid/Prelude.hs:81:43: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : {v : [GHC.Types.Char] | v == "liquidAssume fails"
                                   && len v == strLen "liquidAssume fails"
                                   && len v >= 0
                                   && v ~~ "liquidAssume fails"}
    .
      VV : {VV : [GHC.Types.Char] | false}
    .
    Constraint id 111
   |
81 | liquidAssume b x = if b then x else error "liquidAssume fails"
   |                                           ^^^^^^^^^^^^^^^^^^^^
```

facundominguez:
Merged these to #2129.

PR #2142, opened 2023-02-03
liquid-fixpoint: use smtlib-backends

qaristote:
The purpose of this PR is to test the integration of smtlib-backends into liquid-fixpoint. See https://github.com/ucsd-progsys/liquid-fixpoint/pull/641 for more details.


facundominguez:
Thanks @qaristote. I'm updating LF in #2149.

PR #2143, opened 2023-02-04
Remove legacy executable

facundominguez:
We are not testing it in CI anymore, and there are no plans to improve it either. Plus leaving it in the code and the documentation has the potential to confuse newcomers.


facundominguez:
I remember the ability to profile LH being a concern to retain the legacy executable. Apparently, it is possible to invoke the plugin with a driver compiled for profiling without having to recompile GHC itself: https://mpickering.github.io/posts/2020-03-26-tip-for-profiling.html

ranjitjhala:
Nice — we should add this to the docs so I can find it the next time.

facundominguez:
I added docs for profiling in fedd171

PR #2153, opened 2023-02-17
Fix some broken external links as well in the docs

k4rtik:
I couldn't figure out where 'Prelude.hquals' is now:

```
liquidhaskell/docs/mkDocs/docs/specifications.md:
[✖] https://github.com/ucsd-progsys/liquidhaskell/blob/develop/include/Prelude.hquals
```


ranjitjhala:
Fantastic, thanks so much!!!

Issue #2106, opened 2022-11-17
Proposal: get rid of wrapper packages


facundominguez:
### Problem

Verifying a program with liquid haskell requires one to state the assumptions about how dependencies work. The notable exception is when a dependency is also verified with LH, in which case it will contain specifications that can be used to verify the program.

The current mechanism to manage assumptions is to create a Haskell package that states the assumptions (e.g. `liquid-base`) and then have the program use it instead of the actual package (e.g. `base`). In order to state the assumptions, Liquid Haskell requires the `liquid-base` to redefine all the modules of `base` on which assumptions are needed and reexport all of the other modules. This is a hassle because users need to:

 1) change the cabal configuration to replace dependencies by others which are only needed for verification.
 2) create wrapper `liquid-*` packages for every dependency they intend to make assumptions about.
 3) keep the wrapper packages up-to-date as new releases of the dependencies pop up.

### A solution

I propose that we introduce some changes to LH so we don't need wrapper packages anymore.

A first change is to give LH a mechanism to state reusable assumptions about dependencies without having to redefine Haskell modules only for that sake. In this way, if we want to throw all of the assumptions about a dependency into a single module, we can do it, and moreover, we can do it without having to change the dependency of the program. Making the assumptions reusable means that we can somehow point LH to the module containing the assumptions, and LH will be able to use them in other places. 

A second change is to pack inside the LH plugin all of the essential assumptions made for the boot libraries. These libraries depend on the GHC version used, and there is no reason to ask the user to manage the assumptions about boot libraries separately from enabling the LH plugin.

A last change is to allow to configure the LH plugin so it knows where to find the assumptions about dependencies when they are not boot libraries. There are a few options here:
 * Use a directive that brings the assumptions from a module and its imports: e.g. `{-@ import My.Liquid.Assumptions @-}` 
 * Have LH implicitly load assumptions from a module like `Liquid_Dependency_Assumptions` and its imports.
 * Define some overrideable convention so LH knows to look for assumptions for functions in module `M` in some other module `Liquid.Assumptions.M`.

It would be helpful to collect others thoughts on this matter as I'm finding tricky to navigate the design space. Also, I'm not too knowledgeable on how the current LH works, so perhaps there are already solutions which are similar to the proposed changes.

facundominguez:
@kosmikus @adinapoli if you have any thoughts, they will be appreciated.

ranjitjhala:
[...]

PR #2160, opened 2023-03-21
Remove liquid-ghc-prim

facundominguez:
This PR removes the package `liquid-ghc-prim`, but it keeps the spec files in `include/known-packages/ghc-prim`.

There are spec files for three modules `GHC.Types`, `GHC.Classes`, and `GHC.CString`. Every time LH verifies a module that imports any of those files, the specs are loaded.

This is part of the implementation for #2106.


facundominguez:
Almost all modules from `liquid-base` build with this change. The only exception is `GHC.Pack`:
```
[214 of 214] Compiling GHC.Pack         ( src/GHC/Pack.hs, dist-newstyle/build/x86_64-linux/ghc-9.2.5/liquid-base-4.15.1.0/build/GHC/Pack.o, dist-newstyle/build/x86_64-linux/ghc-9.2.5/liquid-base-4.15.1.0/build/GHC/Pack.dyn_o )

include/known-packages/ghc-prim/GHC.CString.spec:8:6: error:
    • Illegal type specification for `GHC.CString.unpackCString#`
    GHC.CString.unpackCString# :: x:GHC.Prim.Addr# -> {v : [GHC.Types.Char] | len v == strLen x
                                                                              && v ~~ x}
    Sort Error in Refinement: {v : [Char] | len v == strLen x
                                            && v ~~ x}
    Unbound symbol len --- perhaps you meant: v, x, head ?
    Just assume
    • 
  |
8 |   :: x:GHC.Prim.Addr#
  |      ^^^^^^^^^^^^^^^^...
Error: cabal: Failed to build liquid-base-4.15.1.0 (which is required by
exe:liquidhaskell from liquid-platform-0.9.0.2).
```

`GHC.CString.spec` is importing `GHC.Types`, `GHC.Types.spec` does define `len`, and the code goes through the trouble of collecting the specs of import declarations in spec files. I don't understand what is missing here, and I was hoping that  @nikivazou could have a clue if I'm forgetting something obvious. [This](https://github.com/ucsd-progsys/liquidhaskell/pull/2160/files#diff-035b1af21b30ba286d42b9e59267424cc7862562b8c2385838214f5c893b39b1R88) is the new code that loads the spec files.

`GHC/Pack.hs` contains
```Haskell
module GHC.Pack (module Exports) where

import "base" GHC.Pack as Exports
```
if I add an import there like `import GHC.Types` or `import GHC.Base`, compilation succeeds.

nikivazou:
It is possible that there is an ordering problem? meaning that the type of `GHC.CString.unpackCString# ` is checked in an environment where `len` (and generally `GHC.Types.spec`) is not imported *yet*?

facundominguez:
I've been playing around some more with the implementation, but so far I have no news.

> It is possible that there is an ordering problem?

Since `GHC.CString.spec` is loaded into a `LiquidLib` and the `LiquidLib` contains `GHC.Types.spec` in the dependencies, an ordering problem would be possible if there is some bug in the pipeline. I'll try understanding the data flow better. The failing stack trace is
```
Plugin.processModule
Bare.makeTargetSpec
Bare.makeGhcSpec
Bare.Check.checkTargetSpec
Bare.Check.checkBind allowHO bsc "assume"       emb tcEnv env
Bare.Check.checkTy
Bare.Check.checkRType
probably Bare.Check.checkReft
```

facundominguez:
I think I figured this one. When checking the spec, LH complained that `len` was not in scope. The implementation is surprisingly not putting `len` in scope because its type is referring to type constructors (`GHC.Types.Int`) that are not in scope, and measures are dropped in general in such a case (!). After figuring that, it was fixable by adding the missing type constructors to the environment that LH uses.

Does dropping measures like that serve a purpose? I would rather expect an error about the missing type constructors instead.

facundominguez:
I'm working currently on removing `liquid-base`. Because I've fixed since then some things that were not ideal in this PR, I'll merge it only after getting all the tests passing after removing `liquid-base` as well.

PR #2163, opened 2023-04-03
Relational checking with Translation (WIP)

oquechy:


liquidhaskell.cabal
```diff
@@ -160,6 +161,7 @@ library
                     , mtl                  >= 2.1
                     , optics               >= 0.2
                     , optparse-applicative < 0.18
+                    , ormolu               < 0.5.1
```
facundominguez, line 164: :wave: I had a brief exchange with @oquechy in private about this `ormolu` dependency. IIUC, this pr gives LH the ability to produce code that is then formatted with `ormolu`.

Because `ormolu` brings many other transitive dependencies, I think it would be worth exploring an approach that doesn't use `ormolu` to format the code, that is, generating it with modest indentation and line breaks to make it readable enough.

Besides, growing build times, adding dependencies is a concern if LH is ever going to be in `head.hackage`.

oquechy: Agreed, I will close this pull request for now, until I refactor this code without `ormolu` 👍

PR #2167, opened 2023-04-18
Use LHAssumptions for the other liquid-* packages

facundominguez:
This PR moves assumptions from `liquid-bytestring` and `liquid-containers` to `liquidhaskell`.

It also rewrites `liquid-vector` and `liquid-parallel` so they offer LHAssumptions modules instead of reexporting everything from `vector` and `parallel`.


facundominguez:
Builds atop https://github.com/ucsd-progsys/liquidhaskell/pull/2166

PR #2170, opened 2023-04-27
Update documentation after removal of liquid-base

facundominguez:
Belongs to the effort of #2106.


facundominguez:
I'm leaving some references in the blog posts that are in the `docs` folder. I think fixing those will require adding a note linking to a post explaining how specs in dependencies are done now. I would be happy to write such a post when I get back from vacations.

ranjitjhala:
Super, thanks so much @facundominguez !!!

PR #2171, opened 2023-04-28
remove top level k vars from first arguments of mutli argument functions

nikivazou:
because all the properties should be captured by the last argument. 
This should lead to generation of fewer constraints!


ranjitjhala:
hmm i am a bit skeptical that this will work ... because of how the qualifiers are setup.

nikivazou:
Yeap, it does not... Will leave it a little more open to think again if there is something smart to be done.

PR #2172, opened 2023-05-09
WIP: RImpF removal

clayrat:

PR #2173, opened 2023-05-19
Remove implicit types

clayrat:
This removes all mechanisms related to implicit types, including the `~>` syntax and corresponding tests. Here's a performance chart generated on benchmark test subset: 
perf

I'll add more charts detailing top differences on the full test suite.


facundominguez:
Thanks @clayrat. Is this motivated by the discussion of https://github.com/ucsd-progsys/liquidhaskell/issues/2133 ?

clayrat:
> Thanks @clayrat. Is this motivated by the discussion of #2133 ?

This was requested by @nikivazou as a continuation of the refactoring started in #2172. As I understand, the feature isn't fully functional and there's no dedicated researcher/developer working on it for the foreseeable future.

facundominguez:
I see. In that case, I think I would need some help to write my types without implicit types, I'm discussing the problem [here](https://github.com/ucsd-progsys/liquidhaskell/issues/2133#issuecomment-1523329795).

clayrat:
Updated bench chart:

perf-bench

Top 50 speedups (+ removed tests):

hi

Top 50 slowdowns:

lo

nikivazou (approved):

PR #2176, opened 2023-06-02
Add upper bound on ansi-wl-pprint for stitch-lh

clayrat:
@facundominguez It seems that version 1.0.2 of `ansi-wl-pprint` released on 2023-05-18 now breaks `stitch-lh`, a benchmark in our test suite. I propose adding this upper bound as a temporary workaround, but I think we should eventually switch to `prettyprinter` as the deprecation warning suggests.


facundominguez:
Thanks @clayrat!

> I think we should eventually switch to prettyprinter as the deprecation warning suggests.

Sounds good to me.

PR #2178, opened 2023-06-20
remove decrease annotation

clayrat:


facundominguez:
:wave: I'm just learning that there is a `decrease` directive. Could we have some context for the record on what it does and why it is removed?

clayrat:
As far as I understand, it's an old way of specifying termination metrics, you have to add a number of the argument that is decreasing, or a list of them if you want to do a lexicographic metric. It's superseded by the `/ [foo,bar]` syntax, however there's a small caveat that the latter needs numerical metrics, whereas `decrease` can be structural; so if you have `f : a -> xs:[a] -> b` you can just write `decrease f 2`, but with the other syntax it has to be `/ [len xs]`.

nikivazou:
Exactly, it was the first (and easiest) attempt to implement the termination checks.

ranjitjhala:
@facundominguez -- here is a quite old description from an early blog post on termination

https://github.com/ucsd-progsys/liquidhaskell/blob/cca0b21e848a6f241941628f19dc49611eace815/docs/blog/todo/termination.lhs#L97-L115

as you can see this was subsumed by the "decreasing expressions" -- so

```
{-@ Decrease sum' 3 @-}
{-@ sum' :: Vec -> Val -> Nat -> Val @-}
```

is really the same as

```
{-@ sum' :: Vec -> Val -> i:Nat -> Val / [i] @-}
```

clayrat:
Benchmark diffs:

filtered

Top 50 slowdowns:

bot

Top 50 speedups:

top

clayrat:
@nikivazou I've added the benchmarks

clayrat:
I think most of it is random-ish, except for the top 2 slowdowns & top 2 speedups - these 4 benchmarks received substantial changes in annotations.

nikivazou (approved):

PR #2180, opened 2023-06-20
Remove more dead code only used in unbuildable executables

facundominguez:
This PR removes `liquid-platform/src/Gradual.hs` and `liquid-platform/src/Target.hs`, programs that were marked as non-buildable in the cabal file and were not built in CI for at least the last 3 years.

Removing these files allows us to remove quite a bit of code related to the GHC API in `liquidhaskell-boot`.

Please, let me know if any of it is meant to stay.


nikivazou (approved):

nikivazou:
Amazing! Thanks! (For the record, I recently saw some gradual dead code in the fixpoint too that I thought it would be great if removed :))

PR #2208, opened 2023-08-03
Fix link from ICFP'15

shingarov:
Commit a826d42 moves `benchmarks/` to `tests/benchmarks/`. Unfortunately this has the unintended consequence of breaking "Reference 23" on p.12 of Vazou–Bakst–Jhala's ICFP'15 paper "Bounded Refinement Types": that reference is now a 404.

This is of course not to blame commit a826d42 itself, as the problem goes all the way back to Berners-Lee's original design assuming an immutable Web; in other words we shouldn't be prohibited from making any change whatsoever just because there might exist some random link from some random outside resource.

However, in *this* case, ICFP'15 is NOT "some random outside resource". And we can't fix it retroactively to reflect the post-a826d42 codebase. Therefore, it would seem like the only solution is to fix the codebase instead so that the link works again.

### DIFF
--- a/benchmarks/icfp15/README.md
+++ b/benchmarks/icfp15/README.md
@@ -1,0 +1,4 @@
+# Examples from _Bounded Refinement Types_
+
+The source code of the examples from Vazou–Bakst–Jhala's ICFP'15 paper is now
+in [tests/benchmarks/icfp15](../../tests/benchmarks/icfp15).


facundominguez:
Thanks for the report @shingarov! I think it would suffice to leave a `README.md` file in `benchmarks/icfp15` that redirects to the new location of the source code.

shingarov:
@facundominguez Something like 5302d97 then?

benchmarks/icfp15/README.md
```diff
@@ -0,0 +1,4 @@
+# Examples from _Bounded Refinement Types_
+
+The source code of the examples from Vazou–Bakst–Jhala's ICFP'15 paper is now
+in `tests/benchmarks/icfp15`.
```
facundominguez: This works! Probably there is some way to make it a link.
```
[tests/benchmarks/icfp15](../../tests/benchmarks/icfp15)
```
or
```
[tests/benchmarks/icfp15](https://github.com/ucsd-progsys/liquidhaskell/tree/develop/tests/benchmarks/icfp15)
```

facundominguez (approved):

benchmarks/icfp15/README.md
shingarov: Let it be so.  I am definitely going for the first one (relative path) because if one already is working on a local copy, the link should still point to within the local copy.

PR #2211, opened 2023-08-05
Strengthen the lookup of assumption identifiers

facundominguez:
The needed TyThing may not be the first found, and this causes LH to fail on some GHCs but not in others.


clayrat:
Wow, thanks!

PR #2212, opened 2023-08-06
Some simplifications to plot-performance

facundominguez:


facundominguez:
@clayrat, please, let me know what you think. None of these commits should change the behavior of the code. I think they should be easier to review one commit at a time.

clayrat:
@facundominguez looks nice overall! However, it does change the semantics - the idea for the separate handling of the removed/added tests and the `rev` flag was to have removed tests at the top when showing speedups (and symmetrically, added tests when showing slowdowns), so you could quickly assess which ones are added/removed, as opposed to the tests that didn't run for some reason (I was thinking to highlight these in the future by e.g. changing the color of the label, this was also the rationale behind keeping the `result` field), whereas in the current version this potential distinction is lost:

top

facundominguez:
That's good insight.

> have removed tests at the top when showing speedups

I realize that now these things don't show up at the top, but are mixed with the rest of the results depending on how long the removed benchmark measured.

Rather than displaying the removed and added benchmarks in the plots, maybe we could just issue warnings about them in stderr. What do you think?

clayrat:
My original idea was to try to have as much relevant info as possible in a single place (i.e., the images). In my mind, the workflow behind this tool is that the PR author generates the charts and provides them to the reviewer; if we shift this info to the console, then the author has to do the additional work of copying it manually from the console into the PR message; also the reviewer has to be more attentive not to skip it when reading.

facundominguez:
I added the warnings on the plots.
top

clayrat:
@facundominguez Looks good, although we now duplicate the missing info on both the speedup and the slowdown chart, I guess it can be viewed as more consistent that way. Btw, do you know what's the deal with Appveyor now? The error is always

```
[ 7 of 41] Compiling GHC.Int_LHAssumptions
<no location info>: error:
    src\GHC\.liquid\.coCD3B.tmp: copyFile:atomicCopyFileContents:withReplacementFile:renameFile:renamePath:MoveFileEx "\\\\?\\C:\\projects\\liquidhaskell\\src\\GHC\\.liquid\\.coCD3B.tmp" Just "\\\\?\\C:\\projects\\liquidhaskell\\src\\GHC\\.liquid\\liquid.css": permission denied (Access is denied.)
```

This hasn't happened for #2211.

facundominguez:
I have no insights on the Windows failures. I'm guessing it is some bug in the interaction of stack and Windows.

PR #2215, opened 2023-08-29
Let LH know that Bags only store non-negative numbers

facundominguez:


facundominguez:
I was reviewing [an old discussion](https://github.com/ucsd-progsys/liquidhaskell/pull/2129/commits/b982fcf35519c80d68fae8094b14c6bfa6314bb3#r1088537496) and realized that there was a nicer fix for the Bag specs.

tests/pos/MergeSort_bag.hs
```diff
@@ -9,6 +9,7 @@ module MergeSort_bag (bag, sort) where
 import qualified Language.Haskell.Liquid.Bag as B
 
 {-@ measure bag @-}
+{-@ bag :: Ord a => [a] -> Bag a @-}
```
facundominguez, line 12: It is awkward referring to Bag in unqualified form here. It refers to the LH type synonym, not the Haskell type. If we qualify this, LH doesn't seem to pick up the meaning of the LH type synonym.

facundominguez:
@nikivazou, let me know what you think. The changes here do force users of `Bag` to add specs for functions that use `Bag` in their type signatures.

liquid-prelude/src/Language/Haskell/Liquid/Bag.hs
```diff
@@ -17,6 +17,7 @@ import qualified Data.Map      as M
 @-}
 
 
+{-@ type Bag a = Data.Map.Map a Nat @-}
```
nikivazou, line 20: why not name the LH alias `NatBag`?

nikivazou (approved):

nikivazou:
Users of `Bag` need to add the spec to let LH know that is contains natural numbers correct? 
Would this requirement go away if we give a different name to the Liquid Haskell `Bag` alias, like `NatBag`?

liquid-prelude/src/Language/Haskell/Liquid/Bag.hs
facundominguez: The alias could be renamed. Then `NatBag String` is a bag of `String`s (not Nats). Will that be confusing?

nikivazou: I see... How about `RBag` from Refined Bag?
Just trying to prevent the requirement to have to give refined signatures to each user of `Bag`.

facundominguez:
> Users of Bag need to add the spec to let LH know that is contains natural numbers correct?
> Would this requirement go away if we give a different name to the Liquid Haskell Bag alias, like NatBag?

I think users would still need to add a spec even if the type alias were renamed. The need for the spec streams not from the naming, but from the fact that `put` requires as input a `NatBag a` while `bag xs` returns only a `B.Bag a` when the spec is absent.
```haskell
{-@ bag :: Ord a => [a] -> NatBag a @-}
bag :: (Ord a) => [a] -> B.Bag a
bag []     = B.empty
bag (x:xs) = B.put x (bag xs)
```

nikivazou:
Makes sense. Since we cannot avoid the refinement specifications, then let's merge with the same name!

PR #2244, opened 2023-11-28
Remove strip-ansi-escape-0.1.0.0@rev:2

philderbeast:
Removes the unused `strip-ansi-escape` dependency from the Stack project. 

This is a follow up from https://github.com/ucsd-progsys/liquidhaskell/commit/e16382c0006f6bd205485576847fba16afa5a304. Within that change the explicit import from this package that was removed was:

```haskell
import Data.String.AnsiEscapeCodes.Strip.Text (stripAnsiEscapeCodes)
```

Found when adding `liquidhaskell` as an [Updo example](https://github.com/up-do).

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -27,8 +27,6 @@ extra-deps:
 - smtlib-backends-process-0.3
 - git: https://github.com/qnikst/ghc-timings-report
   commit: 45ef3498e35897712bde8e002ce18df6d55f8b15
-# for tests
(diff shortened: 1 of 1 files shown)


facundominguez:
Thanks @philderbeast!

philderbeast:
Just seen that I missed the lock file change, https://github.com/ucsd-progsys/liquidhaskell/compare/develop...typechecker:liquidhaskell:remove/strip-ansi-escape. Sorry about that.

philderbeast:
GitHub desktop has this glitch where it will only update the changes if focus changes. That's what got me.

PR #2256, opened 2024-01-23
Recommend Z3 when the user doesn't have an opinion

facundominguez:


ranjitjhala (approved):

ranjitjhala:
Thanks @facundominguez -- very good addition!

PR #2280, opened 2024-04-09
Deemphasize LIQUID_DEV_MODE in the README

facundominguez:
The former description would signal to new contributors that it was something desirable or necessary. Here I attempt to describe it in simpler terms and stop encouraging its use.


ranjitjhala:
Makes sense, thanks!

PR #2282, opened 2024-05-14
update to latest liquid-fixpoint

### LINKED ISSUE #2272: Set measures generate erroneous constraints when applied to non-polymorphic datatypes
Typechecking the following program:

```haskell
{-@ LIQUID "--reflection" @-}

module SngBug where

import Data.Set

data Lst a = Emp | Cons a (Lst a)

{-@ measure lstHd @-}
lstHd :: Ord a => Lst a -> Set a
lstHd  Emp       = empty
lstHd (Cons x _) = singleton x

lcons ::  Lst l -> Lst (Lst l)
{-@ lcons :: p: Lst l -> {v:Lst (Lst l) | v = Cons p Emp } @-}
lcons p = Cons p Emp
```

crashes with the error:

`crash: SMTLIB2 respSat = Error "line 3 column 25217: unknown constant smt_set_sng ((SngBug.Lst Int)) declared: (declare-fun smt_set_sng (Int) (Array Int Bool)) "`

The error disappears if `lstHd` is declared `reflect` instead of `measure`. It triggers both with and without PLE.

It looks like the internal `Lst l` type doesn't get converted to `Int` which is what the encoding of sets as `Array Int Bool` expects. We currently think there are two possible solutions for this:

1. Only generate set operations for elements whose type can be encoded as `Int` (simpler solution, will turn a crash into a proper typechecking error)
2. Generate multiple set types for each potential set element type (more complex, but allows for expressing more programs)

clayrat:
Don't merge yet (before https://github.com/ucsd-progsys/liquid-fixpoint/pull/688), the LF submodule currently points to a branch.

Fixes #2272

@ranjitjhala I had to remove two things to make LH work with the `develop` version of LF: 

1. `fTyConP` as a parser with the same name was made public at https://github.com/ucsd-progsys/liquid-fixpoint/pull/683
2. conflicting JSON instances for `SourcePos` which were added at https://github.com/ucsd-progsys/liquid-fixpoint/pull/680

The tests pass without those, but I'm unsure if we should remove them in favor of liquid-fixpoint implementations (as done currently), or if we need to de-duplicate somehow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ commands:
       - checkout
       - add_ssh_keys
       - run: git submodule sync
-      - run: git submodule update --init --remote
+      - run: git submodule update --init
 
   cabal_build_and_test:
     description: "Build the project and run the tests"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ commands:
       - checkout
       - add_ssh_keys
       - run: git submodule sync
-      - run: git submodule update --init
+      - run: git submodule update --init --remote
 
   cabal_build_and_test:
     description: "Build the project and run the tests"
@@ -150,14 +150,14 @@ jobs:
 
   stack_900:
     machine:
-      image: ubuntu-2004:202107-02
+      image: default
     steps:
         - stack_build_and_test:
             stack_yaml_file: "stack.yaml"
             extra_build_flags: "--flag liquidhaskell-boot:devel"
   cabal_900:
     machine:
-      image: ubuntu-2004:202107-02
+      image: default
     steps:
       - cabal_build_and_test:
           ghc_version: "9.8.1"
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,3 @@
 [submodule "liquid-fixpoint"]
 	path = liquid-fixpoint
-	url = https://github.com/clayrat/liquid-fixpoint.git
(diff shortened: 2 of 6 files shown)


ranjitjhala:
@clayrat -- I couldn't follow, I would suggest using the LF versions (which is the current setup?) if possible? Is that the path of least resistance?

clayrat:
> @clayrat -- I couldn't follow, I would suggest using the LF versions (which is the current setup?) if possible? Is that the path of least resistance?

Indeed, the simplest solution is just to delete the conflicting LH code, I was just worried there might be some non-obvious consequences not covered by the test suite, e.g. the code for `fTyConP` in LF [has clauses for `num` and `Str`](https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/src/Language/Fixpoint/Parse.hs#L1031) which are missing in LH version - not sure how important it is that LiquidHaskell will also start accepting those.

nikivazou:
The latest fixpoint is not compatible with the latest liquid haskell, that is why these changes were required. 
It looks good to me (esp since the tests pass). 

So, let's merge both pull requests!

nikivazou:
@clayrat this is WIP only because fixpoint was not fixed yet?

clayrat:
Yes, I'll point the LF submodule to the latest commit and drop the WIP prefix.

clayrat:
@nikivazou done

PR #2284, opened 2024-05-21
update GHC version to 9.8.2

clayrat:
Doesn't add new features, but brings back the HLS support for the devs (9.8.1 is no longer HLS-supported)


facundominguez:
Thanks @clayrat!

PR #2296, opened 2024-06-03
Remove some redundant imports

archbung:
This PR addresses several comments marked by XXX in the stitch-lh test suite. Could you review @facundominguez?


facundominguez:
Looks great. Thanks @archbung!

PR #2297, opened 2024-06-03
fix for coercion

nikivazou:
Strengthen coercions with refinements, fixes https://github.com/ucsd-progsys/liquidhaskell/compare/coercion?expand=1#diff-19569e6ce9a566368341fdf5e65c7da5c8a4abc710b174e0180957af136bc402  that was failing before.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs
@@ -351,7 +351,7 @@ cconsE' γ (Cast e co) t
   = cconsE γ (f e) t
 
 cconsE' γ e@(Cast e' c) t
-  = do t' <- (`strengthen` (uTop (rTypeReft t))) <$> castTy γ (exprType e) e' c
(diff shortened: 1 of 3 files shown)


nikivazou:
@facundominguez any idea why there is an error?
`stack_900` says: 

```
/home/circleci/project/tests/datacon/neg/NewTypes0.hs:1:1: error:
    Could not match the following expected errors with actual thrown errors:
    <Any Liquid error>
  |                
1 | {-@ LIQUID "--expect-any-error" @-}
  | ^              
                   
```

But in cabal and locally in my machine `NewTypes0` fails. Is there any check against expected error message?

facundominguez:
Hello @nikivazou! The test is not finding errors in CI, nor with stack, neither with cabal, but when tried locally it finds the following error:
```
tests/datacon/neg/NewTypes0.hs:7:13: error:
    Bad Data Specification
Mismatch in number of type variables for `Foo`
  |
7 | {-@ newtype Foo a = Bar {x :: Nat} @-}
  |
```

Rerunning the CI jobs gives the same result. At the moment I don't have insights on what could be causing the difference.

nikivazou:
Yes, this is very weird... I am quite confident that the edit does not interfere with this test (since there should not be any coercions in this test). I will just remove this test for now to see what happens.

PR #2299, opened 2024-06-06
Update ProofCombinators.hs

nikivazou:
`==.` is faster than `===` because it is not checking intermediate steps!

### DIFF
--- a/liquid-prelude/src/Language/Haskell/Liquid/ProofCombinators.hs
+++ b/liquid-prelude/src/Language/Haskell/Liquid/ProofCombinators.hs
@@ -153,8 +153,6 @@ infixl 3 ==!
 
 infixl 3 ==.
 
-{-# DEPRECATED (==.) "Use (===) instead" #-}
-
 {-# INLINE (==.) #-} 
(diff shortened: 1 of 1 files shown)


liquid-prelude/src/Language/Haskell/Liquid/ProofCombinators.hs
```diff
@@ -153,8 +153,6 @@ infixl 3 ==!
 
 infixl 3 ==.
 
-{-# DEPRECATED (==.) "Use (===) instead" #-}
-
 {-# INLINE (==.) #-} 
 (==.) :: a -> a -> a
```
facundominguez, line 157: It could be helpful to have a comment in the code explaining the difference between `==.` and `===` and when to use each.

nikivazou: Done!

facundominguez: Thank you!

PR #2302, opened 2024-06-17
WIP update to latest liquid-fixpoint

### LINKED ISSUE #2288: PLE crashes when using Set theory symbols
The following code crashes silently with `Sorry, unexpected panic in liquid-fixpoint!` when either the `assert` is uncommented or the line starting with `===` in the `otherwise` branch is commented out.

```haskell
{-@ LIQUID "--reflection" @-}
{-@ LIQUID "--ple"        @-}

module Test where

import Prelude
import Data.Set as S

{-@ reflect assert @-}
{-@ assert :: b:{Bool | b} -> x:a -> {v:a | v == x && b} @-}
assert :: Bool -> a -> a
assert _ x = x

{-@ reflect cat @-}
cat :: [a] -> [a] -> [a]
cat [] ys = ys
cat (x : xs) ys = x : cat xs ys

{-@ lemma :: (Ord a, Eq a)
              => xs:[a] -> ys:[a]
              -> z:{a | S.member z (S.fromList xs) }
              -> { S.member z (S.fromList (cat xs ys)) }
              @-}
lemma :: (Ord a, Eq a) => [a] -> [a] -> a -> ()
lemma [] ys z = ()
lemma (x : xs) ys z
  | z == x =
    ---assert (S.member z (S.fromList (cat (x : xs) ys)))
              ()
  | otherwise =
    S.member z (S.fromList (cat (x : xs) ys))
    === S.member z (S.fromList (x : cat xs ys))
    ? lemma xs ys z
    *** QED
```

It is likely that the PLE code throws some unhandled error, which is perhaps triggered by the change in the interpretation of Set symbols in https://github.com/ucsd-progsys/liquid-fixpoint/pull/688

clayrat:
Fixes #2288 

Don't merge yet (before https://github.com/ucsd-progsys/liquid-fixpoint/pull/696), the LF submodule currently points to a branch of a fork.

This updates LF to the version with compacted sort coercions for Sets, which fixes crashes for PLE-heavy code resulting from missing coercions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ commands:
       - checkout
       - add_ssh_keys
       - run: git submodule sync
-      - run: git submodule update --init
+      - run: git submodule update --init --remote
 
   cabal_build_and_test:
     description: "Build the project and run the tests"
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "liquid-fixpoint"]
 	path = liquid-fixpoint
-	url = https://github.com/ucsd-progsys/liquid-fixpoint.git
+	url = https://github.com/clayrat/liquid-fixpoint.git
+	branch = ple-theories
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 ## Next
 
+- Support for GHC 9.8.2.
+- Fixed the polymorphism-related crash in liquid-fixpoint caused by a restrictive theory encoding [#2272](https://github.com/ucsd-progsys/liquidhaskell/issues/2272).
+
 ## 0.9.8.1 (2024-02-05)
 
 - Set support for GHC 9.8.1 [#2248](https://github.com/ucsd-progsys/liquidhaskell/pull/2248)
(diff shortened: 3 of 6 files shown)


clayrat:
filtered
bot
top

ranjitjhala:
Odd to see the `Data.Map.Base` slowdown; it doesn't really use `Set` does it?

clayrat:
@ranjitjhala I think everything after the first ~8 slowdowns is more or less statistical noise, and we should focus first on regaining performance on large benchmarks which do not use `Set`/`Map`. I've found out that disabling the coercions just on the `Elaborate` code path (while keeping them on `Checkable`) is enough to get the perf back, so the plan is currently to automatically decide whether the elaborator should perform the full coercion on the environment based on whether `Set`/`Map` sort symbols are used in it.

clayrat:
After some investigating, I've discovered that the environment coercion on the `Elaborate` code path is the biggest offender in terms of performance and moved it earlier to the point of environment creation. I think the charts look much better now:

filtered
bot
top

ranjitjhala:
Great, thank you!!!

- Ranjit.

nikivazou:
Much better! Let's merge! 

BTW, did you add a test that could not run before (i.e., was creating a crash) and passes now? I believe the feature of this PR should close some issues related with sets and SMT crashing!

clayrat:
> Much better! Let's merge!
> 
> BTW, did you add a test that could not run before (i.e., was creating a crash) and passes now? I believe the feature of this PR should close some issues related with sets and SMT crashing!

Oh yeah, that's a good point, the initial issue was one missing local coercion that I subsequently added to make some PLE-enabled code in dynamic-policies go through. I'll add a minimized form of that code as a new test.

clayrat:
Added a new test and inserted an additional coercion in LF (see https://github.com/ucsd-progsys/liquid-fixpoint/pull/696#issuecomment-2225652088), here are the new performance charts:

filtered
bot
top

PR #2303, opened 2024-06-18
Add a minimal failing test case of redundant transitive imports

archbung:
This test case adds modules `A`, `B`, and `C` where `B` uses a definition from `A` and `C` imports `B`. However, when compiling `C`, LH complains that it cannot find a definition from `A` even though it should be in scope.

Could you take a look @facundominguez?


facundominguez:
Looks great, thanks @archbung!

PR #2310, opened 2024-07-23
Specification of a new `assume reflect` annotation

jarctan:
This PR specifies the role and use of the `assume reflect` annotation, designed to address issue #2126. I'm looking for any feedback on the specification, before I can have a PR for the implementation for it!


facundominguez:
I'm merging the spec to keep it for future reference in the repo.

Issue #2285, opened 2024-05-21
Specification of div seems to be incorrect


facundominguez:
The following example doesn't go well
```Haskell
iterateTo0 :: Int -> Int
iterateTo0 0 = 0
iterateTo0 n = iterateTo0 (div n 2)

main :: IO ()
main = print $ iterateTo0 (-1)
```
It passes verification, but then fails to terminate.

I think the problem might be that [the specification](https://github.com/ucsd-progsys/liquidhaskell/blob/5fbff05aa0ed725a75097a4807f3347fe10df6d6/src/GHC/Real_LHAssumptions.hs#L24-L29) of `div` uses real division where integer division is expected.

nikivazou:
This specification was written for div on Reals this is why the filename has Real. If I recall correctly, there was a different one for the integer division.

facundominguez:
`div` comes from `GHC.Real`.
```
ghci> :i div
type Integral :: * -> Constraint
class (Real a, Enum a) => Integral a where
  ...
  div :: a -> a -> a
  ...
  	-- Defined in ‘GHC.Real’
infixl 7 `div`
```
but it seems to be instantiated only with types that need integer division.
```
ghci> :i GHC.Real.Integral
type Integral :: * -> Constraint
class (Real a, Enum a) => Integral a where
  ...
  div :: a -> a -> a
  ...
  {-# MINIMAL quotRem, toInteger #-}
  	-- Defined in ‘GHC.Real’
instance Integral Int -- Defined in ‘GHC.Real’
instance Integral Integer -- Defined in ‘GHC.Real’
instance Integral Word -- Defined in ‘GHC.Real’
```

ranjitjhala:
Good catch @facundominguez !!! Specifically the clause on line 27 
https://github.com/ucsd-progsys/liquidhaskell/blob/5fbff05aa0ed725a75097a4807f3347fe10df6d6/src/GHC/Real_LHAssumptions.hs#L27

Is wrong: the output is only < x if 0 < x

We should have quickchecked these specs!

facundominguez:
Fixed by #2339.

PR #2311, opened 2024-07-26
Add missing assumptions in GHC.Real.div

marco-bazzani:
Previously GHC.Real.div had the assumptions `((1 < y)  => v < x )`  and  ` ((y >= 1)  => v <= x)`. These are incorrect as written. If `x = -1` and `y = 2` then `v = -1/2`, which means `v > x` and contradicts both assumptions. The fixed versions are `((x > 0 && 1 < y)  => v < x )`  and  ` ((x >= 0 && y >= 1)  => v <= x)`.


facundominguez:
Thanks @mbazzani! This is almost fixing #2285. How about adding some test, and using `div` instead of `/` in the spec?

facundominguez:
Superseeded by #2339.

PR #2313, opened 2024-07-31
Implementation of `assume-reflect`

jarctan:
Here is the implementation of assume-reflect, following the specification given in #2310. Any comments welcome!

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
@@ -62,6 +62,8 @@ import qualified Language.Haskell.Liquid.Transforms.CoreToLogic as CoreToLogic
 import           Control.Arrow                    (second)
 import Data.Hashable (Hashable)
 import qualified Language.Haskell.Liquid.Bare.Slice as Dg
+import Data.Bifunctor (bimap)
+import Data.Function (on)
 
 --------------------------------------------------------------------------------
 -- | De/Serializing Spec files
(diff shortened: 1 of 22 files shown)


ranjitjhala:
Wow this is awesome, thanks!!!

jarctan:
I see that there are compile errors for the `stack_900` test suite that I didn't have locally, is there a command that I can use to reproduce these errors? I was using `stack build liquidhaskell` and then `./scripts/test/test_plugin.sh`

facundominguez:
I think the error is due to the fact that liquidhaskell is build with `-Werror` in CI. I think you can pass `--pedantic` to `stack` in order to enable that.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -694,6 +709,12 @@ addReflSigs :: Bare.Env -> ModName -> BareRTEnv -> GhcSpecRefl -> GhcSpecSig ->
 ------------------------------------------------------------------------------------------
 addReflSigs env name rtEnv refl sig =
   sig { gsRefSigs = F.notracepp ("gsRefSigs for " ++ F.showpp name) $ map expandReflectedSignature reflSigs
+      -- We make sure that the reflected functions are excluded from `gsAsmSigs`, except for the signatures of
+      -- actual functions in assume-reflect. The latter are added here because 1. it's what makes tests work
+      -- and 2. so that we probably "shadow" the old signatures of the actual function correctly. Note that even if the
+      -- signature of the actual function was asserted and not assumed, we do not put our new signature for the actual function
+      -- in `gsTySigs` (which is for asserted signatures). Indeed, the new signature will *always* be an assumption since we
+      -- add the extra post-condition that the actual and pretended function behave in the same way.
```
facundominguez, line 717: @nikivazou, does this ring well to you?

facundominguez: Even though this is merged, I still would like to know what you think and would contribute edits if necessary.

ranjitjhala: This makes sense to me!

PR #2318, opened 2024-08-21
chore: Update docs to make clearer how to set options as pragmas

OlaoluwaM:
Option pragmas need to begin with two dashes.


facundominguez:
Thanks @OlaoluwaM!

PR #2326, opened 2024-09-05
Reflection using interface files

jarctan:
Here's the implementation of a new feature related to reflection, which is the ability to reflect symbols from imported modules using the unfoldings in interface files.
Feel free to comment or suggest!

### DIFF
--- a/docs/mkDocs/docs/options.md
+++ b/docs/mkDocs/docs/options.md
@@ -187,6 +187,41 @@ Opaque reflections:
 - GHC.Internal.Real.even
 ```
 
+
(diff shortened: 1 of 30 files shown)


facundominguez:
I would be merging this next Friday.

This feature is not deemed very useful at the moment, because the unfoldings in interface files often involve features that cannot be reflected. Either they use local recursive bindings, or they use lambda abstractions, or they use type-classes. However, it has potential to become useful in the future, and the author and I think it may be worth keeping it alive by merging it, while we wait for that future.

Let me know if you would like to manage it in some other way.

Issue #2319, opened 2024-08-23
Falsity introduced by shadowing symbols of the logic


jarctan:
I think there is a way to introduce falsity by shadowing names used in the logic. An example:

```Haskell
{-@ LIQUID "--ple"    	@-}

module Test where

import Prelude hiding (not)

not :: Bool -> Bool
not = id

{-@ reflect shouldBeId @-}
shouldBeId :: Bool -> Bool
shouldBeId x = not x

{-@ lemma :: {shouldBeId False} @-}
lemma = ()
```
The .fq file tells us the reflection of `shouldBeId` which is incorrect:
```
define Test.shouldBeId (lq1 : bool) : bool = {((~ (lq1)))}
```

The issue seemingly comes from the `toPredApp` function of `CoreToLogic` which is not matching against the fully-qualified name of symbols, but just against their local name, hence allowing for "puns" to pretend to be the original function. I might have a PR for this soon.

facundominguez:
Fixed in #2327

PR #2327, opened 2024-09-06
Fully qualify names in CoreToLogic

jarctan:
Related to issue #2319, in this PR we fully qualify symbols when transforming CoreExpr into Expr of the logic so we don't mistake builtin symbols with functions defined by the user (see example in the issue).

Worth mentioning, it means that this does not pass verification:

```
{-@ LIQUID "--ple"    	@-}
{-@ LIQUID "--reflection"    	@-}

module Inc00 where

import Prelude hiding (not)

{-@ reflect not @-}
not :: Bool -> Bool
not a = a

{-@ lemma :: {not True} @-}
lemma = ()
```

If you want to refer to your reflected functions that clash with builtin symbols of the logic, you have to fully qualify them:

```
{-@ LIQUID "--ple"    	@-}
{-@ LIQUID "--reflection"    	@-}

module Inc00 where

import Prelude hiding (not)

{-@ reflect not @-}
not :: Bool -> Bool
not a = a

{-@ lemma :: {Inc00.not True} @-}
lemma = ()
```

So that there's no ambiguity. Without qualification, we assume `not` (or the like) to refer to the builtin symbols from the logic.

Any comments welcome!

### DIFF
--- a/liquid-prelude/src/Language/Haskell/Liquid/Prelude.hs
+++ b/liquid-prelude/src/Language/Haskell/Liquid/Prelude.hs
@@ -134,10 +134,18 @@ safeZipWith f (a:as) (b:bs) = f a b : safeZipWith f as bs
 safeZipWith _ []     []     = []
 safeZipWith _ _ _ = error "safeZipWith: cannot happen!"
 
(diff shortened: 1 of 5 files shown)


facundominguez (approved):
LGTM!

PR #2333, opened 2024-09-09
Use the new Bag theory

### LINKED ISSUE #2283: Map measures also generate erroneous constraints when applied to non-polymorphic datatypes
As expected, we can create a test case analogous to https://github.com/ucsd-progsys/liquidhaskell/issues/2272 that crashes LH when working with Maps containing types that are not convertible to `Int`.

```haskell
{-@ LIQUID "--reflection"     @-}

module PolyMapBug where

import Language.Haskell.Liquid.Bag

data Lst a = Emp | Cons a (Lst a)

{-@ measure lstHd @-}
lstHd :: Ord a => Lst a -> Bag a
lstHd  Emp       = empty
lstHd (Cons x _) = put x empty

{-@ lcons :: p: Lst l -> {v:Lst (Lst l) | v = Cons p Emp } @-}
lcons :: Lst l -> Lst (Lst l)
lcons p = Cons p Emp
```

This crashes with

`crash: SMTLIB2 respSat = Error "line 3 column 34267: unknown constant smt_map_sel ((Array Int Int) (PolyMap.Lst Int)) declared: (declare-fun smt_map_sel ((Array Int Int) Int) Int) "`

Since https://github.com/ucsd-progsys/liquid-fixpoint/pull/688 deals only with Sets, this issue is intended to serve as a test case for a follow-up PR that would also elaborate the Map theory into Arrays.

clayrat:
A companion PR to https://github.com/ucsd-progsys/liquid-fixpoint/pull/703, refactoring `Language.Haskell.Liquid.Bag` to use the new LF Bag theory and adding a test for polymorphic Bags. It also removes the rewrite rules for `Data.Map.Base.insert` and `Data.Map.Base.select`, as Haskell Maps do not correspond 1-to-1 to SMT arrays - the latter have their default element determined at the time of creation, whereas in Haskell one has to supply the default element when querying a map (or unwrapping the result from `Maybe`).

Fixes #2283

_Please do not merge before https://github.com/ucsd-progsys/liquid-fixpoint/pull/703 is merged and the submodule is reverted to point to the main LF repository._


clayrat:
filtered
bot
top

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "liquid-fixpoint"]
 	path = liquid-fixpoint
-	url = https://github.com/ucsd-progsys/liquid-fixpoint.git
\ No newline at end of file
+	url = https://github.com/clayrat/liquid-fixpoint.git
```
nikivazou: As you mention, this should change before merging!

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/CoreToLogic.hs
```diff
@@ -30,13 +30,10 @@ coreToLogic = unlines
...
   , "define GHC.Internal.Base.$ f x        = (f x)"
   , ""
-  , "define Language.Haskell.Liquid.Bag.get k m   = (Map_select m k)"
-  , "define Language.Haskell.Liquid.Bag.put k m   = (Map_store m k (1 + (Map_select m k)))"
-  , "define Language.Haskell.Liquid.Bag.union m n = (Map_union  m n)"
-  , "define Language.Haskell.Liquid.Bag.empty     = (Map_default 0)"
-  , ""
-  , "define Data.Map.Base.insert k v m     = (Map_store m k v)"
-  , "define Data.Map.Base.select k v       = (Map_select m k)"
+  , "define Language.Haskell.Liquid.Bag.get k b   = (Bag_count b k)"
```
nikivazou: It would be better if that was a text file. Even better extensible by the user.

facundominguez: It used to be a text file, but it required some contortions to find it at runtime. One of the pains was having to define an environment variable `liquidhaskell_datadir` whenever someone wanted to run liquidhaskell after building it.

If the user should be able to write her own `define` annotations, maybe they should be supported in source code, instead of constraining them to a specific file.

clayrat: Hm, so should we implement these local annotations instead?

nikivazou: @facundominguez, so if we defined the specifics for example of the `Data.Map` to a `Liquid.Haskell.Data.Map` haskell file and the user imported it together with the `Data.Map`, would that be ok? 

I think the current setting is ok for some core theories, but more specific things, like Bag could be selectively imported by the user. Esp. if we aim to define new theories. 

We could anyways merge this and open another PR if it is extracted.

facundominguez: > so if we defined the specifics for example of the Data.Map to a Liquid.Haskell.Data.Map haskell file and the user imported it together with the Data.Map, would that be ok?

Perhaps the place to put them would be `Data.Map_LHAssumptions` (there is one such file for [Data.Set](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/src/Data/Set_LHAssumptions.hs) already).

Looks like some changes to the code will be necessary as I don't think LH is making an effort to collect `define` annotations from other places than `CoreToLogic.hs`.

facundominguez:
Is this good to merge? It seems that I cannot update LF without breaking tests in LH related to Bags.

clayrat:
Yes, everything is clear from my side

PR #2339, opened 2024-09-17
Add missing assumptions in GHC.Internal.Real.div and its tests. Use div instead of / in the spec

RCoeurjoly:
This PR is a continuation of https://github.com/ucsd-progsys/liquidhaskell/pull/2311.

Co-authors are:

- @facundominguez (for the `iterateTo0` definition from https://github.com/ucsd-progsys/liquidhaskell/issues/2285)
- @mbazzani (for the assumptions in `GHC.Internal.Real.div` in https://github.com/ucsd-progsys/liquidhaskell/pull/2311)
- and myself.

I would also be interested in adding `quickcheck` tests for the specs, as suggested by @ranjitjhala in https://github.com/ucsd-progsys/liquidhaskell/issues/2285#issuecomment-2122794785, but I think it would be better to do that in a separate PR.

Tested locally.

### DIFF
--- a/src/GHC/Real_LHAssumptions.hs
+++ b/src/GHC/Real_LHAssumptions.hs
@@ -26,7 +26,7 @@ class (GHC.Internal.Real.Real a, GHC.Internal.Enum.Enum a) => GHC.Internal.Real.
                                                     ((x >= 0 && y >= 1) => v <= x) &&
                                                     ((1 < y && x >= 0)  => v < x) &&
                                                     ((1 < y && x < 0)   => v > x) &&
-                                                    ((y >= 1 && x > 0)  => v <= x) &&
+                                                    ((y >= 1 && x >= 0)  => v <= x) &&
                                                     ((x < 0 && y > 0)   => v <= 0) &&
                                                     ((x > 0 && y < 0)   => v <= 0) &&
                                                     ((x < 0 && y < 0)   => v >= 0)
--- a/src/GHC/Real_LHAssumptions.hs
+++ b/src/GHC/Real_LHAssumptions.hs
@@ -26,7 +26,7 @@ class (GHC.Internal.Real.Real a, GHC.Internal.Enum.Enum a) => GHC.Internal.Real.
                                                     ((x >= 0 && y >= 1) => v <= x) &&
                                                     ((1 < y && x >= 0)  => v < x) &&
                                                     ((1 < y && x < 0)   => v > x) &&
-                                                    ((y >= 1)           => v <= x) &&
+                                                    ((y >= 1 && x > 0)  => v <= x) &&
                                                     ((x < 0 && y > 0)   => v <= 0) &&
                                                     ((x > 0 && y < 0)   => v <= 0) &&
                                                     ((x < 0 && y < 0)   => v >= 0)
--- a/src/GHC/Real_LHAssumptions.hs
+++ b/src/GHC/Real_LHAssumptions.hs
@@ -21,11 +21,15 @@ class (GHC.Internal.Real.Real a, GHC.Internal.Enum.Enum a) => GHC.Internal.Real.
   GHC.Internal.Real.rem :: x:a -> y:{v:a | v /= 0} -> {v:a | ((v >= 0) && (v < y))}
   GHC.Internal.Real.mod :: x:a -> y:{v:a | v /= 0} -> {v:a | v = x mod y && ((0 <= x && 0 < y) => (0 <= v && v < y))}
 
-  GHC.Internal.Real.div :: x:a -> y:{v:a | v /= 0} -> {v:a | (v = (x / y)) &&
+  GHC.Internal.Real.div :: x:a -> y:{v:a | v /= 0} -> {v:a | (v = div x y) &&
                                                     ((x >= 0 && y >= 0) => v >= 0) &&
-                                                    ((x >= 0 && y >= 1) => v <= x) && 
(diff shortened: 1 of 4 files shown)


tests/terminate/pos/Div.hs
```diff
@@ -0,0 +1,19 @@
+module Div where
+
+{-@ iterateTo0 :: n:Int -> {v:Int | v == 0} @-}
+iterateTo0 :: Int -> Int
+iterateTo0 n
+  | n <= 0    = 0
+  | otherwise = iterateTo0 (div n 2)
+
+{-@ test :: {v:Int | v == 0} @-}
+test = iterateTo0 (-1)
```
facundominguez, line 12: How about adding a test that also checks that non-termination is correctly detected? Something like

```haskell
{-@ fail iterateTo0 @-}
iterateTo0 :: Int -> Int
iterateTo0 0 = 0
iterateTo0 n = iterateTo0 (div n 2)
```

or it could use the `--expect-error-containing` flag as other tests in the testsuite.

```haskell
{-@ LIQUID "--expect-error-containing=..." @-}

{-@ fail iterateTo0 @-}
iterateTo0 :: Int -> Int
iterateTo0 0 = 0
iterateTo0 n = iterateTo0 (div n 2)
```

tests/terminate/pos/Div.hs
```diff
@@ -0,0 +1,19 @@
+module Div where
```
facundominguez, line 3: Could be helpful to start the test with a comment that makes explicit what is being tested.

src/GHC/Real_LHAssumptions.hs
```diff
@@ -21,11 +21,15 @@ class (GHC.Internal.Real.Real a, GHC.Internal.Enum.Enum a) => GHC.Internal.Real.
...
-  GHC.Internal.Real.div :: x:a -> y:{v:a | v /= 0} -> {v:a | (v = (x / y)) &&
+  GHC.Internal.Real.div :: x:a -> y:{v:a | v /= 0} -> {v:a | (v = div x y) &&
                                                     ((x >= 0 && y >= 0) => v >= 0) &&
-                                                    ((x >= 0 && y >= 1) => v <= x) && 
-                                                    ((1 < y)            => v < x ) && 
-                                                    ((y >= 1)           => v <= x)  
+                                                    ((x >= 0 && y >= 1) => v <= x) &&
+                                                    ((1 < y && x >= 0)  => v < x) &&
+                                                    ((1 < y && x < 0)   => v > x) &&
+                                                    ((y >= 1)           => v <= x) &&
```
facundominguez: `((y >= 1)           => v <= x)`: this is true only if `x` is non-negative.

facundominguez (commented):
Thanks @RCoeurjoly. I left some comments.

tests/terminate/pos/Div.hs
RCoeurjoly: Isn't it implied by the location of the test that we are testing termination? Are we testing something else? As far as I can tell, we are not testing the `GHC.Internal.Real.div` assumptions.

src/GHC/Real_LHAssumptions.hs
RCoeurjoly: Fixed. Thanks!

tests/terminate/pos/Div.hs
RCoeurjoly: Added a test for this in last commit. Thanks!

src/GHC/Real_LHAssumptions.hs
```diff
@@ -21,11 +21,15 @@ class (GHC.Internal.Real.Real a, GHC.Internal.Enum.Enum a) => GHC.Internal.Real.
...
-  GHC.Internal.Real.div :: x:a -> y:{v:a | v /= 0} -> {v:a | (v = (x / y)) &&
+  GHC.Internal.Real.div :: x:a -> y:{v:a | v /= 0} -> {v:a | (v = div x y) &&
                                                     ((x >= 0 && y >= 0) => v >= 0) &&
-                                                    ((x >= 0 && y >= 1) => v <= x) && 
-                                                    ((1 < y)            => v < x ) && 
-                                                    ((y >= 1)           => v <= x)  
+                                                    ((x >= 0 && y >= 1) => v <= x) &&
+                                                    ((1 < y && x >= 0)  => v < x) &&
+                                                    ((1 < y && x < 0)   => v > x) &&
+                                                    ((y >= 1 && x > 0)  => v <= x) &&
```
facundominguez:
```suggestion
                                                    ((y >= 1 && x >= 0)  => v <= x) &&
```

tests/terminate/neg/Div.hs
```diff
@@ -0,0 +1,18 @@
+{-@ LIQUID "--expect-error-containing=Liquid Type Mismatch" @-}
+module Div where
```
facundominguez, line 4:
```suggestion
-- Tests that non-termination is detected when div is used in the
-- argument of a recursive call.
module Div where
```

tests/terminate/pos/Div.hs
facundominguez: It is true that the reader might assemble the explanation. But something more explicit is cheap and can make it harder to get it wrong. Here's a try.

```suggestion
-- Tests that termination is detected when using div in the argument
-- of the recursive call.
module Div where
```

facundominguez (approved):
Thanks @RCoeurjoly!

PR #2351, opened 2024-09-26
Simplify makeGhcSpec0

facundominguez:
No changes of behavior here.


facundominguez:
This PR contains some refactorings that I discovered when analyzing name resolution.

PR #2360, opened 2024-10-04
Only try to expand newtype applications that are saturated

### LINKED ISSUE #2359: Trying to rewrite unsaturated newtype applications
With this input file:

```
{-# LANGUAGE NoImplicitPrelude #-}
module One where

class Cls f where
    meth :: (a -> b) -> f a -> f b

newtype One a = One{ getOne :: a}

instance Cls One where
    meth f (One x) = One (f x)
```

When using a build of GHC that has assertions enabled, we can see that LH ends up calling `newTyConInstRhs` for an undersaturated application of `One`:

```
[1 of 1] Compiling One              ( input/One.hs, input/One.o )
WARNING: Missing RenamedSource
WARNING: Missing RenamedSource

<no location info>: error:
    panic! (the 'impossible' happened)
  GHC version 9.8.2:
	ASSERT failed!
  One
  []
  [a_amb]
  Call stack:
        assertPpr, called at compiler/GHC/Core/Type.hs:1651:7 in ghc-9.8.2-inplace:GHC.Core.Type
    panic, called at compiler/GHC/Utils/Error.hs:503:29 in ghc-9.8.2-inplace:GHC.Utils.Error


Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
```

gergoerdi:
Fixes #2359

### DIFF
--- a/liquidhaskell-boot/src-ghc/Liquid/GHC/API.hs
+++ b/liquidhaskell-boot/src-ghc/Liquid/GHC/API.hs
@@ -271,6 +271,9 @@ import GHC.Core.Coercion              as Ghc
     )
 import GHC.Core.Coercion.Axiom        as Ghc
     ( coAxiomTyCon )
+import GHC.Core.TyCon                 as Ghc
+    ( newTyConEtadArity
+    )
 import GHC.Core.ConLike               as Ghc
     ( ConLike(RealDataCon) )
 import GHC.Core.DataCon               as Ghc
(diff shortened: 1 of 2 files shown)


facundominguez:
Thanks @gergoerdi, would you add a regression test for this? Some new module in `tests:exe:unit-pos-3` would do, I think.

gergoerdi:
> Thanks @gergoerdi, would you add a regression test for this? Some new module in `tests:exe:unit-pos-3` would do, I think.

How exactly will that work? The bug is only easy to detect on builds of GHC with assertions enabled. Without that, it "just" goes off the invariants, with no easy way to observably crash it.

facundominguez:
> The bug is only easy to detect on builds of GHC with assertions enabled.

Oh, that renders my question moot. Please, disregard.

Issue #2357, opened 2024-10-02
Using the plugin without file access


gergoerdi:
It seems that the `LiquidHaskellBoot` plugin assumes that whatever module it is processing is originally coming from a file on disk. This assumption, for example, manifests itself in `Language.Haskell.Liquid.GHC.Interface.modSummaryHsFile`, where it is assumed that the currently processed module's `ModSummary` has a non-`Nothing` value in `ml_hs_file`.

I'm working on a compiler where this assumption doesn't, in general, hold.

Is there a way to avoid that? It seems that this file is re-parsed by LH to get the refinement types. But those types are already available (in text form) in comments, which should be kept by the normal GHC parser as far as `parsedResultAction` plugins are concerned. So would it be possible for the LHB plugin to install a `parsedResultAction` hook instead of trying to get comments from the supposed original text source files?

facundominguez:
> So would it be possible for the LHB plugin to install a parsedResultAction hook instead of trying to get comments from the supposed original text source files?

Maybe. I don't know of anyone trying that before. Relatedly, there are provisions in the code to ensure that comments are kept by the lexer (setting `Opt_KeepRawTokenStream`), but I think you saw that already.

gergoerdi:
I've added a pull request #2367 that implements the idea in the original comment. It's not quite full file-less operation, since there's also some caching code that tries to copy files around, but I think it's a good step in the direction.

gergoerdi:
Other than #2367, I have stumbled upon two more points where LH (or, more specifically, the `liquidhaskell-boot` plugin) tries to access source files (all findings based on aef80ef01f5445e22fc79384537a1e70420538e3):

- [ ] In `Language.Haskell.Liquid.UX.Annotate.annotate`, when `noannotations` is `False` (i.e. when annotations are enabled... as a side note, I hate flipped boolean flags...). This, of course, is easy for clients to work around by just making sure `noannotations` is always set.

  https://github.com/ucsd-progsys/liquidhaskell/blob/aef80ef01f5445e22fc79384537a1e70420538e3/liquidhaskell-boot/src/Language/Haskell/Liquid/UX/Annotate.hs#L93-L110

- [x] In `Language.Haskell.Liquid.Liquid.checkTargetInfo`. This is seemingly gated by `compileSpec`, but of course clients can't just turn off `compileSpec` without also disabling actual verification.
  
[...]

PR #2367, opened 2024-10-08
Avoid re-parsing of input source files 

gergoerdi:
Part of the work towards #2357

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE RecordWildCards            #-}
+{-# LANGUAGE LambdaCase                 #-}
 
 module Language.Haskell.Liquid.GHC.Plugin (
 
@@ -40,17 +41,22 @@ import           GHC.LanguageExtensions
 import           Control.Monad
 import qualified Control.Monad.Catch as Ex
 import           Control.Monad.IO.Class (MonadIO)
+import           Control.Monad.Catch                        (bracket)
 
 import           Data.Coerce
 import           Data.Function                            ((&))
 import qualified Data.List                               as L
+import           Data.IORef
 import qualified Data.Set                                as S
 import           Data.Set                                 ( Set )
+import qualified Data.Map                                as M
+import           Data.Map                                 ( Map )
 
 
 import qualified Data.HashSet                            as HS
 import qualified Data.HashMap.Strict                     as HM
 
+import           System.IO.Unsafe                         ( unsafePerformIO )
 import           Language.Fixpoint.Types           hiding ( errs
                                                           , panic
                                                           , Error
@@ -138,6 +144,26 @@ plugin = GHC.defaultPlugin {
                 pure newGblEnv'
 
 --------------------------------------------------------------------------------
+-- | Inter-phase communication -------------------------------------------------
+--------------------------------------------------------------------------------
+--
+-- Since we have no good way of passing extra data between different
+-- phases of our plugin, we resort to horrible global mutable state.
+
+data Breadcrumb
+  = Typechecking
+
+breadcrumbs :: IORef (Map Module Breadcrumb)
+breadcrumbs = unsafePerformIO $ newIORef mempty
+{-# NOINLINE breadcrumbs #-}
+
+{- HLINT ignore "Use tuple-section" -}
+swapBreadcrumb :: (MonadIO m) => Module -> Maybe Breadcrumb -> m (Maybe Breadcrumb)
+swapBreadcrumb mod new = liftIO $ atomicModifyIORef' breadcrumbs $ \breadcrumbs ->
+  let (old, breadcrumbs') = M.alterF (\old -> (old, new)) mod breadcrumbs
+  in (breadcrumbs', old)
+
+--------------------------------------------------------------------------------
 -- | GHC Configuration & Setup -------------------------------------------------
 --------------------------------------------------------------------------------
 
@@ -212,8 +238,25 @@ unoptimiseDynFlags df = updOptLevel 0 df
 --    grab from parsing (again) the module by using the GHC API, so we are really
 --    independent from the \"normal\" compilation pipeline.
 --
-typecheckHook :: Config -> ModSummary -> TcGblEnv -> TcM (Either LiquidCheckException TcGblEnv)
-typecheckHook cfg0 modSummary0 tcGblEnv = do
+typecheckHook  :: Config -> ModSummary -> TcGblEnv -> TcM (Either LiquidCheckException TcGblEnv)
+typecheckHook cfg0 modSummary0 tcGblEnv = bracket startTypechecking endTypechecking $ \case
+  Just Typechecking -> do
+    -- We're being called from the `typecheckModuleIO` call in `typecheckHook`, so we avoid looping
+    pure $ Right tcGblEnv
+  Nothing -> do
+    typecheckHook' cfg0 modSummary0 tcGblEnv
+  where
+    thisModule = tcg_mod tcGblEnv
+
+    -- The LH plugin itself calls the type checker (see call to
+    -- `typecheckModuleIO` in `typecheckHook`).  This would lead to a
+    -- loop if we didn't disable the typechecker plugin.
+    startTypechecking = swapBreadcrumb thisModule $ Just Typechecking
+    endTypechecking = \case
+        _ -> pure ()
+
+typecheckHook' :: Config -> ModSummary -> TcGblEnv -> TcM (Either LiquidCheckException TcGblEnv)
+typecheckHook' cfg0 modSummary0 tcGblEnv = do
   debugLog $ "We are in module: " <> show (toStableModule thisModule)
   let modSummary =
         updateModSummaryDynFlags (customDynFlags . unoptimiseDynFlags) modSummary0
@@ -236,10 +279,7 @@ typecheckHook cfg0 modSummary0 tcGblEnv = do
             parsed2 = parsed { pm_mod_summary = modSummary2 }
             env2 = env { hsc_dflags = ms_hspp_opts modSummary2 }
 
-        -- The LH plugin itself calls the type checker (see following line). This
-        -- would lead to a loop if we didn't remove the plugin when calling the type
-        -- checker.
-        typechecked     <- liftIO $ typecheckModuleIO (dropPlugins env2) (LH.ignoreInline parsed2)
+        typechecked     <- liftIO $ typecheckModuleIO env2 (LH.ignoreInline parsed2)
         resolvedNames   <- liftIO $ LH.lookupTyThings env2 tcGblEnv
         availTyCons     <- liftIO $ LH.availableTyCons env2 tcGblEnv (tcg_exports tcGblEnv)
         availVars       <- liftIO $ LH.availableVars env2 tcGblEnv (tcg_exports tcGblEnv)
@@ -255,8 +295,6 @@ typecheckHook cfg0 modSummary0 tcGblEnv = do
   where
     thisModule :: Module
     thisModule = tcg_mod tcGblEnv
-
-    dropPlugins hsc_env = hsc_env { hsc_plugins = emptyPlugins }
 
     continue = pure $ Left (ErrorsOccurred [])
 
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
@@ -46,6 +46,7 @@ import           Control.Monad.Catch                        (bracket)
 import           Data.Coerce
 import           Data.Function                            ((&))
 import qualified Data.List                               as L
+import           Data.Maybe
 import           Data.IORef
 import qualified Data.Set                                as S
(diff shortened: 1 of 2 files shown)


gergoerdi:
Yes, I'm using a global `IORef`. 

If we want to avoid that, we need to find a place to store an arbitrary value (in this case, a `ParsedModule`) and have it make its way all the way to the post-typechecking hook. I was hoping we could use a module annotation (analogously to how you can put already resolved names into a `RdrName`) , but it looks like those only contain text and an expression.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -196,6 +199,40 @@ unoptimiseDynFlags df = updOptLevel 0 df
...
+  pure parsedResult
+  where
+    parsed = ParsedModule
+      { pm_mod_summary = ms
+      , pm_parsed_source = hpm_module . parsedResultModule $ parsedResult
+      , pm_extra_src_files = hpm_src_files . parsedResultModule $ parsedResult
+      }
+
+parsedRef :: IORef ParsedModule
+parsedRef = unsafePerformIO $ newIORef $ error "parsedResultAction hook didn't run"
```
facundominguez: Is this global variable going to be a problem when multiple files are compiled concurrently?

Probably it should be a Set, and the plugin should remove modules from the set when they are no longer needed.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -196,6 +199,40 @@ unoptimiseDynFlags df = updOptLevel 0 df
...
 
+--------------------------------------------------------------------------------
+-- | Parsing phase -------------------------------------------------------------
+--------------------------------------------------------------------------------
+
+-- | We hook at this stage of the pipeline to store the parsed module
+-- for later processing during typechecking:
+--
+-- * Comments are preserved during parsing because we turn on
+--   'Opt_KeepRawTokenStream'. So we can get the /LH/ specs from there.
```
facundominguez, line 237: Note that different flags are set at two different times. There is `lhDynFlags`, set before the plugin. And there is `customDynFlags` set only for parsing, typechecking, and desugaring.

The benefit of `customDynFlags` is that it doesn't affect anything else in the compiler for LH sake. But I'm guessing that you want `Opt_KeepRawTokenStream` to affect how GHC parses all files, and then you will have to move it to `lhDynFlags`.

facundominguez (commented):
Thanks @gergoerdi.

I left a couple of comments for you to consider.

Does this PR solve your use case? If not, what else would be necessary?

gergoerdi:
> Does this PR solve your use case? If not, what else would be necessary?

Not fully yet, no. I'll add details to #235, but basically there are two other points where source files are read in by LH. One of them is easily disabled by just setting `noannotations = True` in the config. The other one is some random place for caching I guess? I'll look it up.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
gergoerdi: OK I've added a new global mutable state that stores the LH breadcrumb for each module, shared with #2368.

gergoerdi:
Note that due to the breadcrumb stuff, this PR now includes #2368 as well. I think #2368 has less potential for disruption, and it fixes a manifest bug, so I think it will go in before this anyway.

gergoerdi:
FWIW, I 100% disagree with the HLint suggestion here:

> warning	Use tuple-section
> Suggestion in `swapBreadcrumb` in module `Language.Haskell.Liquid.GHC.Plugin`: 
> # Use tuple-section
>   * Found: `\ old -> (old, new)`
>   * Perhaps: `(, new)`
>   * Note: may require `{-# LANGUAGE TupleSections #-}` adding to the top of the file
>

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -196,6 +227,38 @@ unoptimiseDynFlags df = updOptLevel 0 df
...
+--   comment on 'typecheckHook') which requires access to a full
+--   'TypecheckedModule'. Annoyingly, the typechecker hook is given
+--   only a 'TcGblEnv', so we have to redo typechecking there. And
+--   the input to that needs to be a 'ParsedModule'.
+--
+parsedHook :: [CommandLineOption] -> ModSummary -> ParsedResult -> Hsc ParsedResult
+parsedHook _ ms parsedResult = do
+  liftIO $ atomicModifyIORef' breadcrumbs $ \breadcrumbs ->
+      let breadcrumbs' = HM.alter (<|> Just (Parsed parsed)) thisModule breadcrumbs
+      in (breadcrumbs', ())
```
facundominguez:
```suggestion
  _ <- swapBreadcrumbs thisModule (Just (Parsed parsed))
```
?

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -196,6 +227,38 @@ unoptimiseDynFlags df = updOptLevel 0 df
...
+--
+-- * Comments are preserved during parsing because we turn on
+--   'Opt_KeepRawTokenStream'. So we can get the /LH/ specs from there.
+--
+-- * The typechecker hook needs to do desugaring itself (see the
+--   comment on 'typecheckHook') which requires access to a full
+--   'TypecheckedModule'. Annoyingly, the typechecker hook is given
+--   only a 'TcGblEnv', so we have to redo typechecking there. And
+--   the input to that needs to be a 'ParsedModule'.
+--
```
facundominguez, line 244:
```suggestion
--
-- LH used to reparse the module in the typechecker hook, but that doesn't work when
-- the source code is not fed from a file (See #2357).
```

facundominguez (commented):
> FWIW, I 100% disagree with the HLint suggestion here:

Want to try disabling that one either globally or only in Plugin.hs?

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -112,11 +117,11 @@ plugin = GHC.defaultPlugin {
     -- See also: https://github.com/ucsd-progsys/liquidhaskell/issues/1727
     -- for a post-mortem.
     liquidPluginGo cfg summary gblEnv = do
+      breadcrumb <- swapBreadcrumb (toStableModule thisModule) (Just Typechecking)
```
facundominguez: Should `swapBreadcrumb` be called again when `liquidPluginGo` returns to remove the breadcrumb? For instance,
```
bracket
 (swapBreadcrumb (toStableModule thisModule) (Just Typechecking))
 (const $ swapBreadcrumb (toStableModule thisModule) Nothing)
 $ \breadcrumb ->
```

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -196,6 +227,38 @@ unoptimiseDynFlags df = updOptLevel 0 df
...
+--   'Opt_KeepRawTokenStream'. So we can get the /LH/ specs from there.
+--
+-- * The typechecker hook needs to do desugaring itself (see the
+--   comment on 'typecheckHook') which requires access to a full
+--   'TypecheckedModule'. Annoyingly, the typechecker hook is given
+--   only a 'TcGblEnv', so we have to redo typechecking there. And
+--   the input to that needs to be a 'ParsedModule'.
+--
+parsedHook :: [CommandLineOption] -> ModSummary -> ParsedResult -> Hsc ParsedResult
+parsedHook _ ms parsedResult = do
```
facundominguez: Should the breadcrumb be inserted only if the module is not skipped in `liquidPlugin`? Otherwise, the AST is going to be retained until the call to GHC finishes.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -136,6 +141,31 @@ plugin = GHC.defaultPlugin {
...
+--------------------------------------------------------------------------------
+--
+-- Since we have no good way of passing extra data between different
+-- phases of our plugin, we resort to horrible global mutable state.
+
+data Breadcrumb
+  = Parsed ParsedModule
+  | Typechecking
+
+breadcrumbs :: IORef (HM.HashMap StableModule Breadcrumb)
```
facundominguez: Any reason to use `HashMap StableModule` rather than `Map ModuleName`. The later is simpler, while being reasonably efficient, I'd hope.

The stability of `StableModule` seems to matter only across different calls to GHC.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
gergoerdi: Is that what we want, though? Note that here, we only replace the breadcrumb with `Parsed parsed` if there's no other breadcrumb already.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
gergoerdi: OK this is up to you guys -- do you prefer having this kind of historical "the code does this, but before it did that" in comments, or only in the Git history?

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
gergoerdi: This would mean that the recursive typechecker hook would be the one removing the breadcrumb. I don't like that, because that means we now have this prerequisite that "there is only one level of recursive typechecking" encoded diffusely in the interaction between here and `typecheckedHook`.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
gergoerdi: Just because we already had `HashMap` in scope. I'll change it to use `Map ModuleName`.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
facundominguez: If we should expect to encounter breadcrumbs already inserted for a module, it would deserve a comment to explain how could that happen and what the code is expected to do in such a case.

When I proposed the reuse of `swapBreadcrumbs`, I was not aware of the possibility.

If we don't expect to find that case, maybe the plugin should produce an error if it happens.

gergoerdi:
In fact, in review I don't like my current code either :)

So what we really want is:

1. First time we parse a module, there is no breadcrumb for the current module so we can store a `Parsed` breadcrumb. Later on, if we get to the after-parser hook again and do find some existing breadcrumb, that is an assertion failure.

2. First time we typecheck a module, there is a `Parsed` breadcrumb for the current module. We replace it with a `Typechecking` breadcrumb, which also makes sure we don't retain all ASTs. Getting to typechecking without a breadcrumb is an assertion failure.

3. Recrusive typecheckers will see that there is already a `Typechecking` breadcrumb and do nothing.

I'm not sure we should ever remove the `Typechecking` breadcrumb because then we have no way of later detecting some of the invalid states above. Maybe replace it with a `Done` breadcrumb that is an error to encounter in both hooks.

I'll redo the implementation based on these ideas.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
facundominguez: How is the breadcrumb removed otherwise?

While the breadcrumb is there the plugin cannot verify the module a second time, and I'd find this surprising. This is fine when GHC is running per module or package, but I've heard of experiments to make it run as a compilation server.

gergoerdi: Oof, I wasn't thinking of running long enough to typecheck the same module multiple times.

In that case, I think the prudent thing to do is to remove the breadcrumb at the end of `typecheckHook`.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
gergoerdi: Is `ModuleName` unambiguous enough? You mentioned using the typechecker as a long-running server. In that case, can there be multiple units checked with the same server? If yes, you can encounter the same `ModuleName` in multiple units.

gergoerdi: (I realize `StableModule` is actually a `StableModuleName` so we're no worse off using `ModuleName` here, but still.)

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
facundominguez: I'd prefer to get your issue mentioned in the comment to motivate the current approach to the reader more plainly.

facundominguez (approved):
Can't find anything else to complain about :tada:

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
facundominguez: It was my slip. I think `Module` is the best choice.

PR #2368, opened 2024-10-08
Don't disable all plugins wholesale before recursive typechecking

### LINKED ISSUE #2366: Unrelated typechecker plugins are disabled by LH
In `Language.Haskell.Liquid.GHC.Plugin.typecheckHook`, we re-typecheck the input just to then be able to desugar it. This is needed because, sadly, `typeCheckResultAction` is not given the full `TypecheckedModule`.

Inside that function, to avoid looping by running the same post-typechecker hook, the plugin list is cleared:

```
        -- The LH plugin itself calls the type checker (see following line). This
        -- would lead to a loop if we didn't remove the plugin when calling the type
        -- checker.
        typechecked     <- liftIO $ typecheckModuleIO (dropPlugins env2) (LH.ignoreInline parsed2)
```

But that is, of course, disastrous when some unrelated plugins are needed to typecheck the given module!

It is not hard to construct a small demonstration of why this is wrong. Let's take the following module, which can only be typechecked if we can prove that `n + n` is equal to `2 * n`, _in general_:

```
{-# LANGUAGE DataKinds, NoStarIsType, RequiredTypeArguments #-}
module Nat where

import GHC.TypeLits

foo :: forall n -> (n + n ~ 2 * n) => ()
foo _ = ()

bar :: forall (n :: Nat) -> ()
bar n = foo n
```

On its own, without LH or any other plugins, this doesn't typecheck:

```
[1 of 1] Compiling Nat              ( input/Nat.hs, input/Nat.o )
input/Nat.hs:12:9: error: [GHC-18872]
    • Couldn't match type: n + n
                     with: 2 * n
        arising from a use of ‘foo’
    • In the expression: foo n
      In an equation for ‘bar’: bar n = foo n
   |
12 | bar n = foo n
   |         ^^^
```

But we can help GHC out by using e.g. [`ghc-typelits-natnormalise`](https://hackage.haskell.org/package/ghc-typelits-natnormalise):

```
{-# LANGUAGE DataKinds, NoStarIsType, RequiredTypeArguments #-}
{-# OPTIONS -fplugin=GHC.TypeLits.Normalise #-}
module Nat where

import GHC.TypeLits

foo :: forall n -> (n + n ~ 2 * n) => ()
foo _ = ()

bar :: forall (n :: Nat) -> ()
bar n = foo n
```

This, _without LH_, typechecks.

But we're back to square one if LH is enabled: the following code fails to typecheck (*in the second typechecker iteration!*) with the same error message as above:

```
{-# LANGUAGE DataKinds, NoStarIsType, RequiredTypeArguments #-}
{-# OPTIONS -fplugin=GHC.TypeLits.Normalise #-}
module Nat where

import GHC.TypeLits

foo :: forall n -> (n + n ~ 2 * n) => ()
foo _ = ()

bar :: forall (n :: Nat) -> ()
bar n = foo n
```

gergoerdi:
Fixes #2366

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE RecordWildCards            #-}
+{-# LANGUAGE LambdaCase                 #-}
 
 module Language.Haskell.Liquid.GHC.Plugin (
 
@@ -40,17 +41,22 @@ import           GHC.LanguageExtensions
 import           Control.Monad
 import qualified Control.Monad.Catch as Ex
 import           Control.Monad.IO.Class (MonadIO)
+import           Control.Monad.Catch                        (bracket)
 
 import           Data.Coerce
 import           Data.Function                            ((&))
 import qualified Data.List                               as L
+import           Data.IORef
 import qualified Data.Set                                as S
 import           Data.Set                                 ( Set )
+import qualified Data.Map                                as M
+import           Data.Map                                 ( Map )
 
 
 import qualified Data.HashSet                            as HS
 import qualified Data.HashMap.Strict                     as HM
 
+import           System.IO.Unsafe                         ( unsafePerformIO )
 import           Language.Fixpoint.Types           hiding ( errs
                                                           , panic
                                                           , Error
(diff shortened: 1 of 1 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -256,12 +264,15 @@ typecheckHook cfg0 modSummary0 tcGblEnv = do
...
     thisModule = tcg_mod tcGblEnv
 
-    dropPlugins hsc_env = hsc_env { hsc_plugins = emptyPlugins }
-
     continue = pure $ Left (ErrorsOccurred [])
 
     updateModSummaryDynFlags f ms = ms { ms_hspp_opts = f (ms_hspp_opts ms) }
 
+isRecursiveChecking :: IORef Bool
+isRecursiveChecking = unsafePerformIO $ newIORef False
```
facundominguez: Thanks again @gergoerdi.

Similar to #2367, to account for concurrency this should use a Set, where modules are added before they are typechecked by LH, and then they are removed after they have been typechecked.

I'd rename `isRecursiveChecking` to `isLHTypechecking`, and maybe place the explanation of the need to break an infinite loop at this point.

Issue #2349, opened 2024-09-24
"Unbound symbol" when using GADTs


pacastega:
With the following code
```haskell
{-# LANGUAGE GADTs #-}
{-@ LIQUID "--reflection" @-}
module TestModule where

data Expr t where
  I :: Int -> Expr Int

  NIL :: Expr [t]
  CONS :: Int -> Expr t -> Expr [t] -> Expr [t]

{-@
data Expr t where
  I :: Int -> Expr Int

  NIL :: Expr {v:[t] | len v = 0}
  CONS :: l:Nat -> head:Expr t -> tail:Expr {v:[t] | len v = l} ->
          Expr {v:[t] | len v = l+1}
@-}
```
I am getting an error about `l` being unbound:
```
    Illegal type specification for `TestModule.$WCONS`
    TestModule.$WCONS :: forall t .
                         x1:GHC.Types.Int -> head##TestModule.CONS:(TestModule.Expr t) -> tail##TestModule.CONS:(TestModule.Expr {VV : [t] | len VV == l##TestModule.CONS}) -> {VV : (TestModule.Expr {VV : [t] | len VV == l##TestModule.CONS + 1}) | is$TestModule.CONS VV
                                                                                                                                                                                                                                                       && not (is$TestModule.I VV)
                                                                                                                                                                                                                                                       && not (is$TestModule.NIL VV)
                                                                                                                                                                                                                                                       && lqdc##$select##TestModule.CONS##1 VV == x1
                                                                                                                                                                                                                                                       && lqdc##$select##TestModule.CONS##2 VV == head##TestModule.CONS
                                                                                                                                                                                                                                                       && lqdc##$select##TestModule.CONS##3 VV == tail##TestModule.CONS}
    Sort Error in Refinement: {VV : [t##X0] | len VV == l##TestModule.CONS}
    Unbound symbol l##TestModule.CONS --- perhaps you meant: TestModule.CONS, is$TestModule.CONS ?
    Just constructor
  |
9 |   CONS :: Int -> Expr t -> Expr [t] -> Expr [t]
  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

facundominguez:
[...]

PR #2370, opened 2024-10-08
fix for #2349

nikivazou:
The refinement type of the wrapper had an extra type variable (following the design of the worker) which made the type inconsistent with the unrefined type. This edit filters out the unused type variables to ensure the generated refinement types is good.


tests/pos/T2349.hs
```diff
@@ -0,0 +1,17 @@
+{-# LANGUAGE GADTs #-}
+{-@ LIQUID "--reflection" @-}
+module TestModule where
```
facundominguez: The test needs to be listed in `tests/tests.cabal`, or it won't run in CI.

facundominguez: Also, it is better to name the module the same as the file.

facundominguez:
```suggestion
-- | Test that the refinement types produced for GADTs are
-- compatible with the Haskell types.
module T2349 where
```

tests/pos/T2349.hs
```diff
@@ -0,0 +1,17 @@
...
+
+  NIL :: Expr [t]
+  CONS :: Int -> Expr t -> Expr [t] -> Expr [t]
+
+{-@
+data Expr t where
+  I :: Int -> Expr Int
+  NIL :: Expr {v:[t] | len v = 0}
+  CONS :: xxxxx:_ -> head:Expr t -> tail:Expr {v:[t] | len v = xxxxx} ->
+          Expr {v:[t] | len v = xxxxx+1}
```
facundominguez: Should the test include a use of CONS? e.g.

```Haskell
{-@
fail e
e :: Expr {v:[Int] | len v = 0}
@-}
e :: Expr [Int]
e = CONS 0 (I 2) NIL
```

nikivazou: Edited! (I added a negative test instead)

facundominguez:
Thanks Niki. I hooked the tests in `tests.cabal`.

What does reflection have to do with the example? I could check that disabling reflection makes the tests fail, but I don't know why.

nikivazou:
I am not sure why reflection is needed to be honest... But I believe it does not affect this PR :)

PR #2372, opened 2024-10-09
No need to freak out when a module has no imports

### LINKED ISSUE #2364: "WARNING: Missing RenamedSource" warning for modules with no imports
It seems LH thinks a module with no imports is some kind of weird error marker, since checking any module without any (implicit or explicit) imports results in a warning:

```
{-# LANGUAGE NoImplicitPrelude #-}
module NoImplicitPrelude where
```

```
[1 of 1] Compiling NoImplicitPrelude ( input/NoImplicitPrelude.hs, input/NoImplicitPrelude.o )
WARNING: Missing RenamedSource
WARNING: Missing RenamedSource

**** LIQUID: SAFE (0 constraints checked) **************************************
```

This message is coming from two places in `Language.Haskell.Liquid.GHC.Interface`, and both of them look like they assume a module with no imports means something got lost in the way:

```
allImports :: [LImportDecl GhcRn] -> S.HashSet Symbol
allImports = \case
  []-> Debug.trace "WARNING: Missing RenamedSource" mempty
  imps -> S.fromList (symbol . unLoc . ideclName . unLoc <$> imps)

qualifiedImports :: [LImportDecl GhcRn] -> QImports
qualifiedImports = \case
  []   -> Debug.trace "WARNING: Missing RenamedSource" (qImports mempty)
  imps -> qImports [ (qn, n) | i         <- imps
                                          , let decl   = unLoc i
                                          , let m      = unLoc (ideclName decl)
                                          , qm        <- maybeToList (unLoc <$> ideclAs decl)
                                          , let [n,qn] = symbol <$> [m, qm]
                                          ]
```

Note that other than LH otherwise seems to work OK on these inputs, for example if we extend the above example, it still has no imports but is checked by the LH plugin correctly:

```
{-# LANGUAGE NoImplicitPrelude #-}
module NoImplicitPrelude where

data Bool = True | False

{-@
embed Bool as bool
assume True  :: {v:Bool | v     }
assume False :: {v:Bool | (~ v) }
@-}

-- Deliberately wrong refinement on `v`'s type, it should be `v || x` of course
{-@ foo :: x:Bool -> { v:Bool | v } @-} 
foo :: Bool -> Bool
foo True = False
foo False = True
```

This is correctly rejected, except for the annoying warning (which can lead to false positives in e.g. CI systems):

```
[1 of 1] Compiling NoImplicitPrelude ( input/NoImplicitPrelude.hs, input/NoImplicitPrelude.o )
WARNING: Missing RenamedSource
WARNING: Missing RenamedSource

input/NoImplicitPrelude.hs:15:12: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : {v : NoImplicitPrelude.Bool | not v
                                         && v == NoImplicitPrelude.False}
    .
      VV : {VV##373 : NoImplicitPrelude.Bool | VV##373}
    .
    Constraint id 7
   |
15 | foo True = False
   |            ^^^^^

```

gergoerdi:
Some modules genuinely don't have any imports, when using `NoImplicitPrelude`. Fixes #2364

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs
@@ -82,8 +82,6 @@ import Language.Haskell.Liquid.Parse
 import Language.Haskell.Liquid.Types hiding (Spec)
 import Language.Haskell.Liquid.UX.Tidy
 
-import qualified Debug.Trace as Debug
-
 
 --------------------------------------------------------------------------------
 -- | Extract Ids ---------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


facundominguez:
LGTM!

PR #2373, opened 2024-10-09
Since `Z3mem` uses an in-proc API instead of calling out of process, there is no point in trying to find its executable

gergoerdi:


facundominguez:
LGTM!

Issue #2369, opened 2024-10-08
addC: malformed constraint


gergoerdi:
Input file:

```
{-# LANGUAGE NoImplicitPrelude #-}
{-# OPTIONS -fplugin=LiquidHaskellBoot #-}
module Proxy where

data Proxy (a :: k) = Proxy

tupleToDict :: p (Proxy cs) -> ()
tupleToDict _ = ()

getMaster :: forall p cs. p cs -> ()
getMaster _ = tupleToDict (\x -> x :: Proxy cs)
```

Panic message:

```
input/Proxy.hs:11:1: error:
    Uh oh.
    addC: malformed constraint:
cconsE:
[...]

PR #2375, opened 2024-10-09
fix for #2369

nikivazou:
Partial function type applications work again, by adding the multiplicity parameter to the fun type constructor.


facundominguez:
The cabal file is enabling -XHaskell2010 which causes the test to fail. Changing the type annotation fixes it, but see if that interferes with what you are trying to test.

nikivazou:
Well, the main edit fixed #2369 locally for me (without the `XHaskell2010`), so when circle approves we can merge? 
Maybe @gergoerdi can confirm that the original example passes?

facundominguez:
I think we can merge.

I was a bit concerned that my type annotation would hide the bug, but I just checked that without your fix the reported error still shows up with my edits.

gergoerdi:
> Maybe @gergoerdi can confirm that the original example passes?

Indeed I can confirm that the original modules that prompted me to report this now work with LH.

Issue #2379, opened 2024-10-11
GHC-9.8 backwards compatibility?


gergoerdi:
I looked at what it would take to still support GHC 9.8 with `develop` as of 5595507d943cb00f92674c19a5cbafcd93dff92b and it's so little that I think it would be feasible with just a little CPP usage to keep supporting 9.8. For comparison, Clash, which is a similarly involved user of the GHC API, [manages to support all GHC versions between 8.10 and 9.10](https://github.com/clash-lang/clash-compiler/tree/5706eafca799ae04fda0ee7d666a40b6c0e7f22b?tab=readme-ov-file#ghc-compatibility). 

- [x] `foldl'` is not exported by the `Prelude`, so has to be imported from `Data.List`. A triviality that affects 8 files. The imports can be guarded by CPP macros if we want to avoid "redundant imports" warnings on GHC 9.10.

- [x] `Unique`s are built from `Int`s, not `Word64`s. Handling this difference can be centralized by adding a type alias and some utility functions to some new  `Liquid.GHC.API.Compat` module.

- [x]  Instead of `GHC.Internal.Foldable` and `GHC.Internal.Real`, we have `Data.Foldable` and `GHC.Real`.  Again, can be worked around by defining aliases in `Liquid.GHC.API.Compat`.

- [x] Differences in AST annotations / TTG expansion points: This is very annoying, because it requires CPP branching at five places across three modules. 

- [x] `tcValBinds` has a slightly different type, requiring a change in `Language.Haskell.Liquid.GHC.Misc.withWiredIn`. 

Overall, I don't think it would be that big a burden to keep GHC 9.8 around.

gergoerdi:
I've pushed the change for the first three issues to https://github.com/gergoerdi/liquidhaskell/tree/ghc-9.8-compat

I've got code locally that "fixes" the remaining problems (by breaking the GHC 9.10 compatibility), so at least we can know there's no other "unknown unknowns" lurking.

facundominguez:
On the side of costs, we would need to make sure patches pass tests in both compiler versions. The hardest thing to support is probably going to be the move of many definitions from `base` to `ghc-internals`, which affects the annotations in the `liquidhaskell` package and wired-in names in the code.

I'd prefer having a separate branch for ghc-9.8 and backporting fixes there when needed, rather than making harder to accept contributions to the `develop` branch. It seems to me like it would allow to better share the maintenance effort.

[...]

PR #2380, opened 2024-10-11
GHC 9.8 backwards compatibility (WIP)

gergoerdi:
- **GHC-9.8 compatibility: `Unique`s contain `Int`s instead of `Word64`s.**
- **GHC-9.8 compatibility: import `foldl'` from `Data.List`**
- **GHC-9.8 compatibility: Different module names for `Foldable` and `Real`**

### DIFF
--- a/liquidhaskell-boot/src-ghc/Liquid/GHC/API/Compat.hs
+++ b/liquidhaskell-boot/src-ghc/Liquid/GHC/API/Compat.hs
@@ -1,0 +1,15 @@
+module Liquid.GHC.API.Compat (
+    UniqueId
+  , toUniqueId
+  ) where
+
+import Data.Word (Word64)
+
+----------------------
+-- Uniques
+----------------------
+
+type UniqueId = Word64
+
+toUniqueId :: Word64 -> UniqueId
+toUniqueId = id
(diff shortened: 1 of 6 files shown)


gergoerdi:
This is basically the "easy" part of #2379. I'd love to hear suggestions on how to handle the "hard" (actual API differences) part.

gergoerdi:
OK so this new version avoids using `CPP` and basically just arranges for a more centralized way of making a separate GHC 9.8 branch.

gergoerdi:
I've removed the draft status because the current two commits are good to go

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Compat.hs
```diff
@@ -0,0 +1,29 @@
+module Liquid.GHC.API.Compat (
```
facundominguez, line 4:
```suggestion
-- | This module contains definitions that change between different versions
-- of the GHC API. It helps encapsulating differences between branches of LH
-- that could support different versions of GHC.
module Liquid.GHC.API.Compat (
```

facundominguez (approved):
LGTM. Thanks @gergoerdi!

Issue #710, opened 2016-05-14
annotating functions in typeclass should result in an error


sorawee:
Please see http://goto.ucsd.edu:8090/index.html#?demo=permalink%2F1463259507_10885.hs

It results in SAFE, while my friends expect it to be UNSAFE.

(Though it's true that if we annotate it with

```
{-@ class Add a b where
    add :: [a] -> [b] -> {v : [a] | len v > 0}
@-}
```

Then it's UNSAFE)

ranjitjhala:
Yikes! @nikivazou and @gridaphobe is this related to the various other type class related issues.

sorawee:
FYI: @avanhatt

Also, it seems she is trying to refine

```
instance (Num k) => Add k k where
    add x y = addVecWrap x y
```

but LH can't even parse `{-@ instance (Num k) => Add k k where ...`

gridaphobe:
I suspect what's going on here is that the isolated spec for `add` is not being attached to the `Add` class the way the class-spec would be.

I think the most user-friendly solution would be to make the two specs @sorawee mentions equivalent, ie

```
class Add a b where
  {-@ add :: [a] -> [b] -> {v : [a] | len v > 0} @-}
  add :: [a] -> [b] -> [a]
```

would effectively desugar into

```
class Add a b where
  add :: [a] -> [b] -> [a]
{-@ class Add a b where
    add :: [a] -> [b] -> {v : [a] | len v > 0}
@-}
```

nikivazou:
@avanhatt, @sorawee actually I recently extended class instances refinements to support the instance you want to refine. So, as of now in develop you can add 


```
instance (Num k) => Add k k where
{-@ instance (Num k) => Add k k where 
	 add :: [k] -> [k] -> {v : [k] | len v > 0}
  @-}

    add x y = addVecWrap x y
 ```

and your code is UNSAFE. 


Though indeed we should create an error for the initial code. I though `add` is not even a variable in the internal representation, so expected the specification to fail with unknown var....

**Comment by gridaphobe** [comment]
@nikivazou `add` is the dictionary selector, and is what appears at callsites, so we definitely need to give it a refined type.

**Comment by avanhatt** [comment]
@nikivazou @gridaphobe @sorawee 

Thank you for the advice! Now that the instance refinement is getting parsed correctly, though, the behavior seems to not be what we expect. 

As a minimal example, the following gives SAFE when we expect UNSAFE: 

```
[...]

PR #2384, opened 2024-10-11
Local rewrites

AlecsFerra:
Depends on [#710/liquid-fixpoint](https://github.com/ucsd-progsys/liquid-fixpoint/pull/710)


AlecsFerra:
For some reason the test `SKIDC.hs` works locally but hangs on circleci

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Env.hs
```diff
@@ -158,6 +159,14 @@ addBind l x r = do
   put          $ st { binds = bs' } { bindSpans = M.insert i l (bindSpans st) }
   return ((x, F.sr_sort r), {- traceShow ("addBind: " ++ showpp x) -} i)
 
+addRewrites :: F.LocalRewrites -> CG ()
+addRewrites rws = do 
+  st <- get
+  let bid = F.bindEnvSize $ binds st
+  put $ st { localRewrites = F.insertRewrites bid rws $ localRewrites st }
```
facundominguez, line 167: Should this be `bid-1`? That would be the id of the last inserted binding at least.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Env.hs
```diff
@@ -158,6 +159,14 @@ addBind l x r = do
...
+  pure ()
```
facundominguez:
```suggestion
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Env.hs
```diff
@@ -158,6 +159,14 @@ addBind l x r = do
   put          $ st { binds = bs' } { bindSpans = M.insert i l (bindSpans st) }
   return ((x, F.sr_sort r), {- traceShow ("addBind: " ++ showpp x) -} i)
 
+addRewrites :: F.LocalRewrites -> CG ()
```
facundominguez:
```suggestion
-- | Add local rewrites associated to the last inserted bindings.
addRewrites :: F.LocalRewrites -> CG ()
```

facundominguez: What if there is no previously inserted binding. Should this function throw an error then?

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,125 @@
...
+        isCtor  name = name `S.member` ctors
+        isLocal name = not (name `S.member` globals 
+                           || name `S.member` ctors
+                           || isPrefixOfSym anfPrefix name)
+
+flatten :: Expr -> [Expr]
+flatten (EApp f e) = flatten f ++ [e]
+flatten e = [e]
+
+groupUnifiableEqualities :: [(Expr, Expr)] -> [(Expr, Expr)]
```
facundominguez, line 73:
```suggestion
-- | Given equalities `[e11=e12, e21=e22, ..., en1=en2]` this function produces
-- equalities `ei2=ej2` whenever `ei1==ej1` and for no other index `k` `ei1==ek1`.
groupUnifiableEqualities :: [(Expr, Expr)] -> [(Expr, Expr)]
```

facundominguez, line 73: Looking to improve the documentation: Why is it ok to ignore the second component when looking for unifiable equalities? e.g. in `C a=x && C b=x` this function is not going to discover `C a=C b`.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,125 @@
...
+        -- elaborated
+        go (EVar s1) e2 | isLocal s1 = Just [(s1, e2)]
+        go e1 (EVar s2) | isLocal s2 = Just [(s2, e1)]
+        -- TODO: Tecnically we could also unify under lambdas but you have to be
+        -- carefull about alpha equivalence idk if the effort is worth it.
+        go e1 e2 = do
+            -- Performing the unification under constructor is safe because 
+            -- C a₁ ... aₙ = C b₁ ... bₙ ⟺ ∀ n . a₁ = bₙ
+            EVar name1 : args1 <- pure $ flatten e1
+            EVar name2 : args2 <- pure $ flatten e2
```
facundominguez: `flatten` is small but quadratic on the amount of arguments, and redundant with `splitEApp`. Why not
```
let (EVar name1, args1) = splitEApp e1
    (EVar name2, args2) = splitEApp e2
```
?

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Env.hs
AlecsFerra: Can you have `-` inside an identifier?

facundominguez: I don't think so, but the suggestion is not to rename bid, but to subtract 1 from the size of the binding environment to get the id of the last inserted bind.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,125 @@
...
+            guard $ name1 == name2
+            guard $ isCtor name1
+            guard $ length args1 == length args2
+            concat <$> zipWithM go args1 args2
```
facundominguez: The code would be somewhat simplified if the return type of `unify` is changed from `Maybe [(Symbol, Expr)]` to just a list `[(Symbol, Expr)]` which should be empty when no unifications are found. In that case, the final statement here turns to
```
            concat $ zipWith go args1 args2
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,125 @@
...
+
+-- | A collection of rewrites that cannot cause an infinite loop
+newtype AcyclicRewrites = AR (M.HashMap Symbol Expr)
+
+-- We can think of the map a directed graph where every symbol is a vertex and
+-- there is an edge v₁ → v₂ if in the expression that v₁ is rewritten to v₂ is a
+-- free variable.  To guarantee that there arent any infinite loops in the
+-- rewrite procedure that will be executed by PLE, we need to esnure that there
+-- aren't any cyclic rewrites, this is ensured by adding the invariant to our
+-- graph of beeing a DAG.
```
facundominguez:
```suggestion
-- We can think of the map as a directed graph where every symbol is a vertex and
-- there is an edge v₁ → v₂ if v₂ is free in the expression that v₁ is rewritten to.
-- To guarantee that the set of rewrite rules is terminating, we ensure that there
-- aren't any cycles in the graph.
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,125 @@
...
+
+-- | Takes a list of rewrites and returns a list of rewrites and filters out any
+-- rewrites that would cause an infinite loop, the procedure is order biased as
+-- rewrites earlier in the list take precedence.
```
facundominguez:
```suggestion
-- | Drops rewrites that would cause an infinite loop. The procedure is order
-- biased as rewrites earlier in the list take precedence.
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,125 @@
...
+unloop :: [(Symbol, Expr)] -> LocalRewrites
+unloop = LocalRewrites . toRewrites . foldl' doInsert empty 
+    where doInsert ar (s, e) = ar `fromMaybe` insert ar s e 
+
+-- | Get the "raw" list of rewrites
+toRewrites :: AcyclicRewrites -> M.HashMap Symbol Expr
+toRewrites (AR m) = m
+
+-- | Checks if there is a path from s1 to s2 O(Σ(e ∈ m) |exprSymbolSet e|) [O(m)
+-- where m is the number of edges]
```
facundominguez:
```suggestion
-- | @existsPth g s1 s2@ yields @True@ checks if there is a path from @s1@ to @s2@
-- in @g@. Time is @O(Σ(e ∈ m) |exprSymbolSet e|)@, or said otherwise, it is @O(m)@
-- where @m@ is the number of edges.
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,125 @@
...
+existsPath :: AcyclicRewrites -> Symbol -> Symbol -> Bool
+existsPath (AR m) s1' s2 = go s1'
+  where
+    -- Since m is a DAG, we can use DFS to check if there is a path from s1 to
+    -- s2 without worrying about infinite loops
```
facundominguez, line 118: Does not change the result, but this might be slower than necessary since when there are multiple paths between two nodes in the DAG, `go` will be redundantly evaluated as many times on the same node.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,125 @@
...
+    go _ = False
+
+-- | Constructs an empty set of rewrites
+empty :: AcyclicRewrites
+empty = AR M.empty
+
+-- | Inserts the rewrite if it wont't cause an infinite loop
+insert :: AcyclicRewrites -> Symbol -> Expr -> Maybe AcyclicRewrites
+insert ar s e | not $ any (existsPath ar s) $ S.toList $ exprSymbolsSet e
+              , not $ s `S.member` exprSymbolsSet e
```
facundominguez: Two changes here:
1. I'm putting the cheapest test first
2. I'm changing the direction in which the path is searched. Otherwise, cycles are not detected.
```suggestion
insert ar s e | not $ s `S.member` exprSymbolsSet e
              , not $ any (\s2 -> existsPath ar s2 s) $ S.toList $ exprSymbolsSet e
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,125 @@
...
+              -- There are two ways to break the DAG invariant
+              -- 1. If the rewrite is closing a loop
+              -- 2. If the rewrite by itself is a cycle
```
facundominguez, line 134: `existsPath` detects both conditions. Is it necessary to test for (2) separately?

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,125 @@
...
+    go s1 | Just e <- M.lookup s1 m
+          = any go $ S.toList $ exprSymbolsSet e
+    go _ = False
+
+-- | Constructs an empty set of rewrites
+empty :: AcyclicRewrites
+empty = AR M.empty
+
+-- | Inserts the rewrite if it wont't cause an infinite loop
+insert :: AcyclicRewrites -> Symbol -> Expr -> Maybe AcyclicRewrites
```
facundominguez, line 129: I would just return the graph unmodified when a cycle would be introduced, instead of returning a Maybe.

facundominguez (commented):
The overall strategy looks good. I think there might be a bug in the contruction of the DAG. I also proposed edits to the code and the documentation.

Also, can we have the new feature explained in the flag reference. Could be a good time to explain `--etabeta` as well (I forgot to flag that at the time it was contributed :facepalm:).

facundominguez:
I forgot to add, all the calls of `HashSet.toList` are making the code non-deterministic because the order in which the elements are returned might vary from one machine to another. It will depend on what the hash of a particular expression is in that machine.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Env.hs
AlecsFerra: Yeah i think so, where the function is actually called there is always at least one binding

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Env.hs
AlecsFerra: But then we need to add 1 again

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: The unification Just [] is different from Nothing, Nothing means the unification failed so we bail completely, instead Just [] means the unification succeeded but we didn't discovered any equality

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: I based it on the fact that usually we don't have interesting terms on the left, but in general we could also handle those cases

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: exists path is only looking for paths that are already in the path if we have the isolated node x and some new equation x = id x it wont discover the loop

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: I think that it could be interesting to know if the addition of some rewrite could cause a loop (ex. inform a user that is debugging the code) even if now we aren't using it now

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: I forgot it existed, if it's equivalent it's ok

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: I dont get what you are suggesting

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Env.hs
facundominguez: Evidently, I don't get yet what this code is meant to do. So let me ask, why would we need to add 1 to the id of the last inserted binding?

AlecsFerra: Because first we are adding the rewrite and then the binding, the local rewrites are attached to the binding inserted at line 878

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
facundominguez: My bad.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
facundominguez: Say the graph is
```
a -> b -> c -> d -> e
 `----->-´
```
So there is an edge between `a` and `c`.

`existsPath g 'f' 'a'` is going to visit node `c` twice, first through `a` and then through `b`. Which means that the subgraph `c -> d -> e` is also going to be scanned twice in search for `f`.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
facundominguez: The only use of `unify` doesn't distinguish between those two cases. But I'm guessing you mean to leave it as it is in case it makes a difference in the future. That would be fine by me.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: Yo mean that we should do a BFS instead of a DFS?

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: zipWithM should collapse to Nothing if one of the unification is returning Nothing

AlecsFerra:
> I think there might be a bug in the contruction of the DAG.

Where?

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
facundominguez: I guess that if you want to scan each node at most once, then a set of the already scanned vertices should be computed. The search strategy doesn't make a difference, I believe.

AlecsFerra:
> I forgot to add, all the calls of `HashSet.toList` are making the code non-deterministic because the order in which the elements are returned might vary from one machine to another. It will depend on what the hash of a particular expression is in that machine.

Makes sense this could explain the different behavior on the CI

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: Ahh now I got it. Yeah we could do that

facundominguez:
>> I think there might be a bug in the contruction of the DAG.
>
> Where?

You already merged the suggestion for a fix [here](https://github.com/ucsd-progsys/liquidhaskell/pull/2384#discussion_r1803622668).

AlecsFerra:
> > > I think there might be a bug in the contruction of the DAG.
> > 
> > 
> > Where?
> 
> You already merged the suggestion for a fix [here](https://github.com/ucsd-progsys/liquidhaskell/pull/2384#discussion_r1803622668).

Ah you're right and indeed it was a bug

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
facundominguez: Ok, so `C (f a) (g x)` does not unify with `C (h a) y`. I would think we could use `y=g x` despite the failed unification of the first argument.

AlecsFerra: Uhm technically yes I don't know why I wrote it like that

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Env.hs
```diff
@@ -158,6 +159,14 @@ addBind l x r = do
   put          $ st { binds = bs' } { bindSpans = M.insert i l (bindSpans st) }
   return ((x, F.sr_sort r), {- traceShow ("addBind: " ++ showpp x) -} i)
 
+-- | Add local rewrites associated to the last inserted bindings.
```
facundominguez: Given my new understanding of this function:
```suggestion
-- | Add local rewrites associated to the next binding that will be inserted.
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: But i think is something that we should probably benchmark as usually those graphs are reasonably small and using an hash-set could be a slowdown

AlecsFerra:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

AlecsFerra:
The only open question is the optimization in the DFS

.circleci/config.yml
```diff
@@ -9,13 +9,13 @@ commands:
...
-            sudo cp z3-4.8.7-x64-ubuntu-16.04/bin/z3 /usr/local/bin
-            sudo cp z3-4.8.7-x64-ubuntu-16.04/include/* /usr/local/include
-            rm -rf z3-4.8.7-x64-ubuntu-16.04
+            wget https://github.com/Z3Prover/z3/releases/download/z3-4.13.3/z3-4.13.3-x64-glibc-2.35.zip
+            unzip z3-4.13.3-x64-glibc-2.35.zip
+            rm -f z3-4.13.3-x64-glibc-2.35.zip
+            sudo cp z3-4.13.3-x64-glibc-2.35/bin/libz3.a /usr/local/lib
+            sudo cp z3-4.13.3-x64-glibc-2.35/bin/z3 /usr/local/bin
+            sudo cp z3-4.13.3-x64-glibc-2.35/include/* /usr/local/include
+            rm -rf z3-4.13.3-x64-glibc-2.35
```
facundominguez: We should add some check for this version of Z3. Otherwise, tests are going to fail for some users, and they will be clueless about why. [Liquid.Fixpoint.Smt.Interface](https://github.com/ucsd-progsys/liquid-fixpoint/blob/a7e09a68e90b81fbc428d2acf655b25a21298584/src/Language/Fixpoint/Smt/Interface.hs#L332) does version checks already.

Then it would be fine to skip the version check with a flag `--skip-smt-version-check` if the user wants to run an older version anyway.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,127 @@
...
+        ctors   = toSet $ seBinds  $ constEnv γ
+        -- All the global names (top level functions, types, etc...)
+        globals = toSet $ reGlobal $ renv     γ
+    in unloop
+       $ concatMap (uncurry $ unify ctors globals)
+       $ groupUnifiableEqualities 
+       $ getEqualities refinement
+    where toSet = S.fromList . M.keys
+
+-- We don't need the a total unification as constructors are inejctive
```
facundominguez:
```suggestion
-- We don't need a total unification as constructors are injective
```

facundominguez: What would be total unification?

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,127 @@
...
+-- equalities `ei2=ej2` whenever `ei1==ej1` or and for no other index `k` `ei1==ek1`
+-- or in in the other direction
+groupUnifiableEqualities :: [(Expr, Expr)] -> [(Expr, Expr)]
+groupUnifiableEqualities = mapMaybe keep2 . grouping
+    where -- We perform only 2-way unification
+          keep2 [e1, e2] = Just (e1, e2)
+          keep2 _        = Nothing
+         
+          grouping eqs = fmap snd $ M.groupList eqs 
+                                 ++ M.groupList (fmap swap eqs)
```
facundominguez: Idea for a more relaxed implementation:
1. A single call to groupList can mix swapped and non-swap equations
2. Matches are not discarded if there are more than two.

```suggestion
groupUnifiableEqualities :: [(Expr, Expr)] -> [(Expr, Expr)]
groupUnifiableEqualities = concatMap mkEqs . grouping
    where 
          mkEqs (e1: es) = [ (e1, e) | e <- es ]
          mkEqs _        = []
         
          grouping eqs = fmap snd $ M.groupList $ eqs ++ fmap swap eqs
```

facundominguez (commented):
Thanks @AlecsFerra. I left some comments. Let me know what you think.

Also, some user documentation would be deserved for this PR and the `--etabeta` flag.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: That unification success only if we can unify everything

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,138 @@
...
+getCaseRewrites γ spec =
+    let Reft (_, refinement) = ur_reft $ rt_reft spec
+        -- Names of constustors both type constructors and data constructors
+        ctors   = toSet $ seBinds  $ constEnv γ
+        -- All the global names (top level functions, types, etc...)
+        globals = toSet $ reGlobal $ renv     γ
+    in unloop
+       $ concatMap (uncurry $ unify ctors globals)
+       $ groupUnifiableEqualities 
+       $ getEqualities refinement
```
facundominguez, line 27: I'm realizing that if the pipeline is fed a single equality `f = \x -> x` with `f` local, it is not going to generate a rewrite for `f`. While I realize this is not the purpose of `getCaseRewrites`, I'm left wondering how to make a rewrite for it. Is this future work?

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
```diff
@@ -0,0 +1,144 @@
...
+
+-- | Given a list of equalities this function produces the equalities that
+-- result from applying transitivity exactly once through terms that are not
+-- unifiable. For instance, if we have @[e1=e2, e2=e3, e1=e4]@ this function
+-- will produce @[e1=e3, e2=e4]@.
+--
+-- Some equalities are not produced if more than two equalities refer to the
+-- same expression. For instance, @[e1=e2, e1=e3, e1=e4]@, becase @e1@ appears
+-- three times, the equality @e3=e4@ won't be produced. This is an heuristic
+-- to avoid producing equalities that can later produce redundant rewrites.
```
facundominguez: Let me know if this is not accurate. `groupUnifiableEqualities` would be easier to explain if it returned all pairs of equalities when `e1` appears more than twice.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: We could add that case is not going to create any problem in principle, it's the same thing as get equality being right biased it was thought as a way to extract equalities when performing dependant pattern matching on dependant inductive indexed types and in LH we generate constraint for those in a standard format (`prop e = T e1 ... en`) but I believe that in spirit of making the code more general as we did by removing the right bias we could also handle those cases

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
AlecsFerra: Sorry I don't understand you mean `[(Expr, [Expr])]`?

.circleci/config.yml
AlecsFerra: I believe that was just a z3 bug so i don't know if it makes sense signaling to the user, and if we decide to do that we should probably check exactly from which version it's not going to loop forever, on Niki's pc the version was just slightly older than the CI's one

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/RewriteCase.hs
facundominguez: I think the following code would honor the shorter description exactly.
```diff
 -- | Given a list of equalities this function produces the equalities that
 -- result from applying transitivity exactly once. For instance, if we have
 -- @[e1=e2, e2=e3, e1=e4]@ this function will produce @[e1=e3, e2=e4]@.
---                                                                                                                  
--- Some equalities are not produced if more than two equalities refer to the
--- same expression. For instance, @[e1=e2, e1=e3, e1=e4]@, becase @e1@ appears
--- three times, the equality @e3=e4@ won't be produced. This is an heuristic
--- to avoid producing equalities that can later produce redundant rewrites.
 groupUnifiableEqualities :: [(Expr, Expr)] -> [(Expr, Expr)]
-groupUnifiableEqualities = concatMap mkEqs . grouping
+groupUnifiableEqualities = concat . concatMap mkEqs . grouping
   where                                                                                                                                                                                       
-    mkEqs (e1:es) = [ (e1, e) | e <- es ]
+    mkEqs (e1:es) = [ (e1, e) | e <- es ] : mkEqs es
     mkEqs _       = []

     grouping eqs = fmap snd $ M.groupList $ eqs ++ fmap swap eqs
```

.circleci/config.yml
facundominguez: I'm assuming it is a z3 bug, indeed. I think you can drop this if you prefer. We can have an issue specifically for checking the z3 version.

My local version of z3 is 4.8.17.

.circleci/config.yml
```diff
@@ -9,13 +9,13 @@ commands:
...
-            sudo cp z3-4.13.3-x64-glibc-2.35/bin/z3 /usr/local/bin
-            sudo cp z3-4.13.3-x64-glibc-2.35/include/* /usr/local/include
-            rm -rf z3-4.13.3-x64-glibc-2.35
+            wget https://github.com/Z3Prover/z3/releases/download/z3-4.8.17/z3-4.8.17-x64-glibc-2.31.zip
+            unzip   z3-4.8.17-x64-glibc-2.31.zip
+            rm -f   z3-4.8.17-x64-glibc-2.31.zip
+            sudo cp z3-4.8.17-x64-glibc-2.31/bin/libz3.a /usr/local/lib
+            sudo cp z3-4.8.17-x64-glibc-2.31/bin/z3 /usr/local/bin
+            sudo cp z3-4.8.17-x64-glibc-2.31/include/* /usr/local/include
+            rm -rf  z3-4.8.17-x64-glibc-2.31
```
AlecsFerra: Maybe we should refactor the name in a variable

.circleci/config.yml
AlecsFerra: I think an issue would be better

docs/mkDocs/docs/options.md
```diff
@@ -157,6 +157,54 @@ myIsDigit x = '0' <= x && x <= '9'
...
+
+{-@ beta :: { (\x:Int -> x) (12) = 12 } @-}
+beta :: Proof
+beta = trivial
+```
+
+This also allows PLE to unfold partially applied reflected functions.
+
+Note: The eta-expansion rule subsumes the `--extensionality` flag. To save time
+during typechecking, you can disable `--extensionality` when using `--etabeta`.
```
facundominguez: I checked if enabling --etabeta would have the effect of enabling extensionality, but this is not the case since there is [one place](https://github.com/ucsd-progsys/liquidhaskell/blob/d90769a9cbf7641881e8a62500a7a6fc0fa32aff/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs#L506) in LH where the extensionality flag is used and etabeta is ignored.

I don't know what it is meant by the first sentence: "The eta-expansion rule subsumes the `--extensionality` flag".

> To save time during typechecking,

Do you mean that extensionality makes etabeta noticeably slower?

AlecsFerra: It's meant as any theorem that is proved automatically by extensionality should also be automatically proved by eta beta, in practice the extensionality flag works this way: 

If there is some proof `{ e1 = e2 }` where `e1` and `e2` have type `T_1 -> ... -> T_n -> T` the extensionality flag rewrites the goal of the proof as `{ e1 x1 ... xn = e2 x1 ... xn }` adding some "ghost" variables `x1 ... xn` to give a chance to ple to unfold any partially applied function in `e1` and `e2`.

But now the unfolding of partially applied functions is also triggered by eta-expansion, actually is a more general mechanism as eta it works also when functions are used as an argument to some other function/constructor as eta expansion is type preserving but extensionality is not as adding the variables will change the type of the expression so adding the ghost variables there will create an ill typed term.

The suggestion to disable the extensionality flag is meant as: the extensionality flag is not going to automate any proof that is not already automated by etabeta so by enabling extensionality you are executing more code for nothing, is also not going to make PLE faster as we eta-expand everything also fully applied functions.

AlecsFerra: But we can easily not mention this if you think is only going to confuse the users

facundominguez: I made the note more explicit in e6dc9ac.

What do you think of [that use](https://github.com/ucsd-progsys/liquidhaskell/blob/d90769a9cbf7641881e8a62500a7a6fc0fa32aff/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs#L506) of the extensionality flag in LH?

Should that place take into account etabeta as well?

AlecsFerra: I think that code is needed to take into account the fact that extensionality flag is changing the type of some expressions, thing that is not done by doing eta expansion, also the eta expansion is totally transparent to LH as it's actually done completely in LF so it shouldn't require any modification in LH

facundominguez (approved):
Hey, it seems to me we are good to merge! :tada:

PR #2385, opened 2024-10-11
Run the plugin at most once without error with multiple -fplugin=LiquidHaskell flags

facundominguez:


facundominguez:
cc @AlecsFerra @gergoerdi

PR #2387, opened 2024-10-14
Improve GHC API usage

gergoerdi:
This removes a ton of cruft by making better use of facilities already provided by the GHC API.


facundominguez (approved):
LGTM! Thanks @gergoerdi!

Issue #2169, opened 2023-04-25
Resolve names only once


facundominguez:
#2303 adds a minimal test showing the problem.

Currently, names in specifications are resolved multiple times. Name resolution determines for each name in a spec, what is the Haskell definition it is associated to. Type and data constructors exist both in the logic and in Haskell, and signatures with refinement types are also linked to Haskell definitions. 

These names are resolved once when compiling the module containing the specs, and they are resolved again when compiling a module that transitively imports the module containing the specs. This is not ideal for the duplicated work, but it is also problematic because the implementation doesn't ensure that names would resolve in the same way on all of the attempts.

Resolving names requires looking up an identifier in an environment storing all the things that GHC knows to be in scope. When resolving names for transitive dependencies, a single environment is used to resolve the names of all of the transitively imported modules. In order to disambiguate names that could refer to one of many definitions in Haskell, the resolution algorithm considers whether the name has a qualifier, and what is the name in which the spec has been defined. But this strategy not always yields the same resolution as obtained when compiling the imported module, which uses, understandably, a smaller environment to resolve the names.

This issue is about making name resolution more predictable. The current workaround is to manually fully qualify names in specifications, although there are restrictions to do this when it comes to data specifications.

Part of the solution would likely be storing specifications in interface files after their names have been fully qualified with the module of origin of the thing referred by the name.

Another related concern is that variable names should also include the package of origin, in case multiple packages define modules with the same names.

There is work in progress in [liquidhaskell](https://github.com/ucsd-progsys/liquidhaskell/compare/fd/parameterize-spectypes) and [liquid-fixpoint](https://github.com/ucsd-progsys/liquid-fixpoint/pull/708).

kd1729:
Hi @facundominguez, I would like to take up this idea for GSOC. I am making the proposal, how can I get it reviewed?

facundominguez:
Hello @kd1729! You can email it to [me](https://github.com/facundominguez). I'm also reachable via the liquidhaskell [slack](https://github.com/ucsd-progsys/liquidhaskell?tab=readme-ov-file#ask-for-help).

ninioArtillero:
Greetings @facundominguez , I'm also interested in the idea! Would get in touch :)

facundominguez:
A little bit of analysis follows.

[...]

PR #2388, opened 2024-10-14
Towards a direct (non plugin-based) API

gergoerdi:
For users that need to do custom steps on the way to a full `TypecheckedModule`, it'd make sense to provide an API that directly consumes the `TypecheckedModule`.

### DIFF
--- a/liquidhaskell-boot/src-ghc/Liquid/GHC/API.hs
+++ b/liquidhaskell-boot/src-ghc/Liquid/GHC/API.hs
@@ -37,6 +37,7 @@ import           GHC                  as Ghc
     , GhcException(CmdLineError, ProgramError)
     , GhcLink(LinkInMemory)
     , GhcMode(CompManager)
+    , GhcMonad
     , GhcPs
     , GhcRn
     , HsDecl(SigD)
@@ -111,6 +112,7 @@ import           GHC                  as Ghc
     , isRecordSelector
     , isTypeSynonymTyCon
     , isVanillaDataCon
+    , lookupName
     , mkHsApp
     , mkHsDictLet
     , mkHsForAllInvisTele
@@ -414,9 +416,6 @@ import GHC.Driver.Config.Diagnostic as Ghc
     ( initDiagOpts
     , initDsMessageOpts
     , initIfaceMessageOpts
-    )
-import GHC.Driver.Main                as Ghc
-    ( hscTcRcLookupName
     )
 import GHC.Driver.Plugins             as Ghc
     ( ParsedResult(..)
--- a/liquidhaskell-boot/src-ghc/Liquid/GHC/API.hs
+++ b/liquidhaskell-boot/src-ghc/Liquid/GHC/API.hs
@@ -23,7 +23,7 @@ import           GHC                  as Ghc
     ( Class
     , DataCon
     , DesugaredModule(DesugaredModule, dm_typechecked_module, dm_core_module)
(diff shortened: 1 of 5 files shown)


facundominguez:
The title of the PR intrigues me, but the changes in the last commit look ok.

facundominguez:
Regarding the `unoptimiseDynFlags` setting, I think it is possible that we can get rid of it. 

When I remove it, many failures seem to be related to the fact that `tcg_exports` does not contain all identifiers that are in scope anymore. For instance, LH complains that `bar` is not in scope in the following test

```Haskell
module LocalSpec() where

import Language.Haskell.Liquid.Prelude (choose)

prop = if x > 0 then bar x else x
  where x = choose 0
    {-@ bar :: Nat -> Nat @-}
        bar :: Int -> Int
        bar x = x

{-@ bar :: a -> {v:Int | v = 9} @-}
bar :: a -> Int
bar _ = 9
```

but if I open the export list, then verification passes again
```diff
-module LocalSpec() where
+module LocalSpec where
```

Fixing this might require building LH environment from names in scope rather than `tcg_exports` (perhaps `tcg_rdr_env`).

Update: Looking up for TyThings for everything in `tcg_rdr_env` seems too slow. We might need to rearrange things to only lookup TyThings that are actually needed. This connects a bit to #2169.

gergoerdi:
> The title of the PR intrigues me, but the changes in the last commit look ok.

Basically, I'm thinking of a "bring-your-own-`TypecheckedModule`" API for users that are not GHC itself, but rather programs using the GHC API and using any kind of out-of-band inter-phase communication. 

For example, in that setting, "the" parsed module as seen by a plugin (LH or otherwise) might not actually be the right module if the API client needs to do some postprocessing on it.

gergoerdi:
> Regarding the `unoptimiseDynFlags` setting, I think it is possible that we can get rid of it.

Indeed, the only part of `unoptimiseDynFlags` that should matter anymore is setting the backend to `interpreterBackend`. Relying on `tcg_exports` to find out what is in a module is the wrong approach anyway, so I hope you'll be able to remove that. Meanwhile I'll remove all other `DynFlags` fiddling in an update to #2389.

gergoerdi:
> Fixing this might require building LH environment from names in scope rather than `tcg_exports` (perhaps `tcg_rdr_env`).
> 
> Update: Looking up for TyThings for everything in `tcg_rdr_env` seems too slow. We might need to rearrange things to only lookup TyThings that are actually needed. This connects a bit to #2169.

Could you point me to where exactly this is? Do you mean `availVars` / `availTyCons` as computed in `mkPipelineData`? 

Whether you should use `tcg_exports` or not depends on why you need these `Id`s and `TyCon`s. If you need them to check them, then it should be completely divorced from questions of exporting -- imagine if GHC only typechecked definitions that are exported...! You should only care about `tcg_exports` (or, actually, `mg_exports` since we should move as much as we can to only consuming the `ModGuts`, in anticipation of a glorious future where we can plug into GHC's desugared output directly instead of starting from a `TcGblEnv` and doing it ourselves) if you need to make decisions based on whether a given thing is exported or not. 

Also, if you need all this so you can match user-written occurrences in LH specs to the GHC side, you shouldn't need to build anything from the `mg_rdr_env :: GlobalRdrEnv`, because that *is* already the thing you can use to look up names in!

facundominguez:
> Could you point me to where exactly this is? Do you mean availVars / availTyCons as computed in mkPipelineData?

That's the place.

Issue #2257, opened 2024-01-24
Innocent edit causes tests/pos/Map0.hs to fail


facundominguez:
`tests/pos/Map0.hs` has a function
```Haskell
deleteFindMax t
  = case t of
      Bin _ k x l Tip -> (k, x, l)
      Bin _ k x l r -> let (km3, vm, rm) = deleteFindMax r in (km3, vm, (balance k x l rm))
      Tip           -> (error ms, error ms, Tip)
  where ms = "Map.deleteFindMax : can not return the maximal element of an empty Map"
```
If we rewrite this function to
```Haskell
deleteFindMax t
  = case t of
      Bin _ k x l Tip -> (k, x, l)
      Bin _ k x l r -> let ds = deleteFindMax r in
        (case ds of { (km3, vm, rm) -> km3 }
        , case ds of { (km3, vm, rm) -> vm }
        , balance k x l (case ds of { (km3, vm, rm) -> rm })
        )

      Tip           -> (error ms, error ms, Tip)
  where ms = "Map.deleteFindMax : can not return the maximal element of an empty Map"
```
then Liquid Haskell produces errors like,
```
**** LIQUID: UNSAFE ************************************************************

tests/pos/Map0.hs:53:22: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : a
    .
    is not a subtype of the required type
      VV : {VV : a | VV < ?a}
    .
    in the context
      ?a : a
    Constraint id 63
   |
53 |                LT -> balance kx x (delete k l) r
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/pos/Map0.hs:54:22: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : a
    .
    is not a subtype of the required type
      VV : {VV : a | VV > ?a}
    .
    in the context
      ?a : a
    Constraint id 19
   |
54 |                GT -> balance kx x l (delete k r)
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/pos/Map0.hs:55:22: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : a
    .
    is not a subtype of the required type
      VV : {VV : a | VV > ?a}
    .
    in the context
      ?a : a
    Constraint id 45
   |
55 |                EQ -> glue kx l r
   |                      ^^^^^^^^^^^
```

This bug is relevant because it happens when attempting to upgrade to GHC 9.8.1, where
[...]

PR #2389, opened 2024-10-16
Avoid re-typechecking 🚀╰(*°▽°*)╯

gergoerdi:
This finally solves the long-standing problem of the LH plugin running the GHC typechecker. Note that this works *without requiring anything new from GHC*.

I guarantee you this will be the best PR you get from me this week :P

### DIFF
--- a/liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
+++ b/liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
@@ -177,46 +177,50 @@ apiCommentsParsedSource ps =
 -- https://gitlab.haskell.org/ghc/ghc/-/issues/24386
 --
 addNoInlinePragmasToBinds :: TcGblEnv -> TcGblEnv
-addNoInlinePragmasToBinds tcg = tcg{ tcg_binds = go False (tcg_binds tcg) }
+addNoInlinePragmasToBinds tcg = tcg{ tcg_binds = go (tcg_binds tcg) }
   where
-    go :: forall a. Data a => Bool -> a -> a
-    go isAbs = gmapT $ go isAbs `extT` markHsBind isAbs
+    go :: forall a. Data a => a -> a
+    go = gmapT $ go `extT` markHsBind
 
     -- Mark all user-originating `Id` binders as `NOINLINE`.
-    markHsBind :: Bool -> HsBind GhcTc -> HsBind GhcTc
-    markHsBind isAbs = \case
-        bind@VarBind{ var_id = var, var_rhs = rhs } -> bind{ var_id = markId isAbs var, var_rhs = go False rhs }
-        bind@FunBind{ fun_id = var, fun_matches = matches } -> bind{ fun_id = markId isAbs <$> var, fun_matches = go False matches }
-        bind@PatBind{ pat_lhs = lhs, pat_rhs = rhs } -> bind{ pat_lhs = markPat isAbs <$> lhs, pat_rhs = go False rhs }
+    markHsBind :: HsBind GhcTc -> HsBind GhcTc
+    markHsBind = \case
+        bind@VarBind{ var_id = var, var_rhs = rhs } -> bind{ var_id = markId var, var_rhs = go rhs }
+        bind@FunBind{ fun_id = var, fun_matches = matches } -> bind{ fun_id = markId <$> var, fun_matches = go matches }
+        bind@PatBind{ pat_lhs = lhs, pat_rhs = rhs } -> bind{ pat_lhs = markPat <$> lhs, pat_rhs = go rhs }
         PatSynBind{} -> error "markNoInline: unexpected PatSynBind, should have been eliminated by the typechecker"
+        XHsBindsLR absBinds -> XHsBindsLR (markAbsBinds absBinds)
 
-        -- The AbsBinds come from the GHC typechecker to handle
-        -- polymorphism, overloading, and recursion, so those don't
-        -- correspond directly to user-written `Id`s. In fact, marking
-        -- them as `NOINLINE` results in Core that LH can't process.
-        XHsBindsLR absBind@AbsBinds{ abs_binds = binds
-                                   , abs_exports = abes } -> XHsBindsLR $ absBind
-          { abs_binds = go True binds
-          , abs_exports = map markABE abes }
+    markPat :: Pat GhcTc -> Pat GhcTc
+    markPat = \case
+        VarPat ext var -> VarPat ext (markId <$> var)
+        pat -> gmapT (id `extT` markPat) pat
 
-    markABE :: ABExport -> ABExport
-    markABE abe@ABE{ abe_poly = poly
-                   , abe_mono = mono } = abe
-      { abe_poly = markId False poly
-      , abe_mono = markId False mono }
+    markId :: Id -> Id
+    markId var = var `setInlinePragma` neverInlinePragma
 
-    markPat :: Bool -> Pat GhcTc -> Pat GhcTc
-    markPat isAbs = \case
-        VarPat ext var -> VarPat ext (markId isAbs <$> var)
-        pat -> gmapT (id `extT` markPat isAbs) pat
+    -- The AbsBinds come from the GHC typechecker to handle polymorphism,
+    -- overloading, and recursion, so those don't correspond directly to
+    -- user-written `Id`s. In fact, marking them as `NOINLINE` results in Core
+    -- that LH can't process.
+    markAbsBinds :: AbsBinds -> AbsBinds
+    markAbsBinds absBinds0 =
+        absBinds0
+          { abs_binds = fmap (fmap skipFirstHsBind) $ abs_binds absBinds0
+          , abs_exports = map markABE (abs_exports absBinds0)
+          }
+      where
+        skipFirstHsBind :: HsBind GhcTc -> HsBind GhcTc
+        skipFirstHsBind = \case
+          XHsBindsLR absBinds -> XHsBindsLR (markAbsBinds absBinds)
+          b -> gmapT go b
 
-    markId :: Bool -> Id -> Id
-    markId isAbs var
-      | isAbs
-      = var
+        markABE :: ABExport -> ABExport
+        markABE abe@ABE{ abe_poly = poly
+                       , abe_mono = mono } = abe
+          { abe_poly = markId poly
+          , abe_mono = markId mono }
 
-      | otherwise
-      = var `setInlinePragma` neverInlinePragma
 
 lookupModSummary :: HscEnv -> ModuleName -> Maybe ModSummary
 lookupModSummary hscEnv mdl = do
--- a/liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
+++ b/liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
@@ -208,10 +208,10 @@ addNoInlinePragmasToBinds tcg = tcg{ tcg_binds = go (tcg_binds tcg) }
     -- https://github.com/ucsd-progsys/liquidhaskell/issues/2257 for more
     -- context.
     markAbsBinds :: AbsBinds -> AbsBinds
-    markAbsBinds absBinds0 =
+    markAbsBinds absBinds0@AbsBinds{ abs_binds = binds, abs_exports = exports }  =
         absBinds0
-          { abs_binds = fmap skipFirstHsBind <$> abs_binds absBinds0
-          , abs_exports = map markABE (abs_exports absBinds0)
+          { abs_binds = fmap skipFirstHsBind <$> binds
+          , abs_exports = map markABE exports
           }
       where
         skipFirstHsBind :: HsBind GhcTc -> HsBind GhcTc
--- a/liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
+++ b/liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
@@ -147,8 +147,8 @@ data ApiComment
   deriving (Eq, Show)
 
 -- | Extract top-level comments from a module.
-apiComments :: ParsedModule -> [Ghc.Located ApiComment]
-apiComments pm = apiCommentsParsedSource (pm_parsed_source pm)
+apiComments :: HsParsedModule -> [Ghc.Located ApiComment]
+apiComments = apiCommentsParsedSource . hpm_module
 
 apiCommentsParsedSource :: Located (HsModule GhcPs) -> [Ghc.Located ApiComment]
 apiCommentsParsedSource ps =
--- a/liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
+++ b/liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
@@ -6,7 +6,7 @@
 module Liquid.GHC.API.Extra (
     module StableModule
   , ApiComment(..)
-  , addNoInlinePragmasToLocalBinds
+  , addNoInlinePragmasToBinds
   , apiComments
   , apiCommentsParsedSource
   , dataConSig
@@ -44,7 +44,6 @@ import GHC.Core.Coercion              as Ghc
 import GHC.Core.DataCon               as Ghc
 import GHC.Core.Make                  (pAT_ERROR_ID)
 import GHC.Core.Type                  as Ghc hiding (typeKind , isPredTy, extendCvSubst, linear)
-import GHC.Data.Bag                   (bagToList)
 import GHC.Data.FastString            as Ghc
 import GHC.Data.Maybe
 import qualified GHC.Data.Strict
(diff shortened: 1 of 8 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -276,82 +281,75 @@ parsedHook _ ms parsedResult = do
 --    if we call /LH/ any later than here;
 --
 -- 2. Although /LH/ works on \"Core\", it requires the _unoptimised_ \"Core\" that we
---    grab from parsing (again) the module by using the GHC API, so we are really
---    independent from the \"normal\" compilation pipeline.
+--    grab from desugaring a postprocessed versin of the typechecked module, so we are
```
hsyl20:
```suggestion
--    grab from desugaring a postprocessed version of the typechecked module, so we are
```

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
```diff
@@ -170,46 +170,43 @@ apiCommentsParsedSource ps =
...
+    markABE :: ABExport -> ABExport
+    markABE abe@ABE{ abe_poly = poly
+                   , abe_mono = mono } = abe
+      { abe_poly = markId False poly
+      , abe_mono = markId False mono }
+
+    markId :: Bool -> Id -> Id
+    markId isAbs var
+      | isAbs
+      = var
```
facundominguez: I think I'd need some comment about why ids in `AbsBinds` need to be left alone.

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
```diff
@@ -170,46 +170,43 @@ apiCommentsParsedSource ps =
...
-
+    go :: forall a. Data a => Bool -> a -> a
+    go isAbs = gmapT $ go isAbs `extT` markHsBind isAbs
+
+    markHsBind :: Bool -> HsBind GhcTc -> HsBind GhcTc
+    markHsBind isAbs = \case
+        VarBind ext var rhs -> VarBind ext (markId isAbs var) (go False rhs)
+        FunBind ext var matches -> FunBind ext (markId isAbs <$> var) (go False matches)
+        PatBind ext lhs mult rhs -> PatBind ext lhs mult (go False rhs)
+        PatSynBind{} -> error "markNoInline: PatSynBind"
```
facundominguez: How is this error case avoided when the module has pattern synonyms?

And why can't patterns synonyms be ignored?

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
```diff
@@ -170,46 +170,43 @@ apiCommentsParsedSource ps =
...
-      bnds -> bnds
-
+    go :: forall a. Data a => Bool -> a -> a
+    go isAbs = gmapT $ go isAbs `extT` markHsBind isAbs
+
+    markHsBind :: Bool -> HsBind GhcTc -> HsBind GhcTc
+    markHsBind isAbs = \case
+        VarBind ext var rhs -> VarBind ext (markId isAbs var) (go False rhs)
+        FunBind ext var matches -> FunBind ext (markId isAbs <$> var) (go False matches)
+        PatBind ext lhs mult rhs -> PatBind ext lhs mult (go False rhs)
```
facundominguez: Why are the ids bound by the lhs not marked?

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -153,28 +152,28 @@ plugin = GHC.defaultPlugin {
...
 --
--- The @typeCheckResultAction@ sets the breadcrumb to @Typechecking@ on entry,
--- and clears the breadcrumb on exit.
+-- The @typeCheckResultAction@ clears the breadcrumb on entry.
 --
 -- If the @typeCheckResultAction@ does not find the breadcrumb, it skips the
 -- module, assuming that it has been verified already. This could happen if
 -- the plugin is activated more than once on the same module (by passing
 -- multiple times @-fplugin=LiquidHaskell@ to GHC).
+{- HLINT ignore Breadcrumb "Use newtype instead of data" -}
```
facundominguez, line 176: Probably a comment is deserved here to explain why `data` is preferred to `newtype` in this case.

facundominguez (commented):
Thanks @gergoerdi! Quite an achievement. I've sprinkled some questions in parts that I don't fully understand.

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
gergoerdi: "Because that is what matches the old behaviour" is the only answer I can give you. You will have to answer that yourself -- why did you patch the module before typechecking exactly the way you did it?

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
gergoerdi: The GHC desugarer already errors out on `PatSyBind`s -- they're not supposed to be here, since the typechecker replaces them with the binders for the matcher and the builder.

We could just ignore them, but that wouldn't be a "motivated ignoring". 

Best would be to change GHC so that `XPatSynBind GhcTc` is isomorphic to `Void`. I can prepare a patch tomorrow, but knowing GHC, it could take a long time until it lands in `master`.

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
gergoerdi: Good question. The GHC desugarer says:

```
        ; (force_var,sel_binds) <- mkSelectorBinds var_ticks pat PatBindRhs body'
          -- We silently ignore inline pragmas; no makeCorePair
          -- Not so cool, but really doesn't matter
```

So it was already a case before that any inline pragmas on these were ignored.

gergoerdi:
> Thanks @gergoerdi! Quite an achievement.

Yeah I'm quite proud of getting this to work.

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
facundominguez: > why did you patch the module before typechecking?

It looked equally good to do it before or after typechecking, no other reason.

If there is no special motivation to skip ids in AbsBinds, then I'd say they are better marked as the rest.

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
facundominguez: That comment does not make explicit the purpose for which inline pragmas are ignored. And  the inline pragmas that `makeCorePair` reads are carried already in the Ids.

In addition, I have recollections of the NOINLINE pragmas making a difference to the simple optimizer.

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
gergoerdi: No no, I think you misunderstand.

I've put in a whole lot of work into making the desugarer output match what we had before. Everything in this patch is carefully orchestrated to . You HAVE to leave `AbsBind` `Id`s alone for that to work. If you make them non-inlined, what you get out of the desugarer is a garbled mess that LH can't make heads or tails of.

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
gergoerdi: Of course the noinline pragmas, IN GENERAL, make a difference. THis whole patch is about replicating that!

But in the desugarer, `PatBind`s don't seem to take them into account.

I can try changing the logic so that we make all `PatBind`-bound `Id`s noinline as well, but I wouldn't be surprised if that starts breaking things instead of fixing anything.

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
facundominguez:
```suggestion
        PatSynBind{} -> error "markNoInline: unexpected PatSynBind, should have been eliminated by the typechecker"
```

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
facundominguez: Oh, I pretty much can misunderstand everything in LH and the GHC API. When it comes to motivating this code, I think it is better to put it explicitly in comments, or other contributors and myself won't know why things are done the way they are.

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
facundominguez: Here's an example where `NOINLINE` pragmas make a difference in pattern bindings.
```Haskell
module Map0 where                                                                                                                                                                             
                                                                                                                                                                                              
data Map k a = Tip                                                                                                                                                                            
             | Bin Int k a (Map k a) (Map k a)                                                                                                                                                
                                                                                                                                                                                              
deleteFindMax t
  = case t of
      Bin _ k x l Tip -> (k, x, l)
      Bin _ k x l r ->
        let {-# NOINLINE km3 #-}
            {-# NOINLINE vm #-}
            {-# NOINLINE rm #-}
            (km3, vm, rm) = deleteFindMax r
         in (km3, vm, (balance k x l rm))
      Tip           -> (error ms, error ms, Tip)
  where ms = "Map.deleteFindMax : can not return the maximal element of an empty Map"

balance :: k -> a -> Map k a -> Map k a -> Map k a                                                                                                                                            
balance k x l r = undefined
```
This is taken from the discussion in #2257.

Using `ghc-9.10.1 -ddump-ds -fforce-recomp -dsuppress-all Map0.hs`, without `NOINLINE`
```Haskell
      letrec {
        deleteFindMax_aLW
          = \ t_awC ->
              let {
                ms_awD
                  = unpackCString#
                      "Map.deleteFindMax : can not return the maximal element of an empty Map"# } in
              case t_awC of {             
                Tip -> (error $dIP_aN4 ms_awD, error $dIP_aN9 ms_awD, Tip);
                Bin _ k_awE x_awF l_awG ds_dSG ->
                  case ds_dSG of {
                    __DEFAULT ->
                      let { ds_dSE = deleteFindMax_aLW ds_dSG } in
                      (case ds_dSE of { (km3_aMB, _, _) -> km3_aMB },
                       case ds_dSE of { (_, vm_aMC, _) -> vm_aMC },
                       balance
                         k_awE x_awF l_awG (case ds_dSE of { (_, _, rm_aMD) -> rm_aMD }));
                    Tip -> (k_awE, x_awF, l_awG)
                  }
              }; } in
      deleteFindMax_aLW
```

With `NOINLINE`
```Haskell
...
      letrec {
        deleteFindMax_aLX
          = \ t_awD ->
              let {
                ms_awE
                  = unpackCString#
                      "Map.deleteFindMax : can not return the maximal element of an empty Map"# } in
              case t_awD of {
                Tip -> (error $dIP_aN5 ms_awE, error $dIP_aNa ms_awE, Tip);
                Bin _ k_awF x_awG l_awH ds_dSH ->
                  case ds_dSH of {
                    __DEFAULT ->
                      let { ds_dSF = deleteFindMax_aLX ds_dSH } in
                      let { km3_aMC = case ds_dSF of { (km3_aMC, _, _) -> km3_aMC } } in
                      let { vm_aMD = case ds_dSF of { (_, vm_aMD, _) -> vm_aMD } } in
                      let { rm_aME = case ds_dSF of { (_, _, rm_aME) -> rm_aME } } in
                      (km3_aMC, vm_aMD, balance k_awF x_awG l_awH rm_aME);
                    Tip -> (k_awF, x_awG, l_awH)
                  }
              }; } in
      deleteFindMax_aLX
...
```


I think we should mark ids in lhs to preserve such behavior.

gergoerdi: Please add this as a test today so I'll have it on `develop` tomorrow when I'll hope to finalize this PR

facundominguez: Oh, my days are not as long as I want them, but I imagine such a test could go in
[GhcApiTests.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/liquidhaskell-boot/ghc-api-tests/GhcApiTests.hs).

The test could parse and typecheck the source code, then patch the bindings in the `TcgEnv`, then desugar, and then check that the bindings are indeed not inlined.

For the sake of this PR, I don't think the automation is essential. We could manually inspect with `-ddump-ds` that the desugaring doesn't inline the bindings in this example.

Update: I just ran the example with your changes, and the variables `km3`, `vm`, and `rm` appear marked as NOINLINE already. The transformation in `addNoInlinePragmasToBinds` is marking them when marking the ids in `abe_poly`. Perhaps if the ids are marked in `PatBind`, it is not necessary to mark ids in `abs_exports` at all.

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
gergoerdi: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/13441 FYI

gergoerdi:
Net -30 lines of code, that's how you know it's the good stuff!

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -213,18 +213,6 @@ lhDynFlags _ hscEnv =
            `gopt_set` Opt_KeepRawTokenStream
       }
 
-maybeInsertBreakPoints :: Config -> DynFlags -> DynFlags
```
facundominguez, line 205: > Breakpoints?! Where we're going, we don't need breakpoints.

Another question to consider is where we are [coming from](https://gitlab.haskell.org/ghc/ghc/-/issues/24386#note_544961).

The flag `--insert-core-break-points` is there to make it convenient to test if breakpoints have an effect in failing examples that used to work in earlier versions of LH.

If you want to remove the effect, you would also have to remove that flag. But perhaps there is no urgency to do this now.

gergoerdi: But with the new implementation of `addNoinlinePragmasToBinds`, we don't need to play games like htis anymore. So I think we should just remove the flag as well, then.

gergoerdi: But with the new implementation of `addNoinlinePragmasToBinds`, we don't need to play games like htis anymore. So I think we should just remove the flag as well, then.

facundominguez: Fine by me.

liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
```diff
@@ -482,11 +482,6 @@ config = cmdArgsMode $ Config {
     = def &= help "Dump pre-normalized core (before a-normalization)"
           &= name "dump-pre-normalized-core"
           &= explicit
-  , insertCoreBreakPoints
```
facundominguez, line 485: The flag is still mentioned in the changelog for the next release.

facundominguez:
I contributed a commit to remove the boolean from the recursion in `addNoInlinePragmasToBinds`. I hope it doesn't change the meaning of your solution, while it encapsulates the concern of skipping binds to the function that deals with `AbsBinds`.

gergoerdi:
Don't know how to add comment to specific line on mobile...

I think it's nicer if you use a record pattern match to take the `AbsBind` apart, similar to how we take e.g. `PatBind`s apart.

liquidhaskell-boot/src-ghc/Liquid/GHC/API/Extra.hs
```diff
@@ -170,45 +169,57 @@ apiCommentsParsedSource ps =
...
+    markId var = var `setInlinePragma` neverInlinePragma
+
+    -- The AbsBinds come from the GHC typechecker to handle polymorphism,
+    -- overloading, and recursion, so those don't correspond directly to
+    -- user-written `Id`s. In fact, marking them as `NOINLINE` results in Core
+    -- that LH can't process.
+    markAbsBinds :: AbsBinds -> AbsBinds
+    markAbsBinds absBinds0 =
+        absBinds0
+          { abs_binds = fmap (fmap skipFirstHsBind) $ abs_binds absBinds0
```
facundominguez: Actually, all tests continue to pass if I replace `skipFirstHsBind` with `go` here. So it is only the bit of `markABE` that makes a difference to LH.

facundominguez: I edited the motivation for `markAbsBinds` in 25c218a

facundominguez (approved):

Issue #1956, opened 2022-03-30
Plugin always colors the `**** LIQUID: ` header, regardless of compiler settings


Fizzixnerd:
This complicates tools which want to parse the output of the plugin and generate summary data (see #1955 for an example).

`src/Language/Haskell/Liquid/UX/CmdLine.hs` has a function `reportResult` that contains this enigmatic comment (see https://github.com/ucsd-progsys/liquidhaskell/blob/develop/src/Language/Haskell/Liquid/UX/CmdLine.hs#L744):

```haskell
         -- For now, always print the \"header\" with colours, irrespective to the logger
         -- passed as input.
```

ranjitjhala:
Hi Matt, we could certainly disable printing the header in “quiet” mode?

PR #2391, opened 2024-10-16
Don't print `LIQUID` banner in `--quiet` mode #1956

gergoerdi:


facundominguez (approved):
LGTM. Thanks!

I'll leave it open for another day in case someone else wants to comment.

PR #2394, opened 2024-10-17
Adapt to GHC 9.8

gergoerdi:
This makes `liquidhaskell-boot` work with GHC 9.8. I haven't bothered with the `base` annotations, because those would be a huge patch due to the `ghc-internals` stuff. I am hoping there is / will be a `ghc-internals` compatibility package at some point that just reexports the old `base` definitions.


liquidhaskell-boot/src/Language/Haskell/Liquid/UX/QuasiQuoter.hs
```diff
@@ -28,6 +28,7 @@ import qualified Language.Fixpoint.Types as F
 
 import Language.Haskell.Liquid.GHC.Misc (fSrcSpan)
 import Liquid.GHC.API  (SrcSpan)
+import Liquid.GHC.API.Compat
```
facundominguez, line 31: Compat could be arranged so this import is mandatory with ghc-9.10.1 as well.

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/RTypeOp.hs
```diff
@@ -51,6 +51,7 @@ module Language.Haskell.Liquid.Types.RTypeOp (
   where
 
 import qualified Liquid.GHC.API as Ghc
+import           Liquid.GHC.API.Compat
```
facundominguez, line 54: Compat could be arranged so this import is mandatory with ghc-9.10.1 as well.

liquidhaskell-boot/src/Language/Haskell/Liquid/Parse.hs
```diff
@@ -39,6 +39,7 @@ import qualified Text.PrettyPrint.HughesPJ              as PJ
 import           Text.PrettyPrint.HughesPJ.Compat       ((<+>))
 import           Language.Fixpoint.Types                hiding (panic, SVar, DDecl, DataDecl, DataCtor (..), Error, R, Predicate)
 import           Language.Haskell.Liquid.GHC.Misc       hiding (getSourcePos)
+import           Liquid.GHC.API.Compat
```
facundominguez, line 42: Compat could be arranged so this import is mandatory with ghc-9.10.1 as well.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Misc.hs
```diff
@@ -931,7 +931,7 @@ withWiredIn m = discardConstraints $ do
   --   ) wiredIns
 
   sigs wiredIns = concatMap (\w ->
-      let inf = maybeToList $ (\(fPrec, fDir) -> Ghc.L locSpanAnn $ Ghc.FixSig Ghc.noAnn $ Ghc.FixitySig Ghc.NoNamespaceSpecifier [Ghc.L locSpanAnn (tcWiredInName w)] $ Ghc.Fixity Ghc.NoSourceText fPrec fDir) <$> tcWiredInFixity w in
+      let inf = maybeToList $ (\(fPrec, fDir) -> Ghc.L locSpanAnn $ Ghc.FixSig Ghc.noAnn $ Ghc.FixitySig noExtField [Ghc.L locSpanAnn (tcWiredInName w)] $ Ghc.Fixity Ghc.NoSourceText fPrec fDir) <$> tcWiredInFixity w in
```
facundominguez, line 925: This difference would not be necessary if Compat offers a function to build FixSig.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Misc.hs
```diff
@@ -909,7 +909,7 @@ withWiredIn m = discardConstraints $ do
   -- undef <- lookupUndef
   wiredIns <- mkWiredIns
   -- snd <$> tcValBinds Ghc.NotTopLevel (binds undef wiredIns) (sigs wiredIns) m
-  (_, _, a) <- tcValBinds Ghc.NotTopLevel [] (sigs wiredIns) m
+  (_, a) <- tcValBinds Ghc.NotTopLevel [] (sigs wiredIns) m
```
facundominguez, line 903: This diff would not be needed if Compat offered a wrapper for tcValBinds that only exposed the last component.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Elaborate.hs
```diff
@@ -626,9 +626,9 @@ fixExprToHsExpr _ e =
 constantToHsExpr :: F.Constant -> LHsExpr GhcPs
 -- constantToHsExpr (F.I c) = noLoc (HsLit NoExt (HsInt NoExt (mkIntegralLit c)))
 constantToHsExpr (F.I i) =
-  noLocA (HsOverLit noExtField (mkHsIntegral (mkIntegralLit i)))
+  noLocA (HsOverLit noComments (mkHsIntegral (mkIntegralLit i)))
 constantToHsExpr (F.R d) =
-  noLocA (HsOverLit noExtField (mkHsFractional (mkTHFractionalLit (toRational d))))
+  noLocA (HsOverLit noComments (mkHsFractional (mkTHFractionalLit (toRational d))))
```
facundominguez: Compat could offer a function to construct `HsOverLit` to avoid this diff.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Elaborate.hs
```diff
@@ -566,7 +566,7 @@ renameBinderSort f = rename
 
 
 mkHsTyConApp ::  IdP GhcPs -> [LHsType GhcPs] -> LHsType GhcPs
-mkHsTyConApp tyconId tyargs = nlHsTyConApp NotPromoted Prefix tyconId (map (HsValArg noExtField) tyargs)
+mkHsTyConApp tyconId tyargs = nlHsTyConApp NotPromoted Prefix tyconId (map HsValArg tyargs)
```
facundominguez: This diff would not be necessary if this definition was moved to `Liquid.GHC.API.Extra`.

facundominguez (approved):
There are a some refactoring tasks that would help avoiding potential conflicts with `develop`. But I'm fine merging this as is, if you prefer to not address those now.

liquidhaskell-boot/src/Language/Haskell/Liquid/Parse.hs
gergoerdi: Yeah, but how? The problem is that if we import `Compat` (or `Compat (foldl')`) on `develop`, it just results in a "redundant import" warning that is turned into an error.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Misc.hs
gergoerdi: Yes, of course. The reason I didn't want to go through all these is because it's very ad-hoc, and on `develop` you wouldn't know when to add a new one anyway, so wouldn't this kind of stuff be better captured in the individual merge commits from `develop` into `ghc-9.8-branch`, for the merges that introduce new such uses of the GHC API?

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Elaborate.hs
gergoerdi: This one I agree with, since this makes sense to move to `API.Extra`.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Misc.hs
gergoerdi: TBH this function looks to me, as the kids these days would say, hella sus. There must be a better way to do this instead of pretend-typechecking a bunch of type signatures.

I'll make a PR to at least clear out some of the cruft...

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Misc.hs
facundominguez: I'm proposing the change only to reduce conflicts when rebasing or cherry picking commits. If anyone were to modify some other part of this line, it would cause a conflict between the branches.

> on develop you wouldn't know when to add a new one anyway

It is possible that when upgrading compilers and there is a definition change like this, the author could put it in Compat instead of doing the modification in place.

But I think it's fine to defer on these if the value is not evident yet.

liquidhaskell-boot/src/Language/Haskell/Liquid/Parse.hs
facundominguez: Good point. We would need to do something sneaky like:
https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/rebindable_syntax.html?highlight=implicitprelude#custom-prelude-modules-named-prelude

I don't have a strong opinion at the moment.

gergoerdi: I do have a strong opinion :), but not a solution, unfortunately.

I wish GHC had a proper way to mark (in the module definition, not at the import site) a module as "does not count towards redundant imports".

facundominguez:
Merging this. Follow up PRs welcome if any. Thanks's @gergoerdi!

PR #2398, opened 2024-10-18
Prove all the map fusion rewrite rules in test FuseMap.hs

AlecsFerra:
[https://hackage.haskell.org/package/ghc-internal-9.1001.0/docs/src/GHC.Internal.Base.html#map](https://hackage.haskell.org/package/ghc-internal-9.1001.0/docs/src/GHC.Internal.Base.html#map)


AlecsFerra:
The new proof it's also useful to to check that we don't break higher order polymorphism

facundominguez (approved):

PR #2407, opened 2024-10-24
Use GHC Name to represent resolved type constructors

facundominguez:
This PR creates a type `LHName` that is used to represent unresolved names, or resolved names that correspond to Haskell entities (`GHC Name`), or names for entities only known to LH (like type aliases), or local names (like parameters of type aliases). While `LHNames` can be used to name all things in LH, this PR only uses them for type constructors.

At first, I tried to parameterize every type in LH by the type of the name, but it was a lot of work to update all the functions, and multiple passes to update the types would have been necessary as the renaming is implemented piece wise. I opted instead to insert the LHNames in the AST without any attempt at changing the type parameters. Keeping track of the state of name resolution in the type of AST expressions can be done separately.

The commit 41fad95 introduces the type `LHName`.

The commit 3035033 implements the serialization and resolution of these names using generic traversals.

The commit 7c3c213 introduces the type `LHName` in the AST to represent names of type constructors. This commit doesn't change the behavior of the code, because whenever LH used Symbols for type constructors, I now insert an LHName, and the LHName contains the old Symbol for all purposes.

The next 9 commits modify bit by bit the usage sites of the old Symbols, and replace them with uses of the resolved names inside LHName. This has been valuable in practice to detect quickly when things broke.

Making use of GHC `Name` for type constructors required using `LHNames` for type constructors in embed directives, which I had skipped at first, and it is now implemented in 178aee8.

See how we need to qualify less in the LHAssumptions modules in the update in df2a21d. All type constructors are picked from the names in scope only, and most importantly, these choices are recorded when serializing the specs to interface files. Note how the test that we had originally identified for this flaw is now passing (commit 6857a42)!


facundominguez:
bot
top

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Resolve.hs
```diff
@@ -915,38 +921,37 @@ ofBRType env name f l = go []
...
-matchTyCon env name lc@(Loc _ _ c) arity
-  | isList c && arity == 1  = Right Ghc.listTyCon
-  | isTuple c               = Right tuplTc
-  | otherwise               = resolveLocSym env name msg lc
-  where
-    msg                     = "matchTyCon: " ++ F.showpp c
-    tuplTc                  = Ghc.tupleTyCon Ghc.Boxed arity
 
+matchTyCon :: Env -> Located LHName -> Lookup Ghc.TyCon
+matchTyCon env lc@(Loc _ _ c0) = unsafePerformIO $ do
```
facundominguez, line 934: I'm using `unsafePerformIO` here to lookup the `TyThing` of a `Name`. I think this is a benign use of `unsafePerformIO` as the result doesn't depend of when exactly this is called, as long ghc hasn't been somehow finalized.

Alternatively, it would be possible to look for the `TyThing` in advance, in a monadic context. The `TyThing` would need to be stored somewhere until it is needed and it interferes a bit with the generic traversals as it has no `Data` instance.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Resolve.hs
```diff
@@ -915,38 +921,37 @@ ofBRType env name f l = go []
...
-matchTyCon :: Env -> ModName -> LocSymbol -> Int -> Lookup Ghc.TyCon
-matchTyCon env name lc@(Loc _ _ c) arity
-  | isList c && arity == 1  = Right Ghc.listTyCon
-  | isTuple c               = Right tuplTc
-  | otherwise               = resolveLocSym env name msg lc
-  where
-    msg                     = "matchTyCon: " ++ F.showpp c
-    tuplTc                  = Ghc.tupleTyCon Ghc.Boxed arity
 
+matchTyCon :: Env -> Located LHName -> Lookup Ghc.TyCon
```
facundominguez, line 933:
```suggestion
-- | Get the TyCon from an LHName.
--
-- This function uses 'unsafePerformIO' to lookup the 'Ghc.TyThing' of a 'Ghc.Name'.
-- This should be benign because the result doesn't depend of when exactly this is
-- called. Since this code is intended to be used inside a GHC plugin, there is no
-- danger that GHC is finalized before the result is evaluated.
matchTyCon :: Env -> Located LHName -> Lookup Ghc.TyCon
```

ranjitjhala:
Thanks @facundominguez -- this is a really big deal!!!

PR #2413, opened 2024-10-28
Disable rewrite rules during desugaring

### LINKED ISSUE #2405: Broken reflection over lists
The following code throws an error on `develop` version of LH:

```haskell
{-@ LIQUID "--reflection" @-}

module Foo where

{-@ reflect singleton @-}
singleton :: a -> [a]
singleton x = [x]
```

The error is:

```
<no location info>: error:
    Uh oh.
    src/Foo.hs:7:13 Cannot lift Haskell function `singleton` to logic
                    Cannot transform lambda abstraction to Logic:	\ (c_dE4 [OS=OneShot] :: a -> a1 -> a1)
  (n_dE5 [OS=OneShot] :: a) ->
  c_dE4 x n_dE5

 Try using a helper function to remove the lambda.
```

A workaround is to replace `[x]` with `x : []` which suggests perhaps the issue is that the list term is being confused with a list type.

gergoerdi:
Fixes #2405

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
@@ -231,7 +231,9 @@ desugarerDynFlags df = (foldl gopt_unset df disabledOpts)
     }
   where
     disabledOpts =
-      [ Opt_EnableRewriteRules -- Disable stream fusion
+      [ Opt_EnableRewriteRules -- Prevent the simple optimizer from firing rewrite rules
+                               -- during desugaring. See tests/reflect/pos/T2405.hs for
+                               -- a discussion of an example where this is unwanted.
       ]
 
 
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
@@ -225,10 +225,15 @@ lhDynFlags _ hscEnv =
 -- TODO: We shouldn't rely on exports to find out what's in a ModGuts
 -- https://github.com/ucsd-progsys/liquidhaskell/pull/2388#issuecomment-2411418479
 desugarerDynFlags :: DynFlags -> DynFlags
-desugarerDynFlags df = df
+desugarerDynFlags df = (foldl gopt_unset df disabledOpts)
     { debugLevel   = 1                  -- To keep source note ticks
     , backend      = interpreterBackend -- To export everything
     }
+  where
+    disabledOpts =
+      [ Opt_EnableRewriteRules -- Disable stream fusion
+      ]
+
 
 --------------------------------------------------------------------------------
 -- | Parsing phase -------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


gergoerdi:
@facundominguez could you show me how to add #2405 as a test case?

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -225,10 +225,15 @@ lhDynFlags _ hscEnv =
...
 -- https://github.com/ucsd-progsys/liquidhaskell/pull/2388#issuecomment-2411418479
 desugarerDynFlags :: DynFlags -> DynFlags
-desugarerDynFlags df = df
+desugarerDynFlags df = (foldl gopt_unset df disabledOpts)
     { debugLevel   = 1                  -- To keep source note ticks
     , backend      = interpreterBackend -- To export everything
     }
+  where
+    disabledOpts =
+      [ Opt_EnableRewriteRules
```
facundominguez: Given that the failure is not too easy to reproduce, more of an explanation would be needed to communicate why this can't be removed.

facundominguez (commented):
Thanks @gergoerdi! Adding another test is demonstrated in 158cc43, for instance.

tests/reflect/pos/T2405.hs
```diff
@@ -0,0 +1,9 @@
+{-# OPTIONS_GHC -O #-}
+
+{-@ LIQUID "--reflection" @-}
+
+module T2405 where
```
facundominguez, line 16:
```suggestion
-- | Compiling with optimizations causes some rewrite rules to fire in the simple
-- optimizer during desugaring. In the case of list functions, stream fusion rewrite
-- rules are enabled, causing @singleton x = [x]@ to be desugared as
--
-- > singleton = \ @a x -> build (\ @a c_dE7 n_dE8 -> c_dE7 x n_dE8)
--
-- which then cannot be reflected because of the lambda expression.
--
-- LH disables rewrite rules when desugaring for verification, and this test is an
-- attempt to show the need for it.
--
module T2405 where
```

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -225,10 +225,15 @@ lhDynFlags _ hscEnv =
...
 -- https://github.com/ucsd-progsys/liquidhaskell/pull/2388#issuecomment-2411418479
 desugarerDynFlags :: DynFlags -> DynFlags
-desugarerDynFlags df = df
+desugarerDynFlags df = (foldl gopt_unset df disabledOpts)
     { debugLevel   = 1                  -- To keep source note ticks
     , backend      = interpreterBackend -- To export everything
     }
+  where
+    disabledOpts =
+      [ Opt_EnableRewriteRules -- Disable stream fusion
```
facundominguez:
```suggestion
      [ Opt_EnableRewriteRules -- Prevent the simple optimizer from firing rewrite rules
                               -- during desugaring. See tests/reflect/pos/T2405.hs for
                               -- a discussion of an example where this is unwanted.
```

facundominguez (approved):

PR #2416, opened 2024-10-30
Simplify TyThing lookups

facundominguez:


facundominguez:
cc @gergoerdi who did [recently](https://github.com/ucsd-progsys/liquidhaskell/commit/e063938) some changes to the lookup code that this PR is removing.

All the tests pass.

PR #2433, opened 2024-11-11
Handle type alias + refinements

### LINKED ISSUE #2431: Props ignore refinements
```haskell
{-# Language GADTs #-}

module False where

data Id where
{-@ MkId :: Prop (Id 12) @-}
    MkId :: Id
data ID = Id Int

{-@ bad :: { v:Prop (Id 12) | False } @-}
bad = MkId
```

The following passes verification, but not because there is some bug that is actually leading to a proof of `False` but because Liquid Haskell is ignoring the added constraint, this can also be seen by looking at the generated `fq` file that is totally missing the constraint.

AlecsFerra:
Closes #2431

### DIFF
--- a/tests/neg/RefinedProp.hs
+++ b/tests/neg/RefinedProp.hs
@@ -1,0 +1,16 @@
+{-# Language GADTs #-}
+
+{-@ LIQUID "--expect-any-error" @-}
+
+module RefinedProp where
+
+import Language.Haskell.Liquid.ProofCombinators
+
+data Id where
+{-@ MkId :: Prop (Id 12) @-}
+    MkId :: Id
+data ID = Id Int
+
+-- Should error as False is supposed to not be satisfied
+{-@ bad :: { v:Prop (Id 12) | False } @-}
+bad = MkId
--- a/tests/neg/RefinedProp.hs
+++ b/tests/neg/RefinedProp.hs
@@ -1,7 +1,5 @@
 {-# Language GADTs #-}
 
-{-@ LIQUID "--expect-any-error" @-}
-
 module RefinedProp where
 
 import Language.Haskell.Liquid.ProofCombinators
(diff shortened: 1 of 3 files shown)


AlecsFerra:
I don't know if we want to perform the meet in all the other possible instances of `strengthen`

facundominguez:
> I don't know if we want to perform the meet in all the other possible instances of `strengthen`

Ah, looks indeed like it could make sense. Maybe @nikivazou could comment on that.

AlecsFerra:
Ok if that is the case I think it should be a different PR, as from my understanding that code is non triggerable by refinement type aliases as you always have a Hole there.

tests/neg/RefinedProp.hs
```diff
@@ -0,0 +1,16 @@
...
+import Language.Haskell.Liquid.ProofCombinators
+
+data Id where
+{-@ MkId :: Prop (Id 12) @-}
+    MkId :: Id
+data ID = Id Int
+
+-- Should error as False is supposed to not be satisfied
+{-@ bad :: { v:Prop (Id 12) | False } @-}
+bad = MkId
```
facundominguez: How about using `{-@ fail bad @-}` instead of allowing any error?
```suggestion
module RefinedProp where

import Language.Haskell.Liquid.ProofCombinators

data Id where
{-@ MkId :: Prop (Id 12) @-}
    MkId :: Id
data ID = Id Int

-- Should error as False is supposed to not be satisfied
{-@ fail bad @-}
{-@ bad :: { v:Prop (Id 12) | False } @-}
bad = MkId
```

AlecsFerra: Did not know you could do that

facundominguez (approved):

nikivazou:
Yes, I think we should do the strengthen in all types. 
(Actually if I remember correctly, the strengthen function should have been written before we added the RHole type, thus this "bug")

PR #2434, opened 2024-11-11
Remove untested typeclass elaboration

facundominguez:


facundominguez:
@ranjitjhala, @nikivazou, there is no test depending on the code in [Bare.Elaborate](https://github.com/ucsd-progsys/liquidhaskell/blob/17bc8d8a41cd0c68df02255c4c502116f9adfdd1/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Elaborate.hs) which this PR is removing.

Section 3.2 of [Verifying Replicated Data Types with Typeclass Refinements
in Liquid Haskell](https://www.cs.umd.edu/~mwh/papers/lh-typeclasses.pdf) seems to be explaining that such code is needed to introduce dictionaries in the predicates of refinement types.

Do you have any tests that rehearse this code? My first attempt to produce a test faced a crash.

facundominguez:
hm, section 3.2 of [Verifying Replicated Data Types with Typeclass Refinements
in Liquid Haskell](https://www.cs.umd.edu/~mwh/papers/lh-typeclasses.pdf) seems to be explaining that code like this is needed to introduce dictionaries in the predicates of refinement types.

ranjitjhala:
@nikivazou -- can you help with this? I am unfamiliar with the code? Is it "dead" code or no?

ranjitjhala:
I agree with @facundominguez btw that we should either add tests that exercise this functionality, or, if there is no such code because the type-class stuff has not been kept compatible, then we should just remove or comment it out...

nikivazou:
Correct! There are the test of the artifacts of the paper, I guess, not all of them are part of the test suite. Will try to find them and let you know.

jprider63:
Here are [all the benchmarks](https://github.com/yiyunliu/liquid-benchmark/blob/4af5b347aa243fe687d8ebd049bbf3128ec8a96d/src/Main.hs#L103-L182) from the paper. Probably the ones from `liquid-base` make the most sense to add to the test suite. It might even make sense to export the verified versions of typeclasses (like `VSemigroup`) in a library for others to use.

nikivazou:
Thanks!!!! _I think_ it does not make sense to move all these in the test repo of Liquid Haskell..

facundominguez:
I just noticed some related tests that used to run on GHCs earlier than 9.
https://github.com/ucsd-progsys/liquidhaskell/blob/92c0a7eeb93045ace8a83df311c51287dee056fc/tests/tests.cabal#L2601

jprider63:
> I just noticed some related tests that used to run on GHCs earlier than 9.
> 
> https://github.com/ucsd-progsys/liquidhaskell/blob/92c0a7eeb93045ace8a83df311c51287dee056fc/tests/tests.cabal#L2601

Ah, maybe we should just update those tests to run with the newer GHC?

facundominguez:
Hello! A GSoC project is running to recover typeclass elaboration. One of the challenges is to figure out how to refresh the desugaring function that inserts dictionaries.
https://github.com/ucsd-progsys/liquidhaskell/blob/4749cced2bda64b5ce8d2e9b71d46cb75ff9de2c/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Misc.hs#L766

Let me know if you have any thoughts about how to look for this in the GHC API, or how to recompose it for ghc-9.14.1. :wave: 

cc @yiyunliu

facundominguez:
https://github.com/ucsd-progsys/liquidhaskell/pull/2716 reenabled the tests that show Elaborate.hs is not dead :)

PR #2435, opened 2024-11-12
Get default methods from the GHC API instead of using maybeResolveSym

facundominguez:
... also updates LF after removing monoid instances from there.


facundominguez:
Another step for #2169

PR #2441, opened 2024-11-15
Various small refactorings

facundominguez:
92ea9147 involves a change in behavior. Various annotations that are supposed to refer to definitions in the current module, could sometimes produce errors about ambiguity if the same names were imported. This has been changed so LH favors the local definition.

I think this is how LH used to work, but the behavior might have been affected by the latest work on name resolution.


facundominguez:
Another step for #2169

PR #2445, opened 2024-11-20
Expand type aliases in dependent pairs

facundominguez:
Before this patch, the following test would fail
```Haskell
{-@ type Id = {v:Int | v >= 0} @-}
{-@ lemma :: l:Id -> (id::Id, { true }) @-}
lemma :: Int -> (Int, ())
lemma x = (x, ())
```
with
```
Test.hs:6:27: error:
    Uh oh.
  panic, called at src/Language/Haskell/Liquid/Bare/Resolve.hs:972:37 in liquidhaskell-boot-0.9.10.1-inplace:Language.Haskell.Liquid.Bare.Resolve
    lookupTyThing: cannot resolve a LHRLogic name "Id"
  |
6 | {-@ lemma :: l:Id -> (id::Id, { true }) @-}
  |                           ^^
```

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
@@ -327,6 +327,16 @@ instance Expand BareType where
     = expandReft     rtEnv l -- apply expression aliases
     . expandBareType rtEnv l -- apply type       aliases
 
+instance Expand () where
+  expand _ _ = id
+
+instance Expand (BRType ()) where
+  expand rtEnv l
+    = expandReft     rtEnv l -- apply expression aliases
+    . void
+    . expandBareType rtEnv l -- apply type       aliases
+    . fmap (const mempty)
+
 instance Expand (RTAlias F.Symbol Expr) where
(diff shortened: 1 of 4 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
```diff
@@ -410,7 +420,7 @@ expandBareType rtEnv _ = go
     go t@RHole{}         = t
     go t@RVar{}          = t
     go t@RExprArg{}      = t
-    goRef (RProp ss t)   = RProp ss (go t)
+    goRef (RProp ss t)   = RProp (map (expand rtEnv l <$>) ss) (go t)
```
facundominguez, line 423: Summoning @nikivazou, in case there is a good reason why types in `ss` were not expanded.

facundominguez:
cc @clayrat

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
nikivazou: No reason that I can remember!

PR #2455, opened 2024-12-02
Fix detection of tyvars when generalizing sorts of gsImpAxioms

facundominguez:
Follow up to 9dc1230 in #2451.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Init.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Init.hs
@@ -222,14 +222,16 @@ measEnv sp xts cbs _tcb lt1s lt2s asms itys hs info = CGE
       lts         = lt1s ++ lt2s
       tcb'        = []
 
--- | Given an equation whose sorts are @FObj "a"@, @FObj "b"@, etc,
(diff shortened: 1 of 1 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Init.hs
```diff
@@ -222,14 +222,16 @@ measEnv sp xts cbs _tcb lt1s lt2s asms itys hs info = CGE
...
       tcb'        = []
 
--- | Given an equation whose sorts are @FObj "a"@, @FObj "b"@, etc,
+-- | Given an equation whose sorts are @FObj "a##..."@, @FObj "b##..."@, etc,
 -- replaces those with @FVar@s.
 eqToPolySort :: F.EquationV v -> F.Sort
 eqToPolySort e  =
     let s = foldr (F.FFunc . snd) (F.eqSort e) (F.eqArgs e)
-        ss = filter (all isLower . take 1 . F.symbolString) $ S.toList $ F.sortSymbols s
+        ss = filter (isTyVarName . F.symbolString) $ S.toList $ F.sortSymbols s
```
facundominguez, line 230: The test as it was would consider `int` and `bool` type variables as well, which is not what we need here.

PR #2458, opened 2024-12-04
Give precedence to export local measures over imported measures

facundominguez:
Follow up to #2456. This makes predictable which measures will be exported when there are multiple measures with the same name.

Ideally, Liquid Haskell would produce an error instead when there are conflicting names to export, and offer a mechanism to resolve them, like an export list for logic names. Until then, the provisions in this PR might do.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
@@ -114,7 +114,13 @@ makeTargetSpec cfg localVars lnameEnv lmap targetSrc bareSpec dependencies = do
     toLegacyDep (sm, ls) = (ModName SrcImport (Ghc.moduleName . Ghc.unStableModule $ sm), fromBareSpecLHName $ unsafeFromLiftedSpec ls)
 
     legacyDependencies :: [(ModName, BareSpec)]
-    legacyDependencies = map toLegacyDep . M.toList . getDependencies $ dependencies
+    legacyDependencies =
+      -- Dependencies are sorted lexicographically to make predictable which
+      -- logic names will have preference when exporting conflicting measures.
+      --
+      -- At the moment it is the measure from the last module after sorting.
+      -- But if there is a local conflicting measure, that one is used.
+      L.sortOn fst $ map toLegacyDep $ M.toList $ getDependencies dependencies
 
     -- Assumptions about local functions that are not exported aren't useful for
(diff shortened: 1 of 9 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -254,7 +260,7 @@ makeGhcSpec0 cfg ghcTyLookupEnv tcg instEnvs lenv localVars src lmap targetSpec
                     | v <- gsReflects refl
                     ]
                 , dataDecls = Bare.dataDeclSize mySpec $ dataDecls mySpec
-                , measures  = mconcat $ map Ms.measures $ mySpec : map snd dependencySpecs
+                , measures  = mconcat $ map Ms.measures $ map snd dependencySpecs ++ [mySpec]
```
sjoerdvisscher, line 265: Maybe this also requires a small comment, lest someone changes this back (for performance).

sjoerdvisscher (approved):
This all makes sense!

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
facundominguez, line 265:
```suggestion
                  -- Placing mySpec at the end causes local measures to take precedence over
                  -- imported measures when their names clash.
                , measures  = mconcat $ map Ms.measures $ map snd dependencySpecs ++ [mySpec]
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Specs.hs
```diff
@@ -700,8 +700,15 @@ instance Monoid (Spec lname ty) where
 -- Apart from less fields, a 'LiftedSpec' /replaces all instances of lists with sets/, to enforce
 -- duplicate detection and removal on what we serialise on disk.
 data LiftedSpec = LiftedSpec
-  { liftedMeasures   :: HashMap LHName (MeasureV LHName LocBareTypeLHName (F.Located LHName))
-    -- ^ User-defined properties for ADTs
+  { -- | Measures (a.k.a.  user-defined properties for ADTs
```
facundominguez:
```suggestion
  { -- | Measures (a.k.a.  user-defined properties for ADTs)
```

facundominguez:
Thanks for the review, @sjoerdvisscher!

PR #2461, opened 2024-12-09
Persist name resolution for class measures

facundominguez:
Another step for #2169

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
@@ -43,6 +43,7 @@
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE NamedFieldPuns             #-}
 {-# LANGUAGE TupleSections              #-}
 {-# LANGUAGE TypeApplications           #-}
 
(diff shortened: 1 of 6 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -682,6 +685,16 @@ resolveLogicNames cfg env globalRdrEnv unhandledNames lmap0 localVars lnameEnv p
     findReflection :: GHC.Name -> Maybe LHName
     findReflection n = GHC.lookupNameEnv (lneReflected lnameEnv) n
 
+mapMeasureNamesM :: Monad m => (LocSymbol -> m LocSymbol) -> MeasureV v ty ctor -> m (MeasureV v ty ctor)
```
sjoerdvisscher, line 688: This is smelling a lot like lenses!

tests/errors/MultiInstMeasures.hs
```diff
@@ -1,4 +1,4 @@
-{-@ LIQUID "--expect-error-containing=Multiple instance measures `sizeOf` for type `GHC.Internal.Ptr.Ptr`" @-}
+{-@ LIQUID "--expect-error-containing=Multiple instance measures `MultiInstMeasures.sizeOf` for type `GHC.Internal.Ptr.Ptr`" @-}
```
sjoerdvisscher, line 1: I would not expect a qualified name for an identifier from the current module.

sjoerdvisscher (approved):
Less comments to check the code against, but I guess that makes sense since nothing out of the ordinary is happening.

tests/errors/MultiInstMeasures.hs
facundominguez: It is an artifact of renaming. Names are qualified now, and I'm guessing the tidy passes for errors would need to be updated to strip the qualifiers or replace them with import aliases.

I'm leaving that for another time, but feel free to object.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
facundominguez: Yeah. I'm not eager to introduce the lense machinery for a few queries. But I'm open to be convinced that there are sufficient use opportunities to be worth it.

facundominguez:
Many thanks for the review @sjoerdvisscher!

PR #2462, opened 2024-12-10
Bless wired-in names as handled

facundominguez:
This wired in names are provided by the SMT solver, or belong to builtin Haskell types like lists and tuples. They can be used in every module and do not need imports.

A second additional commit documents one of my findings when analyzing the creation of selector measures.


sjoerdvisscher (approved):
💯

PR #2464, opened 2024-12-12
Simplify representation of measure names

facundominguez:
This PR changes the name field of measures to use LHName instead of Symbol. This makes possible to store more information in the name field, instead of paring the `Measure` records with LHName in specs.

I considered generalizing the type of the field instead of hard-coding it to LHName, but the new type parameter would have propagated to specs and it was a lot of changes and an extra parameter for no clear benefit.

Another commit provides a show function for LHNames that prints the structure of the name. And yet another commit fixes a problem where name resolution would panic if early stages produced an error and we allowed the resolution to proceed anyway, now we interrupt name resolution early if there are errors.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
@@ -144,7 +144,6 @@ resolveLHNames cfg thisModule localVars impMods globalRdrEnv lmap bareSpec0 depe
                        localVars
                        logicNameEnv0
                        privateReflectNames
-                       thisModule
                        sp1
             return (sp2, logicNameEnv0)
         logicNameEnv' = extendLogicNameEnv logicNameEnv ns
@@ -172,8 +171,12 @@ resolveLHNames cfg thisModule localVars impMods globalRdrEnv lmap bareSpec0 depe
         | otherwise ->
             lookupGRELHName ns lname s
               (fmap (either id GHC.getName) . Resolve.lookupLocalVar localVars (atLoc lname s))
+      LHNUnresolved LHLogicNameBinder s ->
+        pure $ makeLogicLHName s thisModule Nothing
+      n@(LHNUnresolved LHLogicName _) ->
+        -- This one will be resolved by resolveLogicNames
+        pure n
       LHNUnresolved ns s -> lookupGRELHName ns lname s listToMaybe
-      n@(LHNResolved (LHRLocal _) _) -> pure n
       n -> pure n
 
     lookupGRELHName ns lname s localNameLookup =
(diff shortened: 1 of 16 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Names.hs
```diff
@@ -231,7 +254,19 @@ getLHNameSymbol (LHNUnresolved _ s) = s
 -- | Get the unresolved Symbol from an LHName.
 getLHNameResolved :: HasCallStack => LHName -> LHResolvedName
 getLHNameResolved (LHNResolved n _) = n
-getLHNameResolved n@LHNUnresolved{} = error $ "getLHNameResolved: unresolved name: " ++ show n
+getLHNameResolved n@LHNUnresolved{} = error $ "getLHNameResolved: unresolved name: " ++ showLHNameDebug n
+
+showLHNameDebug :: LHName -> String
```
sjoerdvisscher, line 258: This should be the `Show` instance really, and what is now the instance should be a regular function that explains what it is for.

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Names.hs
```diff
@@ -59,14 +61,16 @@ selfSymbol = symbol ("liquid_internal_this" :: String)
...
-      lnSymbol :: !Symbol
-      -- | Module where the entity was defined
-    , lnModule :: !GHC.Module
-      -- | If the named entity is the reflection of some Haskell name
-    , lnReflected :: !(Maybe GHC.Name)
-    }
+data LogicName =
+    LogicName
+      { -- | Unqualified symbol
+        lnSymbol :: !Symbol
```
sjoerdvisscher: This change makes the field selector functions partial. Maybe just remove them?

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -131,22 +131,25 @@ resolveLHNames cfg thisModule localVars impMods globalRdrEnv lmap bareSpec0 depe
...
+                         globalRdrEnv
+                         unhandledNames
+                         lmap
+                         localVars
+                         logicNameEnv0
+                         privateReflectNames
+                         sp1
+              return (sp2, logicNameEnv0)
+            else
+              return (error "invalid spec", error "invalid logic environment")
```
sjoerdvisscher: Add a module name or function name to the error?

sjoerdvisscher (approved):
That cleans up nicely!

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
facundominguez: I'm counting on the error trace to show me the location. But I added the function name anyway in 919776d.

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Names.hs
facundominguez: I removed them in bd7dcd8.

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Names.hs
facundominguez: I tried that, and found that it influenced error messages intended for users. I'll address this in a follow up PR.

facundominguez:
Thanks for the review @sjoerdvisscher!

PR #2465, opened 2024-12-12
Support for lambdas :)

AlecsFerra:
Its breaking two tests related to typeclasses, both caused by the same file that breaks elaboration.

Companion PR in LF: https://github.com/ucsd-progsys/liquid-fixpoint/pull/725

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "liquid-fixpoint"]
 	path = liquid-fixpoint
-	url = https://github.com/alecsferra/liquid-fixpoint.git
+	url = https://github.com/ucsd-progsys/liquid-fixpoint.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "liquid-fixpoint"]
 	path = liquid-fixpoint
-	url = https://github.com/ucsd-progsys/liquid-fixpoint.git
+	url = https://github.com/alecsferra/liquid-fixpoint.git
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
@@ -41,6 +41,7 @@ import           Language.Haskell.Liquid.Bare.Measure as Bare
 import           Language.Haskell.Liquid.UX.Config
 import qualified Data.List as L
 import Control.Applicative
+import Control.Arrow (second)
 import Data.Function (on)
 import qualified Data.Map as Map
 import qualified Data.HashMap.Strict as M
@@ -330,7 +331,13 @@ makeAssumeType
   -> Ghc.Var -> Ghc.CoreExpr
   -> (LocSpecType, F.Equation)
 makeAssumeType allowTC tce lmap dm sym mbT v def
-  = (sym {val = aty at `strengthenRes` F.subst su ref},  F.mkEquation symbolV xts (F.subst su le) out)
+  = ( sym {val = aty at `strengthenRes` F.subst su ref}
+    , F.mkEquation 
+        symbolV 
+        (fmap (second $ F.sortSubst sortSub) xts)
+        (F.sortSubstInExpr sortSub (F.subst su le))
+        (F.sortSubst sortSub out)
+    )
   where
     symbolV = F.symbol v
     rt    = fromRTypeRep .
@@ -348,6 +355,18 @@ makeAssumeType allowTC tce lmap dm sym mbT v def
     ref        = F.Reft (F.vv_, F.PAtom F.Eq (F.EVar F.vv_) le)
     mkErr s    = ErrHMeas (sourcePosSrcSpan $ loc sym) (pprint $ val sym) (PJ.text s)
     bbs        = filter isBoolBind xs
+
+    -- KeepTyVars generalized
+    (tyVars, _) = Ghc.splitForAllTyCoVars τ
+    sortSub     = F.mkSortSubst $ zip (fmap F.symbol tyVars) (F.FVar <$> freeSort)
+    -- We need sorts that aren't polluted by rank-n types, we can't just look at
+    -- the term to determine statically what is the "maximum" sort bound ex
+    -- freeSort    = [1 + (maximum $ -1 : F.sortAbs out : fmap (F.sortAbs . snd)
+    -- xts) ..] as some variable may be bound to something of rank-n type.
+    -- In SortCheck.hs in fixpoint they just start at 42 for some reason.
+    -- I think Negative Debruijn indices (levels :^)) are safer
+    freeSort    = [-1, -2 ..]
+
     (xs, def') = GM.notracePpr "grabBody" $ grabBody allowTC (Ghc.expandTypeSynonyms τ) $ normalize allowTC def
     su         = F.mkSubst  $ zip (F.symbol     <$> xs) xArgs
                            ++ zip (simplesymbol <$> xs) xArgs
@@ -428,6 +447,7 @@ instance Subable Ghc.CoreAlt where
   subst su (Ghc.Alt c xs e) = Ghc.Alt c xs (subst su e)
 
 data AxiomType = AT { aty :: SpecType, aargs :: [(F.Symbol, SpecType)], ares :: SpecType }
+  deriving Show
 
 -- | Specification for Haskell function
 --
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
@@ -357,13 +357,13 @@ makeAssumeType allowTC tce lmap dm sym mbT v def
     mkErr s    = ErrHMeas (sourcePosSrcSpan $ loc sym) (pprint $ val sym) (PJ.text s)
     bbs        = filter isBoolBind xs
 
-    -- Before this modification reflected functions that were polymorphic were
-    -- getting compiled to monomorphic fixpint code.  As an example: 
+    -- rTypeSortExp produces monomorphic sorts from polymorphic types.
+    -- As an example, for 
     -- id :: a -> a ... id x = x 
-    -- Was generating: 
+    -- we got: 
     -- define id (x : a#foobar) : a#foobar = { (x : a#foobar) }
-    -- Using FObj instead of a real type variable (FVar) This code solves the
-    -- issue by creating a sobritution that replaces those "fake" type variables
+    -- Using FObj instead of a real type variable (FVar i) This code solves the
+    -- issue by creating a sort substitution that replaces those "fake" type variables
     -- with actual ones.
     -- define id (x : @-1) : a@-1 = { (x : a@-1) }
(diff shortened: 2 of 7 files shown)


AlecsFerra:
Some [examples](https://gist.github.com/AlecsFerra/c30ed548b3185b6a5d6e1256027b7384)

nikivazou:
That is amazing!!!! Though some tests seems to fail :'(

AlecsFerra:
> That is amazing!!!! Though some tests seems to fail :'(

Yeah is just the usual suspect `tests/classes/pos/STMonad.hs` everything else should work

AlecsFerra:
Added some tests and fixed the issue with `tests/classes/pos/STMonad.hs` thanks to @nikivazou

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "liquid-fixpoint"]
 	path = liquid-fixpoint
-	url = https://github.com/ucsd-progsys/liquid-fixpoint.git
\ No newline at end of file
+	url = https://github.com/alecsferra/liquid-fixpoint.git
```
facundominguez: Note: This needs reversal before merging, and after the LF changes are merged.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
```diff
@@ -328,7 +329,13 @@ makeAssumeType
...
   -> (LocSpecType, F.Equation)
 makeAssumeType allowTC tce lmap dm sym mbT v def
-  = (sym {val = aty at `strengthenRes` F.subst su ref},  F.mkEquation symbolV xts (F.subst su le) out)
+  = ( sym {val = aty at `strengthenRes` F.subst su ref}
+    , F.mkEquation 
+        symbolV 
+        (fmap (second $ F.sortSubst sortSub) xts)
+        (F.sortSubstInExpr sortSub (F.subst su le))
+        (F.sortSubst sortSub out)
+    )
```
facundominguez, line 341: Could we have some documentation for makeAssumeType that explains what it does? In particular for the new code. That could help review and maintenance.

facundominguez (commented):
Thanks @AlecsFerra! I make one request for documentation below.

If there are some limitations to keep in mind, it may be worth mentioning them in the user documentation.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
AlecsFerra: Let me know if the documentation makes sense

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
```diff
@@ -346,6 +354,26 @@ makeAssumeType allowTC tce lmap dm sym mbT v def
...
+
+    -- Before this modification reflected functions that were polymorphic were
+    -- getting compiled to monomorphic fixpint code.  As an example: 
+    -- id :: a -> a ... id x = x 
+    -- Was generating: 
+    -- define id (x : a#foobar) : a#foobar = { (x : a#foobar) }
+    -- Using FObj instead of a real type variable (FVar) This code solves the
+    -- issue by creating a sobritution that replaces those "fake" type variables
+    -- with actual ones.
+    -- define id (x : @-1) : a@-1 = { (x : a@-1) }
```
facundominguez:
```suggestion
    -- rTypeSortExp produces monomorphic sorts from polymorphic types.
    -- As an example, for 
    -- id :: a -> a ... id x = x 
    -- we got: 
    -- define id (x : a#foobar) : a#foobar = { (x : a#foobar) }
    -- Using FObj instead of a real type variable (FVar i) This code solves the
    -- issue by creating a sort substitution that replaces those "fake" type variables
    -- with actual ones.
    -- define id (x : @-1) : a@-1 = { (x : a@-1) }
```

AlecsFerra:
> If there are some limitations to keep in mind, it may be worth mentioning them in the user documentation.

AFAIK there are no limitations

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
facundominguez: Maybe your code renders the function [eqToPolySort](https://github.com/ucsd-progsys/liquidhaskell/blob/0c88b710a0ac7fc20363527ff88224b221ff923d/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Init.hs#L225-L234) unnecessary, which is only used to substitute sorts in equations as you are doing here.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
```diff
@@ -346,6 +354,26 @@ makeAssumeType allowTC tce lmap dm sym mbT v def
...
+    (tyVars, _) = Ghc.splitForAllTyCoVars τ
+    sortSub     = F.mkSortSubst $ zip (fmap F.symbol tyVars) (F.FVar <$> freeSort)
+    -- We need sorts that aren't polluted by rank-n types, we can't just look at
```
facundominguez, line 371: I'm missing what the concern is with rank-n types. Though it is probably relevant and I need to become aware.

facundominguez (commented):
Thanks for the comments!

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
AlecsFerra: No idea if somewhere we are still generating "fake" type variables

AlecsFerra: Still i think that function can lead to bugs as its using positive type variables indices

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
facundominguez: It helps, thanks!

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
AlecsFerra: suppose we have

```
def f (x : @0) : ... = { a x }
```

Where
```hs
a :: forall x. x -> (forall b . b -> ...)
```

During `a x` will be annotated with `a x : function(1, [@0; @0; ...]` Where the new forall (function) shadows the outsize type variable giving a wrong typing to the term that can lead to the best scenario crashes to infinite terms during unification (if was not fun at all to debug :)))

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
facundominguez: > No idea if somewhere we are still generating "fake" type variables

It would be a matter of removing the function and checking if the tests pass. I only added it because I had some failing test.

AlecsFerra: Yeah but I believe that on the end we should aim to get rid of that function anyways we should generate correct stuff in the first place

facundominguez: I contributed the removal of the substitutions in `eqToPolySort` as they were no longer needed :shrug: (293f739)

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
```diff
@@ -348,6 +356,26 @@ makeAssumeType allowTC tce lmap dm sym mbT v def
     ref        = F.Reft (F.vv_, F.PAtom F.Eq (F.EVar F.vv_) le)
     mkErr s    = ErrHMeas (sourcePosSrcSpan $ loc sym) (pprint $ val sym) (PJ.text s)
     bbs        = filter isBoolBind xs
+
+    -- rTypeSortExp produces monomorphic sorts from polymorphic types.
```
facundominguez, line 360: At this point, why not go a step further and fix `rTypeSortExp` so it doesn't produce `FObj` for type variables?

AlecsFerra: I don't know if it breaks stuff somewhere else, for example when I made that modification I broke coercions/GADTs as inside PLE the code that handled them was only taking care of the fake type variables and not of the real ones

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
```diff
@@ -348,6 +356,26 @@ makeAssumeType allowTC tce lmap dm sym mbT v def
...
+    -- define id (x : @-1) : a@-1 = { (x : a@-1) }
+    (tyVars, _) = Ghc.splitForAllTyCoVars τ
+    sortSub     = F.mkSortSubst $ zip (fmap F.symbol tyVars) (F.FVar <$> freeSort)
+    -- We need sorts that aren't polluted by rank-n types, we can't just look at
+    -- the term to determine statically what is the "maximum" sort bound ex:
+    -- freeSort = [1 + (maximum $ -1 : F.sortAbs out : fmap (F.sortAbs . snd) xts) ..] 
+    -- as some variable may be bound to something of rank-n type.  In
+    -- SortCheck.hs in fixpoint they just start at 42 for some reason.  I think
+    -- Negative Debruijn indices (levels :^)) are safer
+    freeSort    = [-1, -2 ..]
```
facundominguez, line 377: These are the sorts produced for composition, according to `--save`:
```
define GHC.Base_LHAssumptions.comp
         ( lq1 : func(0 , [@(-1); @(-2)])
         , lq2 : func(0 , [@(-3); @(-1)])
         , lq3 : @(-3)
         ) : @(-2) = {((lq1 (lq2 lq3)))}
```
Are these correct? The meaning of negative sort variables is unclear to me.

What should be the sort of `comp`? Maybe

```
constant GHC.Base_LHAssumptions.comp
         : func (3,
             [ func(0 , [@(-1); @(-2)])
             ; func(0 , [@(-3); @(-1)])
             ; @(-3)
             ; @(-2)
             ])
```

AlecsFerra: You need negative sort to prevent weird interactions with the rank2's, you need to renormalize them if you want to use func

AlecsFerra: The normalized version should be `func(3, [func(0, [@0, @1]), func(0, [@2, @0]), @2, @1])`

AlecsFerra: That's the same you should see for the constant in the fq file

facundominguez: Alright, that makes sense to me.

facundominguez (approved):
LGTM!

PR #2466, opened 2024-12-12
Persist name resolution for field names

### LINKED ISSUE #2412: Scope and meaning of fields in data specs
When trying to use GHC Names to resolve fields of data types, I found I cannot figure out what is their expected scope.

### Field names mismatch

Given a data declaration and its spec as follows:
```Haskell
{-@ data T = C { fs :: Int, gs :: Int } @-}
data T = C { f :: Int, g :: Int }
```
LH allows this, but then it will get confused if we add a lemma
```Haskell
{-@ lemma :: x:T -> { fs x == 0 } @-}
lemma :: T -> ()
lemma _ = ()
```
It will complain that `fs` is not in scope
```
    Unbound symbol fs --- perhaps you meant: VV, x, fst ?
    Just
   |
20 | {-@ lemma :: x:T -> { fs x == 0 } @-}
   |              ^^^^^^^^^^^^^^^^^^^^^
```
Changing the lemma to use `f` instead complains that `f` is not in scope either.
```Haskell
{-@ lemma :: x:T -> { f x == 0 } @-}
```
```
    Unbound symbol f --- perhaps you meant: VV, x, fst ?
```

The only way to get past this error is to name the field the same in both the data declaration and the spec
```Haskell
{-@ data T = C { f :: Int, g :: Int } @-}
data T = C { f :: Int, g :: Int }
```

### If field names are missing

What do we do if field names are missing in the data declaration? Can we just use `f` and `g` in specifications?
```Haskell
{-@ data T = C { f :: Int, g :: Int } @-}
data T = C Int Int
{-@ lemma :: x:T -> { f x == 0 } @-}
...
```
Conversely, if field names are missing in the specification instead, shall we  still use `f` and `g` in specifications?
```Haskell
{-@ data T = C Int Int @-}
data T = C { f :: Int, g :: Int }
{-@ lemma :: x:T -> { f x == 0 } @-}
...
```
At the moment LH would just complain that the selectors are unbound.

### Field names in GADTs

Should the fields in GADT syntax be available to use in specifications?
```Haskell
{-@ data T where C :: f:Int -> g:Int ->  T @-}
data T = C Int Int
{-@ lemma :: x:T -> { f x == 0 } @-}
...
```
At the moment LH also complains that the field names are unbound.

------------------
In all cases, reflected functions always seem to use the fields in specifications: e.g
```Haskell
{-@ data T ... f field is declared in any style ... @-}
data T = C Int Int

{-@ reflect g1 @-}
g1 :: T -> Int
g1 (C x _) = x
```
produces
```
define Test.g (lq1 : Test.T) : int = {((Test.C##f lq1))}
```

-------------------

Maybe the names of fields in data specifications should be considered synonyms with the names in data declarations. 

Any preferences?

Update: The documentation in the website says that measures are created for every field name in the spec. I'll try to work with that ...

facundominguez:
Another step for #2169.

Also fixes #2412 by making the field names available as measures as the documentation says it should.

The main contribution of this PR is persisting name resolution of field names. They are treated  much the same as the names of measures in #2464.

To get tests passing I had to stop LH from trying to qualify names at some places, which would interfere with name resolution of field names. This made me aware that much of the passes to qualify identifiers are now redundant, so I removed most of them.

An additional commit fixes name resolution for built-in data constructors and names in the logic map, which previously produced names with `makeLocalLHName`, and now they are constructed with `makeLogicLHName`. The big difference is that the later causes names to be qualified when converted to symbols. These omissions were identified when removing calls to qualifying passes.

The commit that makes `showLHNameDebug` the show instance for `LHName` is a follow up from #2456. A few error messages were impacted and there might be still a few places that aren't caught by tests.

Given that I had to remove some qualifying passes, I updated the mechanism to resolve expression aliases, which does require occurrences of expression aliases to be qualified. The name resolution pass now makes sure they are qualified.

Lastly I removed some calls to `getLHNameSymbol` from Bare.hs, replacing them with either `lhNameToUnqualifiedSymbol` or `lhNameToResolvedSymbol`, which defines what is the intended behavior.


facundominguez:
I managed to remove most of the qualify passes after fixing a mistake in `lhNameToResolvedSymbol`.

The only pass that remains is `Bare.qualifyFTycon`, that first requires persisting name resolution for qualifiers.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -317,14 +317,8 @@ makeGhcSpec0 cfg ghcTyLookupEnv tcg instEnvs lenv localVars src lmap targetSpec
...
     allowTC  = typeclass cfg
-    mySpec2  = Bare.qualifyExpand env name rtEnv l [] mySpec1    where l = F.dummyPos "expand-mySpec2"
-    iSpecs2  = Bare.qualifyExpand
-                 env
-                 name
-                 rtEnv
-                 (F.dummyPos "expand-iSpecs2")
-                 []
-                 (M.fromList dependencySpecs)
+    mySpec2  = Bare.expand rtEnv l mySpec1    where l = F.dummyPos "expand-mySpec2"
```
sjoerdvisscher: Clean-up: I think you should inline the `l` here.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -497,8 +491,7 @@ makeSpecQual _cfg env tycEnv measEnv _rtEnv specs = SpQual
 
 makeQualifiers :: Bare.Env -> Bare.TycEnv -> (ModName, Ms.Spec F.Symbol ty) -> [F.Qualifier]
 makeQualifiers env tycEnv (modn, spec)
-  = fmap        (Bare.qualifyTopDummy env        modn)
-  . Mb.mapMaybe (resolveQParams       env tycEnv modn)
+  = Mb.mapMaybe (resolveQParams       env tycEnv modn)
```
sjoerdvisscher: Clean-up: remove whitespace

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -858,24 +847,22 @@ myAsmSig v sigs = Mb.fromMaybe errImp (mbHome `mplus` mbImp)
...
           -> Bare.Lookup [(Ghc.Var, LocBareType, Maybe [Located F.Expr])]
-makeTExpr env name tySigs rtEnv spec = do
+makeTExpr env tySigs rtEnv spec = do
   vExprs       <- M.fromList <$> makeVarTExprs env spec
   let vSigExprs = Misc.hashMapMapWithKey (\v t -> (t, M.lookup v vExprs)) vSigs
-  return [ (v, t, qual t <$> es) | (v, (t, es)) <- M.toList vSigExprs ]
+  return [ (v, t, qual <$> es) | (v, (t, es)) <- M.toList vSigExprs ]
   where
-    qual t es   = qualifyTermExpr env name rtEnv t <$> es
+    qual es   = expandTermExpr rtEnv <$> es
```
sjoerdvisscher: Clean-up: line-up equal signs or remove whitespace

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
```diff
@@ -306,15 +309,14 @@ getExprFromUnfolding (Ghc.CoreUnfolding expr _ _ _ _) = Just expr
...
           -> (LocSymbol, Maybe SpecType, Ghc.Var, Ghc.CoreExpr)
           -> (Ghc.Var, LocSpecType, F.Equation)
 --------------------------------------------------------------------------------
-makeAxiom env tycEnv name lmap (x, mbT, v, def)
+makeAxiom env tycEnv lmap (x, mbT, v, def)
             = (v, t, e)
   where
-    t       = Bare.qualifyTop env name (F.loc t0) t0
-    (t0, e) = makeAssumeType allowTC embs lmap dm x mbT v def
+    (t, e) = makeAssumeType allowTC embs lmap dm x mbT v def
```
sjoerdvisscher: Clean-up: line up equals sign

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/DataType.hs
```diff
@@ -656,7 +656,7 @@ ofBDataCtorTc env name l l' tc αs ps πs _ctor@(DataCtor _c as _ xts res) c' =
     res'          = Bare.ofBareType env name l (Just ps) <$> res
     t0'           = dataConResultTy c' αs t0 res'
     _cfg          = getConfig env
-    (yts, ot)     = qualifyDataCtor (not isGadt) name dLoc (zip xs ts', t0')
+    (yts, ot)     = (zip xs ts', t0')
```
sjoerdvisscher: Assigning a tuple to a tuple

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Plugged.hs
```diff
@@ -137,17 +139,19 @@ makePluggedDataCon allowTC embs tyi ldcp
...
          -> ([Ghc.Var], [Ghc.Type],             Ghc.Type)   -- ^ hs type
-         -> ([RTyVar] , [(F.Symbol, SpecType)], SpecType)   -- ^ lq type
-         -> ([(F.Symbol, SpecType)], SpecType)              -- ^ plugged lq type
+         -> ([RTyVar] , [(LHName, SpecType)], SpecType)   -- ^ lq type
+         -> ([(LHName, SpecType)], SpecType)              -- ^ plugged lq type
 plugMany allowTC embs tyi ldcp (hsAs, hsArgs, hsRes) (lqAs, lqArgs, lqRes)
-                     = F.notracepp msg (drop nTyVars (zip xs ts), t)
+                     = F.notracepp msg (drop nTyVars (zip (map lookupLHName xs) ts), t)
   where
+    lookupLHName s = Mb.fromMaybe (panic (Just (GM.fSrcSpan ldcp)) $ "unexpected symbol: " ++ show s) $ lookup s lhNameMap
```
sjoerdvisscher: Clean-up: line up equals signs

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -1113,10 +1099,10 @@ mkReft _ _ _ _
 -------------------------------------------------------------------------------------------
 makeSpecName :: Bare.Env -> Bare.TycEnv -> Bare.MeasEnv -> ModName -> GhcSpecNames
 -------------------------------------------------------------------------------------------
-makeSpecName env tycEnv measEnv name = SpNames
+makeSpecName env tycEnv measEnv _name = SpNames
```
sjoerdvisscher: Should the `_name` argument be removed?

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
```diff
@@ -75,13 +75,13 @@ makeRTEnv env modName mySpec dependencySpecs lmap
           = renameRTArgs $ makeRTAliases tAs $ makeREAliases eAs
   where
     tAs   = [ t                   | (_, s)  <- specs, t <- Ms.aliases  s ]
```
sjoerdvisscher: Clean-up: remove whitepsace here too.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Measure.hs
```diff
@@ -561,15 +558,15 @@ mkMeasureSort env name (Ms.MSpec c mm cm im) =
 --------------------------------------------------------------------------------
 -- type BareMeasure = Measure LocBareType LocSymbol
 
-expandMeasure :: Bare.Env -> ModName -> BareRTEnv -> BareMeasure -> BareMeasure
-expandMeasure env name rtEnv m = m
+expandMeasure :: BareRTEnv -> BareMeasure -> BareMeasure
+expandMeasure rtEnv m = m
   { msSort = RT.generalize                   <$> msSort m
```
sjoerdvisscher: Clean-up: remove whitespace

sjoerdvisscher (approved):
Looks good, mostly some whitespace to fix.

facundominguez:
Thanks @sjoerdvisscher!

Issue #2090, opened 2022-10-29
[GADTs] Small example of a forgotten refinement


plredmond:
This GADT has a phantom type that gets refined (GHC-sense) by its constructor(s). That phantom type has a refinement (LH-sense) in the GADT-signature of its constructor. When we pattern match on the GADT constructor, GHC learns its type, but LH doesn't learn the refinement-type.

```haskell
{-@
data Foo result where A :: Foo {i:Int | i == 0 } @-}
data Foo result where A :: Foo Int

foo :: Foo result -> result
foo A = 123
```
http://goto.ucsd.edu:8090/index.html#?demo=permalink%2F1667027035_14060.hs

* **Expected:** This example should fail because 123 /= 0
* **Actual:** This example passes because LH applies no refinement-constraints in the branch where `Foo result` is constructed by `A`.

facundominguez:
Another fix here could be to have LH reject the spec of the data constructor `A`.

plredmond:
Yeah, if refining phantom types on gadt constructors isn't supported, then LH reject it early in the pipeline.

PR #2470, opened 2025-01-10
Fix unsafe data constructor refinements

AlecsFerra:
By allowing arbitrary refinements of data constructor we can prove easily false:

```hs
data F where
{-@ MkF :: {_:F | false} @-}
    MkF :: F

{-@ falseProof :: { false } @-}
falseProof = () ? MkF
```

Since deciding if assuming some proposition leads to a contradiction is impossible
and in practice the only usage of refining a data constructor is with data proposition `Prop` from `ProofCombinators` we can just allow refinements of the shape `{ v:T | prop v = foo }`.

All the other occurrences (2) can be solved without resorting to refining the data constructor:
- `isAdmit` from `ProofCombinators` can actually just be defined as an Haskell function
- `{-@ Count :: x:a -> {v:Count a | v == Count x } @-}` from `CountMonad.hs` can pass verification without it

If the user for some reason still wants to assume stuff on some constructor can use `assume` explicitly.


tests/neg/RefCtor.hs
```diff
@@ -0,0 +1,14 @@
+{-# LANGUAGE GADTs #-}
+{-@ LIQUID "--expect-any-error" @-}
```
facundominguez: Better to use `"--expect-error-containing=the error we care about"`, otherwise the test will pass if a different error appears.

tests/neg/RefCtor.hs
```diff
@@ -0,0 +1,14 @@
...
+module RefCtor where
```
facundominguez, line 5:
```suggestion
-- | Tests that the returned refinement type of data constructors
-- is not allowed to be other than @True@.
module RefCtor where
```

tests/pos/CountMonad.hs
```diff
@@ -36,7 +35,6 @@ assertCount _ x = x
 getCount :: Count a -> Int
 getCount _ = 0
 
-{-@ makeCount :: x:a -> {v:Count a | v == Count x} @-}
```
facundominguez, line 39: Is this not true anymore? I would expect it to remain valid.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Check.hs
```diff
@@ -193,6 +194,32 @@ checkTargetSpec specs src env cbs tsp
 
 
 
+checkConstructorRefinement :: [(Var, LocSpecType)] -> Diagnostics
```
facundominguez, line 206:
```suggestion
-- | Tests that the returned refinement type of data constructors has predicate @True@ or @prop v == e@.
--
-- > data T = T Int
-- > {-@ T :: x:Int -> { v:T | v = T x } @-} -- Should be rejected
-- > {-@ T :: x:Int -> { v:T | True } @-} -- Should be fine
-- > {-@ T :: x:Int -> { v:T | prop v = True } @-} -- Should be fine
--
checkConstructorRefinement :: [(Var, LocSpecType)] -> Diagnostics
```

facundominguez (commented):
There is one issue concerned with specs of GADTs (#2090) which is fine to close, I think. But Niki made efforts to fix another ticket that also uses non-trivial refinement types (#2349).

I don't have use cases of my own that require specs for the returned type of data constructors. I guess that if @nikivazou doesn't object, it should be fine to merge.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Check.hs
```diff
@@ -193,6 +194,32 @@ checkTargetSpec specs src env cbs tsp
...
+    getRetTyRef t                = ur_reft $ rt_reft t
+
+    -- True refinement
+    validRef (F.Reft (_, F.PTrue))
+                      = True
+    -- Prop foo from ProofCombinators
+    validRef (F.Reft (v, F.PAtom F.Eq (F.EApp (F.EVar n) (F.EVar v')) _)) 
+      | n == "Language.Haskell.Liquid.ProofCombinators.prop"
+      , v == v' 
+      = True
```
facundominguez, line 225: Does this need to be so strict? Maybe it is fine to allow both `prop v = foo` and  `foo = prop v`.

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Errors.hs
```diff
@@ -483,6 +483,11 @@ data TError t =
                 , dc   :: !Doc
                 }
 
+  | ErrCtorRefinement { pos :: !SrcSpan
+                      , ctorName :: !Doc
+                      } -- ^ The refinement of a doesn't admit
```
facundominguez:
```suggestion
                      } -- ^ The refinement of a data constructor doesn't admit
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Errors.hs
```diff
@@ -1061,6 +1066,11 @@ ppError' _ dCtx (ErrPosTyCon _ tc dc)
              , nest 2 "https://ucsd-progsys.github.io/liquidhaskell/options/#positivity-check"
             ]
 
+ppError' _ dCtx (ErrCtorRefinement _ ctorName)
+  = text "Refinement of the constructor" <+> ctorName <+> "doesn't admit an arbitrary refinements on the return type"
```
facundominguez:
```suggestion
  = text "Refinement of the data constructor" <+> ctorName <+> "doesn't admit an arbitrary refinements on the return type"
```

facundominguez (commented):
I'm curious about whether it is so easy to introduce accidentally `False` to deserve constraining the refinements of data constructors.

tests/pos/CountMonad.hs
AlecsFerra: It is still true bout you need `--exact-data-cons` or some other flag that triggers it

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Check.hs
AlecsFerra: No is fine also in the other direction, I just copied the way that is done in `ProofCombinators`

AlecsFerra:
> There is one issue concerned with specs of GADTs (#2090) which is fine to close, I think. But Niki made efforts to fix another ticket that also uses non-trivial refinement types (#2349).

All these issue seem to mixup Haskells type indexes with liquid value, for what Haskell is concerned GADTs indexes are types so it makes sense that they interact weirdly with LH, if someone wants type indexes that interact well with LH they should use `Prop`.


> I'm curious if it is easy to introduce accidentally `False` to deserve constraining the refinements of data constructors.

You can introduce some proposition on that on it's own is not false but can be inconsistent with some other proposition that you assumed somewhere else, especially since the refinement of constructors feels like function refinement I think the user expects them to be a sound feature (at least I was super convinced they were safe to use), if they had assume before the constructor name then it would be a different story.

Also since I wasn't able to find any other reasonable (where reasonable I believe should also imply sound) usage I think is fine to disallow it. If users want do potentially unsound shenanigans they can always use the assume keyword.

Also F* totally diallows refining the return type of a constructor for example.

nikivazou:
I think this is a very restrictive solution, but is it sound (definitely more sound than before), so merging and we could relax it in the future if needed.

AlecsFerra:
Added a flag to allow the old behavior

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Errors.hs
```diff
@@ -1061,6 +1066,11 @@ ppError' _ dCtx (ErrPosTyCon _ tc dc)
              , nest 2 "https://ucsd-progsys.github.io/liquidhaskell/options/#positivity-check"
             ]
 
+ppError' _ dCtx (ErrCtorRefinement _ ctorName)
+  = text "Refinement of the data constructor" <+> ctorName <+> "doesn't admit an arbitrary refinements on the return type"
+        $+$ dCtx
+        $+$ nest 4 (text "Were you trying to use `Prop` from `Language.Haskell.Liquid.ProofCombinators`?")
```
facundominguez, line 1072: I'm realizing the error message can be more precise to tell users exactly which predicates are allowed: "Please use either of True or prop v = ..."

facundominguez: " ... or flag --allow-unsafe-constructors"

AlecsFerra: Agree on mentioning the flag but idk about exposing the user to `prop` when in practice you always want to use `Prop`

AlecsFerra:
Infrastructure fail?

facundominguez:
Passes after a restart :)

docs/mkDocs/docs/options.md
```diff
@@ -500,6 +500,23 @@ worse = bad (Very bad)
 Note that all positive occurrences are permited, unlike Coq that only allows the strictly positive ones
 (see: https://vilhelms.github.io/posts/why-must-inductive-types-be-strictly-positive/)
 
+## Refined construcotrs
```
facundominguez:
```suggestion
## Refined data constructors
```

docs/mkDocs/docs/options.md
```diff
@@ -500,6 +500,23 @@ worse = bad (Very bad)
...
+
+**Options:** `allow-unsafe-constructors`
+By **default** only safe constructor refinements are allowed.
```
facundominguez:
```suggestion
By **default** only safe data constructor refinements are allowed.
```

PR #2472, opened 2025-01-13
Resolve type constructors in qualifiers via LHNames

facundominguez:
Another step for #2169

This PR reports unresolved type constructors one error at a time for qualifiers, and in a separate phase than the rest of name resolution.

I explored integrating name resolution of qualifiers in a single phase with the rest, but this turned out to be not trivial because the qualifier representation is defined in `liquid-fixpoint`.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
@@ -194,7 +194,6 @@ resolveLHNames cfg thisModule localVars impMods globalRdrEnv bareSpec0 dependenc
     lmap =
         (LH.listLMap <>) $
         mconcat $
-        reverse $
         map (mkLogicMap . HM.map (fmap lhNameToResolvedSymbol) . liftedDefines) $
         HM.elems $
         getDependencies dependencies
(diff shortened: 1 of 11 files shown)


facundominguez:
@clayrat, maybe you would like to review this one.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -191,11 +192,12 @@ resolveLHNames cfg thisModule localVars impMods globalRdrEnv bareSpec0 dependenc
...
     -- add defines from dependencies to the logical map
     lmap =
-      foldr (\ls lmp ->
-                 lmp <> mkLogicMap (HM.map (fmap lhNameToResolvedSymbol) $ liftedDefines ls)
-            )
-            LH.listLMap $
-            (HM.elems . getDependencies) dependencies
+        (LH.listLMap <>) $
+        mconcat $
+        reverse $
```
clayrat: It doesn't look to me like this `reverse` is necessary? `LogicMap` is internally a hashmap, so the order of addition shouldn't really matter.

facundominguez: The preceding `foldr` was careful with the order in which the logic maps were combined. Was it trying to give precedence to some definitions over others perhaps?

facundominguez: Nah, the order of the input is non-deterministic anyway. I'll remove the `reverse`.

clayrat: > The preceding `foldr` was careful with the order in which the logic maps were combined. Was it trying to give precedence to some definitions over others perhaps?

Not really, I just put in `foldr` because it's a canonical recursor for lists :)

Issue #2457, opened 2024-12-04
Type level nat parameter causes LH to check less


kleinreact:
Lets say I wanna have a bounded number type with `Num` capabilities, then I can get that for some fixed lower bound 0 and upper bound 9 as follows:

```haskell
{-@ LIQUID "" @-}

{-# OPTIONS_GHC -fplugin=LiquidHaskell #-}
{-# OPTIONS_GHC -fno-warn-unused-imports #-}
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE StandaloneKindSignatures #-}

module T2457 where

import GHC.Base (Type)
import GHC.TypeNats (Nat, Natural)

{-@ type BI010 = { i : Integer | i >= 0 && i < 10 } @-}

{-@ data T = T BI010 @-}
type T :: Type
data T = T Integer

{-@ inline t2i @-}
{-@ t2i :: t : T -> { i : BI010 | t == T i } @-}
t2i :: T -> Integer
t2i (T i) = i
{-# INLINE t2i #-}

t0 :: T
t0 = T 3

-- does not LH type check, as expected
--t1 :: T
--t1 = T 10

{-@
instance Num T where
  (+) ::
    a : T ->
    { b : T | t2i a + t2i b < 10 } ->
    { t : T | t2i t == t2i a + t2i b }

  (-) ::
    a : T ->
    { b : T | t2i a >= t2i b } ->
    { t : T | t2i t == t2i a - t2i b }

  (*) ::
    a : T ->
    { b : T | t2i a * t2i b < 10 } ->
    { t : T | t2i t == t2i a * t2i b }

  negate :: T -> T
  abs    :: T -> T
  signum :: T -> T

  fromInteger ::
    i : BI010 ->
    { t : T | t2i t == i }
@-}
instance Num T where
  a + b        = T $ t2i a + t2i b
  a - b        = T $ t2i a - t2i b
  a * b        = T $ t2i a * t2i b
  negate       = undefined
  abs          = id
  signum       = undefined
  fromInteger  = T

t2 :: T
t2 = 3 + 3

-- does not LH type check, as expected
--t3 :: T
--t3 = 3 + 8
```

LH checks the bounds in this case as expected. 

However, if I parameterize the bound using a type level natural instead, e.g.

```haskell
{-@ LIQUID "" @-}

{-@ embed Natural as Int @-}
[...]

PR #2473, opened 2025-01-15
Better error message for reflection faliures due to missing reflection of constructors + refactor of core to logic

AlecsFerra:
The error message was very unhelpful before (See #2457)

From
```
   Illegal type specification for `Fin.unFin`
    Fin.unFin :: forall n .
                 lq_tmp$db##0:(Fin.Fin n) -> {VV : GHC.Types.Int | VV == lqdc##$select##Fin.MkFin##1 lq_tmp$db##0
                                                                   && VV >= 0
                                                                   && VV < n}
    Sort Error in Refinement: {VV : int | VV == lqdc##$select##Fin.MkFin##1 lq_tmp$db##0
                                          && VV >= 0
                                          && VV < n}
    Unbound symbol lqdc##$select##Fin.MkFin##1 --- perhaps you meant: Fin.MkFin ?
    Just
   |
26 | unFin (MkFin i) = i
   | ^^^
```
To:
```
test/Testone2.hs:19:13: error:
    Uh oh.
  panic, called at src/Language/Haskell/Liquid/Bare/Axiom.hs:355:26 in liquidhaskell-boot-0.9.10.1-inplace:Language.Haskell.Liquid.Bare.Axiom
    test/Testone2.hs:19:13 Cannot lift Haskell function `p2i` to logic
                           Cannot lift to logic the constructor `Fixme.P` consider enabling either --exactdc or --reflection
   |
19 | {-@ reflect p2i @-}
   |             ^
```

I think to make it even better we should either handle the panic gracefully (best solution) or at least remove all the call stack information since is rather unhelpful to the user

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
@@ -352,7 +352,7 @@ makeAssumeType cfg tce lmap dm sym mbT v def
     _msg  = unwords [showpp sym, showpp mbT]
     le    = case runToLogicWithBoolBinds bbs tce lmap dm cfg mkErr (coreToLogic def') of
               Right e -> e
-              Left  e -> panic Nothing (show e)
+              Left  e -> panic (Just $ sourcePosSrcSpan $ loc sym) (show e)
     ref        = F.Reft (F.vv_, F.PAtom F.Eq (F.EVar F.vv_) le)
     mkErr s    = ErrHMeas (sourcePosSrcSpan $ loc sym) (pprint $ val sym) (PJ.text s)
     bbs        = filter isBoolBind xs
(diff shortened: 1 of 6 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs
```diff
@@ -352,7 +352,7 @@ makeAssumeType cfg tce lmap dm sym mbT v def
     _msg  = unwords [showpp sym, showpp mbT]
     le    = case runToLogicWithBoolBinds bbs tce lmap dm cfg mkErr (coreToLogic def') of
               Right e -> e
-              Left  e -> panic Nothing (show e)
+              Left  e -> panic (Just $ sourcePosSrcSpan $ loc sym) (show e)
```
facundominguez: Maybe `Control.Exception.throw e` would print the error without a stack trace.

facundominguez (approved):
LGTM!

PR #2474, opened 2025-01-17
Upgrade Liquid Haskell to run with ghc 9.12

facundominguez:


facundominguez:
I don't think we should merge yet while ghc-9.12.2 isn't released. See [this announcement](https://mail.haskell.org/pipermail/ghc-devs/2025-January/021915.html).

facundominguez:
ghc-9.12.2 [is out](https://mail.haskell.org/pipermail/ghc-devs/2025-March/021948.html). I updated the PR to build with it.

Issue #2438, opened 2024-11-14
PLE generates polymorphic applys for Sets


clayrat:
The combination of the following two files crashes LH:

`Foo.hs`
```haskell
{-@ LIQUID "--reflection"     @-}

module Foo where

import Prelude hiding (last)

{-@ reflect last @-}
{-@ last :: xs:{ [a] | len xs > 0 } -> a @-}
last :: [a] -> a
last [x]      = x
last (_ : xs) = last xs

data Foo t = Foo1 -- | Foo2
  deriving (Eq, Ord)

{-@ reflect isFoo @-}
isFoo :: Foo t -> Bool
isFoo Foo1 = True
-- isFoo Foo2 = False

{-@ reflect foos @-}
{-@ foos :: Foo t -> [{ v:Foo t | isFoo v }] @-}
foos :: Foo t -> [Foo t]
foos f = if isFoo f then f : [] else []
```

`Bar.hs`
```haskell
{-@ LIQUID "--reflection"     @-}
{-@ LIQUID "--ple"            @-}

module Bar where

import Prelude hiding (last)
import Data.Set
import Foo

{-@ lemma_last_isFoo :: f : Foo t -> { isFoo (last (foos f))} @-}
lemma_last_isFoo :: Foo t -> ()
lemma_last_isFoo f = if isFoo f then () else ()
```

The (prettifed) error is :

```
Unknown func-sort:
 Int : (Array (Foo Int) Bool) for

(apply    : func(0 , [[(Foo t)]; (Array_t (Foo t) bool)]))
(listElts : func(0 , [[(Foo t)]; (Array_t (Foo t) bool)]))
((apply : func(0 , [[(Foo t)]; [(Foo t)]]))
  ((apply : func(0 , [(Foo t); [(Foo t)]; [(Foo t)]]))
   (GHC.Types.: : func(0 , [(Foo t); [(Foo t)]; [(Foo t)]]))
   ((f : (Foo t)) : (Foo t)) : func(0 , [[(Foo t)];
    [(Foo t)]]))
  (GHC.Types.[] : [(Foo t)]) : [(Foo t)])
```

The bug crucially depends on having 2 files and importing `Data.Set`, merging the files or removing the import makes it disappear.

clayrat:
This seems related to the following function and its description: https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/src/Language/Fixpoint/Types/Theories.hs#L116-L143

[...]

PR #2483, opened 2025-02-06
Update to use conditional elaboration

clayrat:
Update to use https://github.com/ucsd-progsys/liquid-fixpoint/pull/734

Some caveats:

* `pos/Maps` and `ple/pos/PolySetBar` tests had to be disabled globally because they don't work with the combination of corrected theory sort signatures (https://github.com/ucsd-progsys/liquid-fixpoint/commit/55211d952aa31dbc3af134fd6830f5efb431f503) and commit https://github.com/ucsd-progsys/liquidhaskell/commit/fac8f5706c04f313b081c3ede4650b1c2f8066e4. The solution to getting them working is most likely to implement the dynamic apply generation.
* `benchmark-bytestring/Data/ByteString/Lazy` and `ple/pos/Permutations` additionally have to be disabled when running the test suite with CVC5.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ commands:
       - checkout
       - add_ssh_keys
       - run: git submodule sync
-      - run: git submodule update --init --remote
+      - run: git submodule update --init
 
   cabal_build_and_test:
     description: "Build the project and run the tests"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ commands:
       - checkout
       - add_ssh_keys
       - run: git submodule sync
-      - run: git submodule update --init
+      - run: git submodule update --init --remote
 
   cabal_build_and_test:
     description: "Build the project and run the tests"
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "liquid-fixpoint"]
 	path = liquid-fixpoint
-    url = https://github.com/ucsd-progsys/liquid-fixpoint.git
+	url = https://github.com/ucsd-progsys/liquid-fixpoint.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "liquid-fixpoint"]
 	path = liquid-fixpoint
-	url = https://github.com/ucsd-progsys/liquid-fixpoint.git
+    url = https://github.com/clayrat/liquid-fixpoint.git
+	branch = cvc5-theories
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,3 @@
 [submodule "liquid-fixpoint"]
 	path = liquid-fixpoint
-    url = https://github.com/clayrat/liquid-fixpoint.git
-	branch = cvc5-theories
+    url = https://github.com/ucsd-progsys/liquid-fixpoint.git
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,7 @@
   side effect of this change is that LH can now pick up names in scope using import aliases
   in most places (but see [#2481](https://github.com/ucsd-progsys/liquidhaskell/issues/2481)).
 - Allow to link Haskell definitions with logical primitives via `define` declarations [#2463](https://github.com/ucsd-progsys/liquidhaskell/pull/2463).
+- CVC5 solver is now supported for all logical theories, including Sets/Bags [#2483](https://github.com/ucsd-progsys/liquidhaskell/pull/2483)
 
 ## 0.9.10.1 (2024-08-21)
 
--- a/README.md
+++ b/README.md
@@ -98,7 +98,11 @@ For example:
 
     $ cabal build tests:unit-neg --ghc-options=-fplugin-opt=LiquidHaskell:--no-termination
 
-Or your favorite number of threads, depending on cores etc.
+Another useful option is to change the underlying solver:
+
+    $ cabal build tests:unit-pos --ghc-options=-fplugin-opt=LiquidHaskell:--smtsolver=cvc5
+
+You can also modify the number of used threads, depending on cores etc.
 
 You can directly extend and run the tests by modifying the files in
 
--- a/docs/mkDocs/docs/options.md
+++ b/docs/mkDocs/docs/options.md
@@ -359,16 +359,15 @@ to run a given file with full-checking, add the pragma:
 
 **Options:** `smtsolver`
 
-By default, LiquidHaskell uses the SMTLIB2 interface for Z3.
+By default, LiquidHaskell uses the SMTLIB2 interface for [Z3](https://github.com/Z3Prover/z3).
 
 To run a different solver (supporting SMTLIB2) do:
 
     $ liquid --smtsolver=NAME foo.hs
 
-Currently, LiquidHaskell supports
+Currently, LiquidHaskell additionally supports
 
-+ [CVC4](http://cvc4.cs.stanford.edu/web/)
-+ [MathSat](http://mathsat.fbk.eu/download.html )
++ [CVC5](https://cvc5.github.io/)
 
 To use these solvers, you must install the corresponding binaries
 from the above web-pages into your `PATH`.
--- a/liquid-prelude/src/Language/Haskell/Liquid/Bag.hs
+++ b/liquid-prelude/src/Language/Haskell/Liquid/Bag.hs
@@ -23,8 +23,8 @@ data Bag a = Bag { toMap :: M.Map a Int } deriving Eq
 empty :: Bag k
 empty = Bag M.empty
 
-{-@ assume get :: (Ord k) => k:k -> b:Bag k -> {v:Nat | v = Bag_count b k}  @-}
-{-@ define get               k      b                    = (Bag_count b k) @-}
+{-@ assume get :: (Ord k) => k:k -> b:Bag k -> {v:Nat | v = Bag_count k b}  @-}
+{-@ define get               k      b                    = (Bag_count k b) @-}
 get :: (Ord k) => k -> Bag k -> Int
 get k b = M.findWithDefault 0 k (toMap b)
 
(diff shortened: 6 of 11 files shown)


clayrat:
filtered
bot
top

clayrat:
The performance charts were created before rebasing on top of https://github.com/ucsd-progsys/liquidhaskell/pull/2482, so the discrepancy may be more pronounced than it really is. I will refresh them in a couple of hours.

clayrat:
filtered
bot
top

nikivazou:
Great!!!!!! Are there tests running with cvc5? I am curious how the time compares between cvc5 and z3 :)

clayrat:
The situtation with LH tests on CVC5 is not very straightforward, unfortunately. There are two issues at the moment:

* The `PolySetBar` test in the `ple-pos` group  fails with an `unknown func-sort` error similar to https://github.com/ucsd-progsys/liquidhaskell/issues/2438 (I am debugging this now)
* Some of the tests in the `benchmark-bytestring` group seem to trigger some pathological behaviour in cvc5, where the solver is left running at 100% CPU for several hours until the `smtlib-backends-process` API crashes with an `Data.ByteString.hGetLine: end of file` error.

If I exclude these two test groups, the performance seems comparable to Z3.

clayrat:
Here are performance diffs comparing the CVC5 run excluding `ple-pos` and `benchmark-bytestring` (`after`) against the Z3 run (`before`). Both runs are done on the latest commit in this PR, i.e. https://github.com/ucsd-progsys/liquidhaskell/pull/2483/commits/d5aa806d7b9224694eb6c8b0159be90d74bca3b1):
 
filtered
bot
top

clayrat:
A combination of the latest LF changes and fac8f5706c04f313b081c3ede4650b1c2f8066e4 causes the current breakage of tests. Previously we've had an incorrect definition of sort signatures (missing a type abstraction prefix) for elaborated set and bag operations in LF, which caused the apply machinery to always generate additional applys for `Array @0 Bool` and `Array @0 Int`. The tests still work with these fixed signatures on commits before fac8f5706c04f313b081c3ede4650b1c2f8066e4, but are broken after.

Looking at the commit, it seems the removal of `giUseVars` might be the underlying cause? @facundominguez

clayrat:
Changing https://github.com/ucsd-progsys/liquidhaskell/blob/develop/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Qualifier.hs#L99 to 
```haskell
`x `S.member` (S.difference (S.fromList (giUseVars (giSrc info))) (S.fromList (specAxiomVars info)))
``` 
indeed fixes the broken https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/pos/Maps.hs but breaks the newly introduced https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/pos/Meas12.hs instead.

facundominguez:
> Looking at the commit, it seems the removal of giUseVars might be the underlying cause?

Hard to say without knowing why you would need to filter qualifiers with giUseVars.

To start somewhere, `tests/pos/Maps.hs` and `tests/pos/Meas12.hs` are passing in `develop`. Could you maybe describe in some detail how they are affected by this PR?

clayrat:
The PR (or rather the underlying https://github.com/ucsd-progsys/liquid-fixpoint/pull/734) introduces many changes, but the relevant change here is in https://github.com/clayrat/liquid-fixpoint/blob/cvc5-theories/src/Language/Fixpoint/Smt/Theories.hs#L495-L509:

```diff
-  , interpSym arrMapNotS "(_ map not)" (FFunc setArrSort setArrSort)
-  , interpSym arrMapOrS  "(_ map or)"  (FFunc setArrSort $ FFunc setArrSort setArrSort)
-  , interpSym arrMapAndS "(_ map and)" (FFunc setArrSort $ FFunc setArrSort setArrSort)
-  , interpSym arrMapImpS "(_ map =>)"  (FFunc setArrSort $ FFunc setArrSort setArrSort)
+  , interpSym arrMapNotS "(_ map not)" (FAbs 0 $ FFunc setArrSort setArrSort)
+  , interpSym arrMapOrS  "(_ map or)"  (FAbs 0 $ FFunc setArrSort $ FFunc setArrSort setArrSort)
+  , interpSym arrMapAndS "(_ map and)" (FAbs 0 $ FFunc setArrSort $ FFunc setArrSort setArrSort)
+  , interpSym arrMapImpS "(_ map =>)"  (FAbs 0 $ FFunc setArrSort $ FFunc setArrSort setArrSort)

....
-  , interpSym arrMapPlusB "(_ map (+ (Int Int) Int))"        (FFunc bagArrSort $ FFunc bagArrSort bagArrSort)
-  , interpSym arrMapLeB   "(_ map (<= (Int Int) Bool))"      (FFunc bagArrSort $ FFunc bagArrSort setArrSort)
-  , interpSym arrMapGtB   "(_ map (> (Int Int) Bool))"       (FFunc bagArrSort $ FFunc bagArrSort setArrSort)
-  , interpSym arrMapIteB  "(_ map (ite (Bool Int Int) Int))" (FFunc setArrSort $ FFunc bagArrSort $ FFunc bagArrSort bagArrSort)
+  , interpSym arrMapPlusB "(_ map (+ (Int Int) Int))"        (FAbs 0 $ FFunc bagArrSort $ FFunc bagArrSort bagArrSort)
+  , interpSym arrMapLeB   "(_ map (<= (Int Int) Bool))"      (FAbs 0 $ FFunc bagArrSort $ FFunc bagArrSort setArrSort)
+  , interpSym arrMapGtB   "(_ map (> (Int Int) Bool))"       (FAbs 0 $ FFunc bagArrSort $ FFunc bagArrSort setArrSort)
+  , interpSym arrMapIteB  "(_ map (ite (Bool Int Int) Int))" (FAbs 0 $ FFunc setArrSort $ FFunc bagArrSort $ FFunc bagArrSort bagArrSort)
```

Adding the extra outer type abstraction (`FAbs 0`) hides the `FFunc` which has the effect that [`applySorts`](https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/src/Language/Fixpoint/SortCheck.hs#L918) no longer uses these signatures for populating the apply-sorts table by default. This is theoretically an improvement, as the size of that table reduces, however this unmasks the fact that there are some situations where the apply-sorts table has erroneously missing entries, as first noted in https://github.com/ucsd-progsys/liquidhaskell/issues/2438. The commit https://github.com/ucsd-progsys/liquidhaskell/commit/fac8f5706c04f313b081c3ede4650b1c2f8066e4 apparently introduces another such situation, with the apply-sorts table missing an entry for `Array Int Int` (which is an LF/SMT elaboration target sort for Haskell `Map`), resulting in the following error on https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/pos/Maps.hs:

```
Unknown func-sort: (SArray SInt SInt,SInt) for (apply : func(0 , [(Array_t int int);
                       int])) (GHC.Types_LHAssumptions.len : func(0 , [(Array_t int int);
                                                                       int])) (Maps.emp : (Array_t int int))
```

clayrat:
One potential workaround for this problem is to force [`funcSorts`](https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/src/Language/Fixpoint/Types/Theories.hs#L139) to generate all possible instantiations for `Map`s as well (the erroneous signatures existing before effectively already caused it to generate all possible instantiations for `Set`s and `Bag`s), however this solution causes a significant performance degradation (most likely by blowing up the appy-sorts table size exponentially), as seen on this chart: 

bot

facundominguez:
Only for my enlightenment, what do you think a well designed fix would be?

clayrat:
I think the proper way forward would be to try and implement dynamic `apply` generation, as described in this comment: https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/src/Language/Fixpoint/Types/Theories.hs#L116-L137 This would get rid of the pregenerated apply-sort table and the problems it causes.

However, this change seems non-trivial and could take a while, in the meantime we could restore the filtering on `giUseVars` (and demote `Meas12` back to a negative test), and then flip everything back when the dynamic solution is implemented (assuming no additional complications arise).

facundominguez:
I'm not acquainted with the sort generation at play in the PR, but AFAIU allowing for more qualifiers, as in fac8f57, is requiring more sorts to make those qualifiers well-sorted.

What the proposed workaround achieves is to make the existing tests pass, while (I sense) other tests could be written that would also make verification crash because of missing sorts. Is this really the case?

If so we might just disable the failing test, and document the limitation in terms that users can understand, since adding `giUseVars` makes the code more involved without really solving the problem. But I'm happy to discuss it more, or to be corrected :)

clayrat:
Ok, I've disabled `pos/Maps` and `ple/pos/PolySetBar` tests, here are the performance charts for Z3:

filtered
bot
top

clayrat:
For CVC5 I've additionally had to disable `benchmark-bytestring/Data/ByteString/Lazy` and `ple/pos/Permutations` tests, as these ultimately are the two triggering some pathological behaviour in the CVC5 solver itself, making it either loop forever, or run for a very long time. The comparison here is against the Z3 run, not `develop`:

filtered
bot
top

clayrat:
Updated to latest LF version and merged with `develop`

ranjitjhala:
Btw, @clayrat -- re hanging with CVC5 -- the `.liquid` directory has the generated `.smt2` file, which you can run directly on `cvc5` (or `z3` as the case may be). Did you confirm that running `cvc5` directly on the file makes it hang? If so, often I've found it useful to file an issue with the relevant solver as they are then able to fix whatever the problem was...

clayrat:
> Btw, @clayrat -- re hanging with CVC5 -- the `.liquid` directory has the generated `.smt2` file, which you can run directly on `cvc5` (or `z3` as the case may be). Did you confirm that running `cvc5` directly on the file makes it hang? If so, often I've found it useful to file an issue with the relevant solver as they are then able to fix whatever the problem was...

Yes, that's exactly what I'm doing now, I've even located the specific statement that's causing it to hang. The problem is that this happens near the end of a ~40 Mb file that LF generates, so I'm trying to figure out if there's any way to minimize it before trying to submit. However, doing so seems to change where the hanging happens :)

PR #2484, opened 2025-02-07
Split testsuites and stop linking during tests

facundominguez:
Splitting the testsuites allows machines with sufficient parallelism to run the tests faster.

In addition, linking is not necessary to run Liquid Haskell, so this PR arranges to skip it in tests.

### DIFF
--- a/tests/harness/Test/Groups.hs
+++ b/tests/harness/Test/Groups.hs
@@ -10,6 +10,8 @@ microTestGroups =
     "unit-pos-1"
   , "unit-pos-2"
   , "unit-pos-3"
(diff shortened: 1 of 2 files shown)


facundominguez:
cc @AlecsFerra

clayrat:
It looks like this PR has removed the second half of what used to be `unit-pos-1` and replicated the first half as `unit-pos-5` (instead of moving the second half under that name, which I presume was the original intention). All the tests from `PolySet` to `StateLib` are no longer run.

facundominguez:
:grimacing: thanks for checking it out!

PR #2486, opened 2025-02-12
Initial version of typed holes support under allow-typed-holes flag

matheussbernardo:
Based on [PR#1140](https://github.com/ucsd-progsys/liquidhaskell/pull/1440)

I am adding a flag --allow-typed-holes
and detecting holes which are bindings with `hole` as infix.

These holes are currently being saved in the CGInfo and a warning is being presented to the user showing the refinement type.

I added one example in the test that shows how to use this experimental feature.

### DIFF
--- a/docs/mkDocs/docs/options.md
+++ b/docs/mkDocs/docs/options.md
@@ -734,3 +734,52 @@ an import of a module coming from package `PACKAGE`. e.g.
 `--exclude-automatic-assumptions-for=vector` would stop loading
 `_LHAssumptions` modules for any imports coming from package
 `vector`.
+
+## Warn on Term Holes
+
+**Options:** `warn-on-term-holes`
+
+LiquidHaskell can be configured to emit warnings whenever it encounters incomplete terms, known as *holes*.
+
+This flag is particularly useful during development, ensuring that any placeholders in your specifications are reviewed and completed before final evaluation.
+
+To activate this behavior, use the `--warn-on-term-holes` flag either:
+
+1. As a pragma in your source file:
+    ```haskell
+    {-@ LIQUID "--warn-on-term-holes" @-}
+    ```
+2. As a plugin option:
+    ```
+    ghc-options: -fplugin-opt=LiquidHaskell:--warn-on-term-holes
+    ```
+
+When enabled, LiquidHaskell issues a warning for each term hole it discovers, for example:
+
+```haskell
+typed-holes/Example1.hs:25:13: error:
+    Hole Found
+Example1.hole :: {v : [a] | false}
+in the context
+  Example1.<> : forall a .
+                x1:[a] -> x2:[a] -> {VV : [a] | VV == Example1.<> x1 x2}
+
+  lq_anf$##7205759403792797257##dWZ : {v : [a] | v == Example1.<> lq_anf$##7205759403792797256##dWY x##aFc
+                                                 && GHC.Types_LHAssumptions.len v >= 0}
+
+  Example1.hole : forall a . a
+
+  Example1.empty : forall a .
+                   {VV : [a] | VV == Example1.empty
+                               && VV == GHC.Types.[]}
+
+  lq_anf$##7205759403792797256##dWY : {v : [a] | v == Example1.empty
+                                                 && v == GHC.Types.[]
+                                                 && GHC.Types_LHAssumptions.len v >= 0}
+
+  x##aFc : {VV##1043 : [a] | GHC.Types_LHAssumptions.len VV##1043 >= 0}
+hole found
+   |
+25 |         === hole
+   |             ^^^^
+```
--- a/docs/mkDocs/docs/options.md
+++ b/docs/mkDocs/docs/options.md
@@ -741,6 +741,21 @@ an import of a module coming from package `PACKAGE`. e.g.
 
 LiquidHaskell can be configured to emit warnings whenever it encounters incomplete terms, known as *holes*.
 
+To create a hole you need to have a binding with the name `hole` in the scope by adding:
+
+      hole = undefined
+
+To use a hole you simply insert the `hole` in any place
+where you want to get more type information. For example:
+```haskell
+hole = undefined
+
+{-@ listLength :: xs:[a] -> {v : Nat | v == len xs} @-}
+listLength :: [a] -> Int
+listLength [] = hole -- Hole inserted here
+listLength (_:xs) = 1 + listLength xs
+```
+
 This flag is particularly useful during development, ensuring that any placeholders in your specifications are reviewed and completed before final evaluation.
 
 To activate this behavior, use the `--warn-on-term-holes` flag either:
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs
@@ -101,6 +101,7 @@ consAct γ cfg info = do
   fcs <- concat <$> mapM (splitC (typeclass (getConfig info))) hcs
   fws <- concat <$> mapM splitW hws
   checkStratCtors γ sSpc
+  when (warnOnTermHoles cfg) emitConsolidatedHoleWarnings
   modify $ \st -> st { fEnv     = fEnv    st `mappend` feEnv (fenv γ)
                      , cgLits   = litEnv   γ
                      , cgConsts = cgConsts st `mappend` constEnv γ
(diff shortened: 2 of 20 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs
```diff
@@ -456,12 +467,26 @@ cconsLazyLet γ (Let (NonRec x ex) e) t
...
+  t <- if (allowTypedHoles (getConfig g)) then synthesizeWithHole else consE' g e
+  return t
+  where 
+    synthesizeWithHole = do
+      isItHole <- detectTypedHole g e
+      t <- consE' g e 
+      _ <- case isItHole of
+        Just (srcSpan, x) -> addHole (RealSrcSpan srcSpan Strict.Nothing) x t g
+        _ -> return ()
+      return t
```
facundominguez: Can this be moved to the App case of `consE`. Instead of renaming `consE` and putting the check on every call to `consE'`?

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs
```diff
@@ -295,14 +297,23 @@ addPToEnv γ π
...
+detectTypedHole  _ _ = return Nothing -- NOT A TYPED HOLE
 --------------------------------------------------------------------------------
 -- | Bidirectional Constraint Generation: CHECKING -----------------------------
 --------------------------------------------------------------------------------
 cconsE :: CGEnv -> CoreExpr -> SpecType -> CG ()
 --------------------------------------------------------------------------------
 cconsE g e t = do
-  -- NOTE: tracing goes here
-  -- traceM $ printf "cconsE:\n  expr = %s\n  exprType = %s\n  lqType = %s\n" (showPpr e) (showPpr (exprType e)) (showpp t)
+  _ <- if (allowTypedHoles (getConfig g))  then (detectTypedHole g e) else return Nothing
```
facundominguez: Can this be added as a new equation for the App case of cconsE', instead of trying to detect holes on every call to `cconsE`?

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs
```diff
@@ -319,9 +330,9 @@ cconsE' γ e t
 cconsE' γ e@(Let b@(NonRec x _) ee) t
   = do sp <- gets specLVars
        if x `S.member` sp
-         then cconsLazyLet γ e t
-         else do γ' <- consCBLet γ b
-                 cconsE γ' ee t
+        then cconsLazyLet γ e t
+        else do γ' <- consCBLet γ b
+                cconsE γ' ee t
```
facundominguez: At your convenience, please consider dropping the formatting changes to keep the review focused on the most relevant changes, or at least consider putting them in a separate commit.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs
```diff
@@ -295,14 +297,23 @@ addPToEnv γ π
...
 
+
+detectTypedHole ::  CGEnv -> CoreExpr -> CG (Maybe (RealSrcSpan, Var))
+detectTypedHole _ (App (Tick genTick (Var x)) _) | isVarHole x
+  = return (Just (getSrcSpanFromTick, x))
+    where
+      getSrcSpanFromTick = case genTick of
+        SourceNote src _ -> src
+        _ -> panic Nothing "Not a Source Note"
+      isVarHole =  L.isInfixOf "hole" . F.symbolString . F.symbol
```
facundominguez: Will this mistakenly catch variables written by the user that include "hole" in their name? e.g. whole, wholesale, etc.

To make this as precise as possible, I need to understand yet how the variable is bound.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Types.hs
```diff
@@ -240,6 +240,7 @@ data CGInfo = CGInfo
   , ghcI          :: !TargetInfo
   , dataConTys    :: ![(Var, SpecType)]                  -- ^ Refined Types of Data Constructors
   , unsorted      :: !F.Templates                        -- ^ Potentially unsorted expressions
+  , hsHoles      :: !(M.HashMap Var (HoleInfo (CGInfo, CGEnv) SpecType)) -- Information about holes
```
facundominguez: The comment should clarify that this are holes in terms, not to be confused with holes in types.

liquidhaskell-boot/src/Language/Haskell/Liquid/Liquid.hs
```diff
@@ -125,7 +125,6 @@ dumpCs cgi = do
   putStrLn $ render $ pprintMany (fixCs cgi)
   putStrLn "***************************** WfCs ********************************"
   putStrLn $ render $ pprintMany (hsWfs cgi)
-
```
facundominguez, line 129: Another formatting change to drop (?)

liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
```diff
@@ -488,6 +488,10 @@ config = cmdArgsMode $ Config {
     = def &= help "Allow refining constructors with unsafe refinements"
           &= name "allow-unsafe-constructors"
           &= explicit
+  , allowTypedHoles
+    = def &= help "Allow typed holes in implementation"
+          &= name "allow-typed-holes"
+          &= explicit
```
facundominguez: Name suggestion to disambiguate from holes in types. Also, the flag doesn't allow the holes, IIUC, it just reports them.
```suggestion
  , warnOnTermHoles
    = def &= help "Warn about holes in terms"
          &= name "warn-on-term-holes"
          &= explicit
```

facundominguez (commented):
Thanks @matheussbernardo. I left some comments. Can we have the feature documented? Maybe it could have a section in `docs/mkDocs/docs/options.md`.

Also, it would be helpful to have an example of a warning message produced by this implementation for the discussion.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs
matheussbernardo: Yes! Fixed!

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs
matheussbernardo: Removed completely from Checking. If needed later I can add again.

docs/mkDocs/docs/options.md
```diff
@@ -733,3 +733,52 @@ an import of a module coming from package `PACKAGE`. e.g.
 `--exclude-automatic-assumptions-for=vector` would stop loading
 `_LHAssumptions` modules for any imports coming from package
 `vector`.
+
+## Warn on Term Holes
+
+**Options:** `warn-on-term-holes`
+
+LiquidHaskell can be configured to emit warnings whenever it encounters incomplete terms, known as *holes*.
```
facundominguez, line 742: This is probably a good place to explain what a hole or an incomplete term is. Otherwise, I don't think the user will know that she needs to define variables named "hole".

docs/mkDocs/docs/options.md
```diff
@@ -733,3 +733,52 @@ an import of a module coming from package `PACKAGE`. e.g.
...
+1. As a pragma in your source file:
+    ```haskell
+    {-@ LIQUID "--warn-on-term-holes" @-}
+    ```
+2. As a plugin option:
+    ```
+    ghc-options: -fplugin-opt=LiquidHaskell:--warn-on-term-holes
+    ```
+
+When enabled, LiquidHaskell issues a warning for each term hole it discovers, for example:
```
facundominguez, line 772: Showing the source code here might be helpful to make sense of the error message that follows.

tests/typed-holes/Example1.hs
```diff
@@ -1,6 +1,6 @@
 {-@ LIQUID "--expect-error-containing=Hole Found" @-}
 {-@ LIQUID "--exact-data-cons" @-}
-{-@ LIQUID "--allow-typed-holes" @-}
+{-@ LIQUID "--warn-on-term-holes" @-}
```
facundominguez, line 2: This could be added to the `basic-pos` testsuite. For `scripts/test/test_plugin.sh` to pick up the module, it will need to be listed in `tests/tests.cabal`.

facundominguez: I just found that it is actually listed in `tests.cabal`. A separate executable seems overkill for a single test. It would be terser to put it in an existing executable.

matheussbernardo: I added more tests to this. I was thinking of adding all my examples to it. But maybe it is a overkill

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Generate.hs
matheussbernardo: I updated the implementation. Can you take a second look?

facundominguez:
:wave: I rebased master to evaluate this PR again.

facundominguez:
We don't have any tooling depending on this yet, but I'm merging it because the implementation is small and unlikely to be problematic. I'll link the master's thesis afterwards.

Enhancing Proof Development in LiquidHaskell: Implementation and Evaluation of Typed Holes
MATHEUS DE SOUSA BERNARDO
Chalmers University of Technology, 2025

https://odr.chalmers.se/server/api/core/bitstreams/640ee29b-b13a-44d5-9e20-200a91a11021/content

PR #2492, opened 2025-02-26
Restore tests of unit-pos-1 in unit-pos-5

facundominguez:
Accidentally deleted in #2484


facundominguez:
cc @clayrat

Maybe we need some logs and some check that tests are not skipped.

PR #2504, opened 2025-03-21
Replace various hardcoded name strings with template haskell quotes

facundominguez:


facundominguez:
I was a bit tired of having to update these strings whenever definitions moved around in base. :)

PR #2509, opened 2025-03-29
Link to latest LF

ranjitjhala:
Tracking https://github.com/ucsd-progsys/liquid-fixpoint/pull/749


nikivazou:
Shall we still merge this?

ranjitjhala:
Yes I think we should!!!

Issue #1391, opened 2018-12-28
Z3 required


mhwombat:
[Installation instructions](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/INSTALL.md) say that at least _one of_ `Z3`, `CVC4` or `MathSat` is required.

However, according to issue #1314, `Z3` is required. Like the reporter of that issue, I had `CVC4` installed, and it was findable on my path, but when I tried to run `liquid` I got an error message. Installing `Z3` solved the problem. So perhaps the documentation is incorrect?

mhwombat:
In case the information is useful, here's what I got when I had `CVC4` installed, but not `Z3`.

```
$ liquid --version
LiquidHaskell Version 0.8.4.0, Git revision 57213512a9d69093c12d644b21dbf9da95811894
Copyright 2013-18 Regents of the University of California. All Rights Reserved.
```

```
$ cat Wombat.hs
{-@ LIQUID "--no-termination" @-}
module Wombat where

{-@ type Even = {v:Int | v mod 2 = 0} @-}

{-@ weAreEven :: [Even] @-}
weAreEven     = [(0-10), (0-4), 0, 2, 666]

weAreEven  :: [Int]
```

```
$ liquid --verbose --no-termination Wombat.hs
LiquidHaskell Version 0.8.4.0, Git revision 57213512a9d69093c12d644b21dbf9da95811894
Copyright 2013-18 Regents of the University of California. All Rights Reserved.

("Module Dependencies",[Wombat])
***************************** GHC CoreBinds ***************************
[Wombat.$trModule :: GHC.Types.Module
 [LclIdX]
 Wombat.$trModule
   = scc<CAF>
     (scc<CAF>
      (scc<CAF> GHC.Types.Module)
        (scc<CAF> (scc<CAF> GHC.Types.TrNameS) "main"#))
       (scc<CAF> (scc<CAF> GHC.Types.TrNameS) "Wombat"#),
 Wombat.weAreEven :: [GHC.Types.Int]
 [LclIdX]
 Wombat.weAreEven
   = break<2>()
     GHC.Types.:
       @ GHC.Types.Int
       (scc<CAF>
        scc<CAF>
        scc<CAF>
        scc<CAF>
        break<0>()
        (scc<CAF> GHC.Num.- @ GHC.Types.Int GHC.Num.$fNumInt)
          (GHC.Types.I# 0#) (GHC.Types.I# 10#))
       (GHC.Types.:
[...]

PR #2513, opened 2025-04-08
scan for cvc5 before cvc4 + generate the error msg for missing smt

clayrat:
This is potentially related to the problem that @ulysses4ever described in #1391, although I couldn't reproduce it.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
@@ -59,7 +59,7 @@ import System.Console.CmdArgs.Implicit     hiding (Verbosity(..))
 import System.Console.CmdArgs.Text
 import GitHash
 
-import Data.List                           (nub)
+import Data.List                           (nub, intersperse)
 
 
 import System.FilePath                     (isAbsolute, takeDirectory, (</>))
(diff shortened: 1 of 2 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
```diff
@@ -524,12 +524,12 @@ withSmtSolver cfg =
...
                      Just _ -> return cfg
                      Nothing -> panic Nothing (missingSmtError smt)
-    Nothing  -> do smts <- mapM findSmtSolver [FC.Z3, FC.Cvc4, FC.Mathsat]
+    Nothing  -> do smts <- mapM findSmtSolver [FC.Z3, FC.Cvc5, FC.Cvc4, FC.Mathsat]
                    case catMaybes smts of
                      (s:_) -> return (cfg {smtsolver = Just s})
                      _     -> panic Nothing noSmtError
   where
-    noSmtError = "LiquidHaskell requires an SMT Solver, i.e. z3, cvc4, or mathsat to be installed."
+    noSmtError = "LiquidHaskell requires an SMT Solver, i.e. z3, cvc5, cvc4, or mathsat to be installed."
```
ulysses4ever: To avoid the burden of updating this line in the future, you might want to generate it. E.g. if you derived `Bounded` and `Ord` for `SMTSolver`, you could have said `show [minBound..maxBound]` instead.

liquidhaskell-boot/src/Language/Haskell/Liquid/UX/Config.hs
```diff
@@ -74,7 +74,7 @@ data Config = Config
   , minPartSize              :: Int        -- ^ Minimum size of a partition
   , maxPartSize              :: Int        -- ^ Maximum size of a partition. Overrides minPartSize
   , maxParams                :: Int        -- ^ the maximum number of parameters to accept when mining qualifiers
-  , smtsolver                :: Maybe SMTSolver  -- ^ name of smtsolver to use [default: try z3, cvc4, mathsat in order]
+  , smtsolver                :: Maybe SMTSolver  -- ^ name of smtsolver to use [default: try z3, cvc5, cvc4, mathsat in order]
```
ulysses4ever: I'd refer to the respective datatype rather than listing the options explicitly -- one fewer place to update in the future...

ulysses4ever (commented):
Thanks for looking into it!

Looking at this PR and related code, I realized that CVC5 support was added just few months ago. My project currently uses a stable release of liquidhaskell (0.9.10.1), so I haven't had a chance to test this feature, and my attempts were doomed to fail... Maybe this is also why you couldn't reproduce it.

facundominguez:
Hello @clayrat, should this be merged? Would you like to address any of the review comments in advance?

clayrat:
I tried the `Ord`+`Bounded` idea, but the fact that we have two different values for Z3 (`Z3` and `Z3mem`) will probably be confusing for the user, and I couldn't think of a quick solution for this.

clayrat:
I think I'll change the error to be generated from the lookup order list; that way, we'll have a single source of truth.

clayrat:
@facundominguez Should be ready.

ulysses4ever:
> I tried the Ord+Bounded idea, but the fact that we have two different values for Z3 (Z3 and Z3mem) will probably be confusing for the user, and I couldn't think of a quick solution for this.

Arguably, Filtering one specific value out of a list would still be a more robust solution than the status quo, even if not very elegant.

facundominguez (approved):
LGTM!

facundominguez:
> Arguably, Filtering one specific value out of a list would still be a more robust solution than the status quo, even if not very elegant.

If you want to pursue this in another PR, I think it would be a nice improvement.

PR #2518, opened 2025-04-29
Speed up CCDep.hs

AlecsFerra:


tests/ple/pos/CCDep.hs
```diff
@@ -158,8 +158,9 @@ data STACK = Stack [Ty]
 {-@ exec :: s:[Ty] -> s':[Ty] -> Prop (Code s s') -> Prop (Stack s) -> Prop (Stack s') @-}
 exec :: [Ty] -> [Ty] -> Code -> Stack -> Stack
 exec s s' (CPush _ _ n c)  i = exec (TNat : s) s' c (SCons TNat s (VNat n) i)
-exec (TNat : TNat : s)  s' (CAdd _ _ c) (SCons _ _ (VNat n) (SCons _ _ (VNat m) i)) =
-    exec (TNat : s) s' c (SCons TNat s (VNat (n + m)) i)
+exec s s' (CAdd _ _ c) i = case (s, i) of
```
facundominguez, line 164: Thanks @AlecsFerra.

Can we have a comment here explaining how this is faster/better than pattern matching on the left-hand side?

I'm also curious about the reason.

AlecsFerra: The more complicated is the pattern matching the harder is to prove totality as we split the the condition in two very easy to check problems instead of a single but very hard problem

AlecsFerra: The first version of this took a few minutes, the one we had before here took 30 sec, now it takes 4sec

facundominguez:
```suggestion
-- Some pattern matching is deferred to the right-hand side to speed up totality checks,
-- as they become simpler. The difference can be an order of magnitude in verification
-- time. See https://github.com/ucsd-progsys/liquidhaskell/pull/2518#discussion_r2066478964
exec s s' (CAdd _ _ c) i = case (s, i) of
```

Issue #2481, opened 2025-02-04
Allow to import type and predicate aliases qualified


facundominguez:
This is a spin off of #2169.

Liquid Haskell allows to express type and predicate aliases. For instance,

```Haskell
module Nat where
{-@ type INat = {v:Int | v >= 0} @-}


module Length where
import Nat

{-@ llength :: [a] -> INat @-}
llength :: [a] -> Int
llength [] = 0
llength (_:xs) = 1 + length xs
```

Unfortunately, aliases live in a flat namespace, which implies that if different modules define the same aliases, they will result in ambiguity.

```Haskell
module Nat1 where
{-@ type INat = {v:Int | v >= 0} @-}


module Nat2 where
import Data.Int (Int32)
{-@ type INat = {v:Int32 | v >= 0} @-}


module Length where
import Nat1
import Nat2

-- Nat is ambiguous in the following spec because it could be either
-- of 'Nat1.Nat' or 'Nat2.Nat'. If we try to qualify 'Nat', however,
-- we will get an error saying that the qualified name is not in scope.
{-@ llength :: [a] -> INat @-}
llength :: [a] -> Int
llength [] = 0
llength (_:xs) = 1 + length xs
```

Ideally, Liquid Haskell would honor the module aliases of import declarations. So in the above example, we can write instead
```Haskell
module Length where
import qualified Nat1 as N
import Nat2

{-@ llength :: [a] -> N.INat @-}
llength :: [a] -> Int
llength [] = 0
llength (_:xs) = 1 + length xs
```

Besides making name resolution of aliases consistent with the rest of the Liquid Haskell constructs, this would also fix #2446.

f-ei8ht:
Hi @facundominguez,

I came across this issue while exploring Qualified Aliases in Liquid Haskell for GSoC 2025. I noticed that it has been open since February 4, 2025. Is there any specific direction or existing work related to this? I’d love to understand its scope and potential challenges before considering it for my GSoC proposal.

I am very interested in this project and have recently started learning Haskell. While I may not be highly experienced yet, I am committed to learning and contributing as much as possible. Even if I don’t get selected for GSoC, I intend to put in my best effort to contribute meaningfully.

Looking forward to your insights!

Best regards,
Saif Ali Khan

facundominguez:
Hello @f-ei8ht, and welcome!

[...]

PR #2530, opened 2025-05-13
Test of the qualified import of type alises

ninioArtillero:
- **failing qualified import of type alias test**

adds example from #2481 to test suit


tests/tests.cabal
```diff
@@ -826,6 +826,31 @@ executable name-resolution-pos
...
+
+    if flag(measure-timings)
+      ghc-options:    -fforce-recomp -ddump-timings -ddump-to-file
+
+    build-depends:    base
+                    , liquidhaskell
+
+    hs-source-dirs:   app
+                    , name-resolution/neg
+
```
facundominguez: Thanks @ninioArtillero. How about adding these to `name-neg` instead of creating a new executable?

tests/name-resolution/neg/Length.hs
```diff
@@ -0,0 +1,13 @@
+{-# OPTIONS_GHC -Wno-unused-imports #-}
+-- | This is an instance of LiquidHaskell having a flat import namespace
+-- for logic names. Here, both Nat modules define the same type alias 'INat',
+-- producing an error even though we explicitly qualify it to avoid ambiguity.
+module Length where
```
facundominguez: Usually neg tests would be expected to fail and use the flag `--expect-error-containing=...`. See [tests/names/neg/DuplicatedAliases.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/names/neg/DuplicatedAliases.hs) for comparison.

ninioArtillero: got it

tests/tests.cabal
ninioArtillero: sounds good, just thought the test was better "tagged" as name resolution... but it is indeed a names issue and is similar in scope to the test you mentioned below

tests/names/neg/QualifiedAliases
```diff
@@ -0,0 +1,14 @@
+{-@ LIQUID "--expect-error-containing=Multiple definitions of Type Alias" @-}
+
+-- | This is an instance of LiquidHaskell having a flat import namespace
+-- for logic names. Here, both Nat modules export the same type alias 'INat',
+-- producing an error even though we explicitly qualify it to avoid ambiguity.
+module QualifiedAliases where
```
facundominguez: The file `QualifiedAliases` should probably be renamed to `QualifiedAliases.hs` so cabal can find it.

ninioArtillero: Hopefuly I would have catch that, but can't run the test on my NixOS machine at the moment (no ghc 9.12.2 on the current channel)... will fix that...

facundominguez: When using nixos, I've been building GHC 9.12 from sources ...

PR #2532, opened 2025-05-14
mermaid graph of LH (specs) architecture

ninioArtillero:
I thought it would be a nice addition to help prospective contributors wrap their head around the design. It is based on the referenced post and the graph from the Types/Specs module.

Wonder if its correct still, and guess it might welcome some improvements.

Also, don´t know if this is the right place to put it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,12 +331,12 @@ that can be tested with `./scripts/test/test_plugin.sh`.
 [processTargetModule]: liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs#L483
 [processModule]:       liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs#L509
 
-# Liquid Haskell Architecture
+# Module Verification Process
 
-The following graph summarizes Liquid Haskell architecture from the perspective of
-a module's specification life cycle.
+The following graph summarizes the verification process of a module by showing
+its specification life cycle.
 See the [Specs module](./liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Specs.hs)
-for detailed documentation.
+for detailed documentation on the specification stages.
 
 ```mermaid
 flowchart LR
--- a/README.md
+++ b/README.md
@@ -331,3 +331,36 @@ that can be tested with `./scripts/test/test_plugin.sh`.
 [processTargetModule]: liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs#L483
 [processModule]:       liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs#L509
 
+# Liquid Haskell Architecture
+
+The following graph summarizes Liquid Haskell architecture.
+See the [Specs](./liquidhaskell–boot/src/Language/Haskell/Liquid/Types/Specs.hs) module for documentation.
+
+```mermaid
+flowchart LR
+  subgraph Liquid Haskell
+    subgraph Inputs
+      direction TB
+      H[HashSet LifttedSpec]
+      C[CoreBinds]
+      B[BarseSpec]
+    end
+
+    subgraph Outputs
+      direction TB
+      T[TargetSpec]
+      L[LiftedSpec]
+    end
+
+    Inputs --> Outputs
+  end
+
+  Deps>DepsOfModule] --> H
+  Mod>Module.hs] --> C & B
+  A>Module_LHAssumptions] -.-> B
+  T -->|checked by| Liquid([liquid/liquidOne])
+  L --> |serialized| File>Module.hi]
+```
+
+Reference:
+[Implementing a GHC Plugin for Liquid Haskell](https://well-typed.com/blog/2020/08/implementing-a-ghc-plugin-for-liquid-haskell/).
--- a/README.md
+++ b/README.md
@@ -334,7 +334,7 @@ that can be tested with `./scripts/test/test_plugin.sh`.
 # Liquid Haskell Architecture
 
 The following graph summarizes Liquid Haskell architecture.
-See the [Specs](./liquidhaskell–boot/src/Language/Haskell/Liquid/Types/Specs.hs) module for documentation.
+See the [Specs](./liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Specs.hs) module for documentation.
 
 ```mermaid
 flowchart LR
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -331,3 +331,36 @@ that can be tested with `./scripts/test/test_plugin.sh`.
 [processTargetModule]: liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs#L483
 [processModule]:       liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs#L509
 
+# Liquid Haskell Architecture
+
+The following graph summarizes Liquid Haskell architecture.
+See the [Specs](./liquidhaskell–boot/src/Language/Haskell/Liquid/Types/Specs.hs) module for documentation.
```
facundominguez:
```suggestion
See the [Specs](./liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Specs.hs) module for documentation.
```
The hyphen character is not the ascii one :)

README.md
```diff
@@ -331,3 +331,36 @@ that can be tested with `./scripts/test/test_plugin.sh`.
...
+      L[LiftedSpec]
+    end
+
+    Inputs --> Outputs
+  end
+
+  Deps>DepsOfModule] --> H
+  Mod>Module.hs] --> C & B
+  A>Module_LHAssumptions] -.-> B
+  T -->|checked by| Liquid([liquid/liquidOne])
```
facundominguez: Grepping for `liquidOne` in the code base turns out no result. It only appears in a couple of comments. Looks like some digging is necessary to find the current call.

README.md
```diff
@@ -331,3 +331,36 @@ that can be tested with `./scripts/test/test_plugin.sh`.
...
+      T[TargetSpec]
+      L[LiftedSpec]
+    end
+
+    Inputs --> Outputs
+  end
+
+  Deps>DepsOfModule] --> H
+  Mod>Module.hs] --> C & B
+  A>Module_LHAssumptions] -.-> B
```
facundominguez: IIRC LHAssumptions kick in when collecting annotations for the dependencies. `A_LHAssumptions.hs` doesn't influence the verification of `A.hs` as this diagram might make someone think. One has to import module `A` to get the annotations in `A_LHAssumptions.hs`.

README.md
```diff
@@ -331,3 +331,36 @@ that can be tested with `./scripts/test/test_plugin.sh`.
 [processTargetModule]: liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs#L483
 [processModule]:       liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs#L509
 
+# Liquid Haskell Architecture
```
facundominguez: I'd expect to see `liquid-fixpoint` in such a diagram, but we could consider that a future refinement :)

README.md
ninioArtillero: typopunct-mode messing around, thanks for spotting it!

README.md
ninioArtillero: I'm guessing that `liquidOne` might be a function interfacing with `liquid-fixpoint`. I will look into it...

README.md
ninioArtillero: I see. Then this might better point to the LiftedSpecs of the dependencies of A as the interface files do?

README.md
ninioArtillero: This goes in the next update.

ninioArtillero:
I gave `Module` the name `A` and added some`.hi` to make clearer where things go and come from.

Now, instead of `liquidOne` I just refer generally to `liquid-fixpoint`, as I understand it is where the verification takes place.

I was thinking about assigning specific function calls to some links, but his would imply some maintenance burden. So my take is to be as general as possible, avoiding implementation details.

By the way, graphs can be rendered and tweaked on this web editor: https://mermaid.live

README.md
ninioArtillero: Maybe it's better pointing to `liquid-fixpoint` here, as that's were the verification takes place, instead of the specific function call that it is subject to change.

README.md
ninioArtillero: What do you think if I change the title to "Specification life cycle" to avoid a misleading expectation?

ninioArtillero: or "Module verification process" of something like that? Not sure which is more accurate.

facundominguez: Maybe "Module verification process" or "Module verification flow" is the most accurate description.

facundominguez:
> Now, instead of liquidOne I just refer generally to liquid-fixpoint, as I understand it is where the verification takes place.

Once `TargetInfo` is constructed, it seems to be given to `Liquid.checkTargetInfo` which will generate verification constraints and feed those to `liquid-fixpoint`. The type that holds the generated constraints is `FInfo` if you want to grep it in Liquid.hs, and `FInfo` is provided by `liquid-fixpoint`.

README.md
facundominguez: I think the latest fixes to the diagram are good enough :+1:

ninioArtillero:
I had the confusion that when `makeTargetSpec` returned the module's `TargetSpec` and `LiftedSpec` constructed by `processModule` verification had already taken place, so I disregarded the call stack towards `Liquid.checkTargetInfo`. So, if I understand correctly, we already have the `LiftedSpec` before the constraint verification of the `TargetSpec`, and after it is _cleared_ by `checkTargetInfo` it is returned for serialization by `checkLiquidHaskellContext`.

facundominguez:
That sounds right.

ninioArtillero:
I removed the link connecting liquid-fixpoint to the LiftedSpec to not give the impression that it is its output. Made explicit that serialization follows only if verified.

I would set for this version unless you find anything worth tweaking further in this PR @facundominguez .

README.md
```diff
@@ -331,3 +331,44 @@ that can be tested with `./scripts/test/test_plugin.sh`.
...
+        Dependencies of A
+        LifttedSpecs
+        ]
+      C[CoreBinds]
+      B[BarseSpec]
+    end
+
+    subgraph Outputs
+      direction TB
+      T[TargetSpec]
```
facundominguez: If `liquid-fixpoint` is to stay in the diagram, the output of LF would be the verification contraints rather than the TargetSpec.
```suggestion
      T[Verification
Constraints]
```
This is spreading a bit beyond `Specs.hs`. `Liquid.hs` has the constraint generation part, and `SpecFinder.hs` has the part about finding dependencies and LHAssumptions.

Alternatively, both TargetSpec and verification constraints could be left in the diagram.

facundominguez (approved):
LGTM!

PR #2538, opened 2025-05-19
Fix for T2535

### LINKED ISSUE #2535: SMT crashes on reflected functions on polymorphic data types
```
{-@ LIQUID "--reflection" @-}
{-@ LIQUID "--save" @-}

module Test where 

{-@ f :: {v:AB (T a)  | check v} -> {v:AB (T a)  | check v} @-}
f ::  AB (T a) -> AB (T a) 
f x = g x 

data T a = T a  
data AB a  = A Int | B Int
{-@ data AB a  = A Int | B Int @-}

{-@ measure check @-}
check :: AB ps -> Bool
check (A _) = True
check (B _) = False 

{-@ reflect g @-}
{-@ g :: {v:AB a | check v} -> {v:AB a | check v} @-}
g :: AB a -> AB a
g (A x) = A x 
g (B x) = B x  
```

Crashes with 

```
<no location info>: error:
    :1:1-1:1: Error
  crash: SMTLIB2 respSat = Error "line 3 column 127416: Sorts (Test.AB (Test.T Int)) and (Test.AB Int) are incompatible"
```

Because, reflecting `g` generates the mal-sorted binding 

```
(= VV$35$$35$F$35$$35$19 (ite (is-Test.A VV$35$$35$F$35$$35$19) ((as Test.A (Test.AB Int))...
```
instead of 
```
(= VV$35$$35$F$35$$35$19 (ite (is-Test.A VV$35$$35$F$35$$35$19) ((as Test.A (Test.AB (T Int)))...
```

nikivazou:
The goal is to fix #2535 but it seems it trigered another bug. 

Concretely, for the type: 

```
g :: forall k (a::k). AB a -> AB a
```

The function [isKind](https://github.com/ucsd-progsys/liquidhaskell/blob/71f0292c7b772f5850df4c0ef7536cfe18796006/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Misc.hs#L156) k, was returning false. 

The fix is temporal, since I believe the [isTYPEorCONSTRAINT](https://hackage.haskell.org/package/ghc-9.12.1/docs/src/GHC.Core.Type.html#isTYPEorCONSTRAINT) function in GHC-9.12.2 is wrong...


liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Misc.hs
```diff
@@ -153,9 +154,13 @@ mapTyVars _ hsT lqT
...
        throwError (err (F.pprint hsT) (F.pprint lqT))
 
+-- Niki believes there is a bug on the isTYPEorCONSTRAINT function in GHC-9.12.2
+-- Hardcoding "GHC.Types.Type" for now because typeKind in in the Prim module... 
 isKind :: Kind -> Bool
-isKind = isTYPEorCONSTRAINT -- TODO:GHC-863 isStarKind k --  typeKind k
-
+isKind k = isTYPEorCONSTRAINT k -- TODO:GHC-863 isStarKind k --  typeKind k
+          || case k of 
+                TyVarTy kk -> showPpr (varType kk) == ghcTypeStr
```
facundominguez, line 162: It would be better not to hardcode the string, as GHC sometimes moves definitions and the string changes, requiring debugging to catch it.

`varType kk == GHC.Builtin.Types.liftedTypeKind` would be cheaper to maintain.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Misc.hs
```diff
@@ -153,9 +154,13 @@ mapTyVars _ hsT lqT
   = do err <- gets errmsg
        throwError (err (F.pprint hsT) (F.pprint lqT))
 
+-- Niki believes there is a bug on the isTYPEorCONSTRAINT function in GHC-9.12.2
+-- Hardcoding "GHC.Types.Type" for now because typeKind in in the Prim module...
```
facundominguez, line 158: Is there an issue reported in GHC? Otherwise, what is the problem referred here?

tests/tests.cabal
```diff
@@ -623,7 +623,7 @@ executable errors
                     -- , ExportReflect0
                     , Fractional
                     , HigherOrderBinder
-                    , HintMismatch
+                    -- , HintMismatch
```
facundominguez: Should the file `tests/errors/HintMismatch.hs` be removed as well?

facundominguez (commented):
:wave: Hello Niki! The PR looks good to me, though I share some requests to make the code easier to maintain.

Also, I think you will want to list `tests/pos/T2535.hs` in `tests.cabal` or it won't be tested when running tests.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Misc.hs
facundominguez, line 158: I don't understand the comment about `typeKind` being in some module `Prim`, but there is a function `GHC.Core.Type.typeKind`.

PR #2540, opened 2025-05-20
reflection of quot and rem

### LINKED ISSUE #1447: `div` and `mod` have interpretations in SMT, but `quot` and `rem` do not
You can't reflect functions which use `quot` and `rem`, but using `div` and `mod` works fine:
```haskell
{-@ LIQUID "--reflection" @-}

{-@ reflect intId @-}
intId :: Int -> Int
intId x = (x `quot` 2) * 2 + (x `rem` 2)

{-@ reflect intId' @-}
intId' :: Int -> Int
intId' x = (x `div` 2) * 2 + (x `mod` 2)
```

I'd fix this myself but I don't know z3 well enough to know which functions to use: `quot` and `rem` differ from `div` and `mod` only in that they round towards 0, rather than towards -infty.

ninioArtillero:
This PR aims to fix #1447

I start by adding a negative test showing the offending behavior.
It is on `/test/reflect/neg/`, but welcome suggestions if there is a more appropriate place.

A lead: both `quot` and `rem` functions are opaque-reflected, while `div` and `mod` are not.


ninioArtillero:
I implemented `quot` and `rem` in the refinement logic and added `define abs x = if x >= 0 then x else -x` to `src/GHC/Num_LHAssumptions.hs`.
I think the implementation of both are correct, but still need to verify it. how should I go about it? Should I implement some unit tests? If so, where? 

Also, I'm still to run all tests to look for regressions (or does ci/circleci does this already?).

facundominguez:
Thanks @ninioArtillero. CI might catch some regressions indeed. For tests, you could add a module under "tests/" with concrete examples like
```Haskell
{-@ exQuot1 :: { -2 = quot -5 2 } @-}
exQuot1 :: ()
exQuot1 = ()
```
and/or create a testsuite in the liquidhaskell package for property based tests.

ninioArtillero:
For the second option, you mean something like what I just pushed? I would add all the properties as post-conditions of that same function.

ninioArtillero:
As for the first option (creating a module at `/tests`). Shouldn´t it be placed into a group for the driver? 

Otherwise, if I'm not mistaken, verification would have to be done by compiling the module directly using
 `cabal exec -- ghc -fplugin=LiquidHaskell tests/QuotRem.hs`
Is it so?

I'm trying to get some general understanding about how to decide appropriate places for tests.

facundominguez:
> For the second option, you mean something like what I just pushed? I would add all the properties as post-conditions of that same function.

Not really. It would have to be a test suite using [QuickCheck](https://hackage.haskell.org/package/QuickCheck), to test for the equivalence of `quot` and `rem` in Haskell and in the logic. But we can skip this and just stick with a module under `tests/`.

tests/reflect/pos/QuotRemInt.hs
```diff
@@ -0,0 +1,23 @@
...
+intId' :: Int -> Int
+intId' x = (x `div` 2) * 2 + (x `mod` 2)
+
+{-@ sumId :: x : Int -> y : Int -> {z : Int | z = x + y}@-}
+sumId :: Int -> Int -> Int
+sumId x y =  intId x + intId y
+
+{-@ sumId' :: x : Int -> y : Int -> {z : Int | z = x + y}@-}
+sumId' :: Int -> Int -> Int
+sumId' x y =  intId' x + intId' y
```
facundominguez: These could be tested more directly with
```
{-@ lemmaQuotRem :: x:Int -> { intId x == x } @-}
lemmaQuotRem :: Int -> ()
lemmaQuotRem _ _ = ()

{-@ lemmaDivMod :: x:Int -> { intId' x == x } @-}
lemmaDivMod :: Int -> ()
lemmaDivMod _ _ = ()
```

Other tests to include here:
```
{-@
examplesQuot
   :: { -2 = quot -5 2
        && -2 = quot 5 -2
        && 2 = quot -5 -2
        && 2 = quot 5 2
        && 0 = quot 0 2
      }
@-}
examplesQuot :: ()
examplesQuot = ()
```
and similarly for `rem`.

tests/tests.cabal
```diff
@@ -840,6 +840,7 @@ executable reflect-pos
 
     other-modules:
                       DoubleLit
+                    , QuotRemInt
```
facundominguez, line 843: Yes, I think this is a fine place. :+1:

ninioArtillero:
> Not really. It would have to be a test suite using [QuickCheck](https://hackage.haskell.org/package/QuickCheck), to test for the equivalence of `quot` and `rem` in Haskell and in the logic. But we can skip this and just stick with a module under `tests/`.

I have played around with QuickCheck and think I can implement the tests. As there are no custom data types involved, guess it should not be much of a problem.

ninioArtillero:
Decided to go by your suggestion for the time being, testing both for the expected behavior (correct reflection) using the lemmas and the values containing the concrete examples in the same module. But, I'm getting the following error (for the divMod lemma alone—which should be ok— and also for the other functions) and haven´t been able to figure it out:

It would seem that non of the properties hold, but should. My previous test using `sumId` was not actually using the reflected functions explicitly (which is probably related to the problem).

```
reflect/pos/QuotRemInt.hs:26:1: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : {v : () | v == GHC.Tuple.()}
    .
      VV : {VV##883 : () | QuotRemInt.intId' ?a == ?a}
    .
    in the context
      ?a : GHC.Types.Int
    Constraint id 6
   |
26 | lemmaDivMod  _ = ()
   | ^^^^^^^^^^^^^^^^^^^

Error: [Cabal-7125]
Failed to build exe:reflect-pos from tests-0.1.0.0.
```

ninioArtillero:
I implemented `quotRem` in terms of `divMod` in the test-suite to test the implementation of `quot` and `rem` at `src/GHC/Real_LHAssumptions.hs` (kind of a inverse-reflect experiment). QuickCheck tests suggest implementation is correct.

Nevertheless, I found some specific cases that prevented verification of my original `reflect-pos` test, and I think I have found a bug: logic `/` and `mod` don´t behave as Haskell's `div` and `mod` when both arguments are negative as implied by the existing `define`s at said module.

src/GHC/Real_LHAssumptions.hs
```diff
@@ -49,6 +50,12 @@ class (Real a, Enum a) => Integral a where
 
 define div x y        = (x / y)
 define mod x y        = (x mod y)
+define quot x y =  if x >= 0
+                   then (if y >= 0 then x / y else -(x / abs y))
+                   else - div (abs x) y
```
facundominguez: This should use `/` instead of `div`. Otherwise, when `div` is changed to behave as Haskell's, `quot` behavior could be affected.
```suggestion
                   else -(abs x / y)
```

liquidhaskell.cabal
```diff
@@ -87,3 +87,18 @@ library
...
+  main-is:          Main.hs
+  other-modules:    QuotRem
+  build-depends:    base < 5,
+                    liquidhaskell,
+                    QuickCheck
+
+--  ghc-options:      -fplugin=LiquidHaskell
+--                      -fkeep-going
+--                      -O0
+--                      -no-link
```
facundominguez: Remove unneeded comments
```suggestion
```

test/QuotRem.hs
```diff
@@ -0,0 +1,57 @@
...
+  case signum (a * b) of
+    1 ->
+      if abs (a - b * q) < abs b && (b * q <= a)
+        then (q, a - b * q)
+        else divModIter (q + 1) a b
+    -1 ->
+      if abs (a - b * q) < abs b && (b * q <= a)
+        then (q, a - b * q)
+        else divModIter (q - 1) a b
+    0 -> if b == 0 then error "divide by zero" else (0, 0)
```
facundominguez: Looks equivalent to the shorter
```suggestion
divModIter q a 0 = error "divide by zero"
divModIter q 0 b = (0, 0)
divModIter q a b =
  if abs (a - b * q) < abs b && (b * q <= a)
    then (q, a - b * q)
    else divModIter (q + signum (a * b)) a b
```

test/QuotRem.hs
```diff
@@ -0,0 +1,57 @@
+{-@ LIQUID "--reflection" @-}
```
facundominguez: You could enable LH on this module only with
```Haskell
{-# OPTIONS_GHC -fplugin=LiquidHaskell #-}
```

test/QuotRem.hs
```diff
@@ -0,0 +1,57 @@
...
+                              -> {z:(a,a) | fst z = x / y && snd z = x mod y}@-}
+-- | A Haskell implementation of logic @/@ and @mod@.
+-- Most notably, `mod` is always positive.
+divModSMT :: (Integral a) => a -> a -> (a, a)
+divModSMT = divModIter 0
+
+{-@ divModIter :: (Integral a) => q:a
+                               -> x:a
+                               -> {y:a | y /= 0}
+                               -> {z:(a,a) | fst z = x / y && snd z = x mod y}@-}
```
facundominguez: The spec doesn't give a clue of the role of `q`. Something like the following perhaps works. But I don't think you need to improve this code if LH is happy with your spec.

```Haskell
divNat :: Nat -> {d:Int| d>0} -> Nat
divNat a 0 = error "div by zero"
divNat a b
    | a > b  = 1 + divNat (a - b) b
    | otherwise = 0

divSMT : a:Int -> {b:Int| d != 0 } -> {v:Int| v = a / b }
divSMT a b = ... in terms of divNat ...

divModSMT :: a:Int -> {b:Int| d != 0} -> {v:_| fst v = a / b && snd v = a mod b }
divModSMT a b = let q = divSMT a b in (q, a - q)
```

test/Main.hs
```diff
@@ -0,0 +1,24 @@
...
+
+main :: IO ()
+main = do
+        quickCheck prop_quotRemAltEuclideanDomain
+        quickCheck prop_quotRemAlt
+
+-- | Throws error if a test fails. Use this alternative to abort build
+-- in case of failure when test condition is enabled.
+main' :: IO ()
+main' = do
```
facundominguez: This is a good observation. In Haskell QuickCheck is used with [tasty](https://hackage.haskell.org/package/tasty) and [tasty-quickcheck](https://hackage.haskell.org/package/tasty-quickcheck) or similar scaffolding for tests.

I've found these tests can be integrated in `tests:test:tasty`. It is a testsuite inside the `tests` package. If you want, you can look into that. Otherwise I could do it too.

facundominguez (commented):
Bravo @ninioArtillero! Here go some comments.

test/QuotRem.hs
ninioArtillero: that's neat! thanks!

test/QuotRem.hs
ninioArtillero: it is actually complaining a lot, which is why I had the plugin commented in the cabal file.. until this point where is time to have LH verify this and wrap up.

ninioArtillero: think I will go with your suggestion, but I'm a bit uneasy about not understanding why I couldn't make verification pass.

ninioArtillero: by the way, is there any difference between using `!=` rather than `/=`?  I'm  guessing the former is used to define the latter in the logic.

facundominguez: It seems they are interchangeable. Grepping in liquid-fixpoint yields this:
```
$ git grep "PAtom Ne"
src/Language/Fixpoint/Parse.hs:     <|> (reservedOp "!=" >> return (PAtom Ne))
src/Language/Fixpoint/Parse.hs:     <|> (reservedOp "/=" >> return (PAtom Ne))
```

ninioArtillero: also, is it better or needed to have my functions monomorphic for verification? Or can the Haskell types and the refinements be kept using the `Integral` constraint?

test/Main.hs
ninioArtillero: I could look into that after having my test pass LH verification.

test/QuotRem.hs
ninioArtillero: Used your suggested implementation with the explicit termination metric.

ninioArtillero:
Used an alternative implementation proposed by @facundominguez that uses an explicit termination metric to make verification pass. Now both the property tests and LH verification pass, so we can be confident that the logic implementation of `quot` and `rem` is fine.

But the `lemmaQuotRem` at the `reflect-pos` test is complaining, even though the test examples work ok :persevere:

ninioArtillero:
The regression was introduced at `7b648be9e` where the definition of `quot` and `rem` at `src/GHC/Real_LHAssumptions.hs` was changed to the match the implementation proved correct at the `test/QuotRem`.

test/QuotRem.hs
```diff
@@ -2,9 +2,14 @@
...
 -- | Properties for testing the definitions of @quot@ and @rem@ found at
 -- @liquidhaskell/src/GHC/Real_LHAssumptions.hs@.
-module QuotRem (prop_quotRemAltEuclideanDivision, prop_quotRemAlt) where
+module QuotRem (prop_divModSMTEuclideanDivision, prop_quotRemAltEuclideanDivision, prop_quotRemAlt) where
 import Test.QuickCheck ( (==>), Property )
 
+{-@ ignore prop_divModSMTEuclideanDivision @-}
+prop_divModSMTEuclideanDivision :: Int -> Int -> Property
+prop_divModSMTEuclideanDivision a b = b/= 0 ==> a == d * b + m && abs m < abs b
+  where (d,m) = (div a b, mod a b)
```
ninioArtillero: This should be `divSMT` and `modSMT`. Corrected in the next commit.

facundominguez:
I'm setting up to merge. We need an issue to investigate the negative test.

Thanks @ninioArtillero!

facundominguez:
hm, the neg test started failing after 280c0e1. I thought I had tested that before.

ninioArtillero:
Thanks for the tuning  and merging @facundominguez ! I'll open the issue.

PR #2542, opened 2025-05-21
Fix for #2535

facundominguez:
This PR is redoing #2538, which was reverted. @nikivazou, please feel free to work on this branch as needed.


nikivazou:
It is ready to merge!

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Plugged.hs
```diff
@@ -282,7 +282,9 @@ goPlug tce tyi err f = go
     -- zipWithDefM: if ts and ts' have different length then the liquid and haskell types are different.
     -- keep different types for now, as a pretty error message will be created at Bare.Check
     go (RApp _ ts _ _)  (RApp c ts' p r)
-      | length ts == length ts'            = RApp c     (Misc.zipWithDef go ts $ Bare.matchKindArgs ts ts') p r
+    -- Removing length check because Haskell types can contain kinds which is safe based on the comment above 
+    --   | length ts == length ts'            
+      = RApp c     (Misc.zipWithDef go ts $ Bare.matchKindArgs ts ts') p r
```
facundominguez, line 287: This could be a good time to document what `goPlug` is supposed to do.

facundominguez (commented):
:+1: I don't find other things to improve about the code itself, though I left a suggestion for improving documentation.

Probably this should be merged after merging https://github.com/ucsd-progsys/liquid-fixpoint/pull/753 and updating the `liquid-fixpoint` submodule.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Plugged.hs
nikivazou: Well, `plug` is definitely replacing the holes (`_`) with their actual types, but to do it gets the type information from the GHC types, thus it checks that the two are compatible. 

I agree that it would be great to get cleaned/documented, but since this PR fixes an actual bug, I suggest we do it later :) 

But I will minimize the fixpoint example!

nikivazou:
Shall we merge? I simplified the fixpoint example!

PR #2550, opened 2025-06-04
Qualified type aliases support

ninioArtillero:
I began by manually following the call stack of the plugin to find relevant places where
aliases are manipulated.

Started with my GSoC 2025 proposal's solution strategy by changing the `rtName` field of the type for aliases from
`Symbol` to `LHName` and fixing compiler warnings to give the `Symbol`s back where they are expected.

This is work towards #2481


ninioArtillero:
tracking current error in [this branch](/ninioArtillero/liquidhaskell/tree/tracking-resolution-error)

ninioArtillero:
After many experiments, some error tracing, and much code analysis, a comment on the progress is due:

* Some local functions—used to build an `InScopeNonReflectedEnv` at `makeLogicEnvs`—were moved to the top level in https://github.com/ucsd-progsys/liquidhaskell/pull/2550/commits/98d3d51f91ce4e4012e689c4aafba9e4446bd6e3. This is the first step in the **updated solution strategy** to allow qualify type aliases: reusing the implementation of  `InScopeNonReflectedEnv`, which includes the _aliases a module is imported with_ and has a lookup function that matches against both qualified and unqualified names (providing close matches in case of failure to report to the user).
 
* In https://github.com/ucsd-progsys/liquidhaskell/pull/2550/commits/6fd94aabf82fa6e7410d59bda41e8f11b1ee35d6 the `InScopeNonReflectedEnv` was parameterized to an `InScopeEnv a` to reimplement the key-value map produced by `collectTypeAliases`. The new parameter lets us include the type alias representation of type `RTAlias Symbol ()` (where the unit comes from emptying its body) in the environment, which is needed just before the first name resolution pass—when fixing the expression arguments in `fixExpressionArgsOfTypeAliases`—.[^1] Now `lookupInScopeNonReflectedEnv` is also used for type alias lookup during the first name resolution pass (implemented as a generic traversal of the spec using `resolveLHName`). After this change, `liquidhaskell` failed to compile and complained about not being able to resolve a `PtrV` alias in `src/Foreigh/Concurrent_LHAssumptions.hs`.

* It turned out that the original implementation depends on bringing along—in the `TargetDependencies` build at `...GHC.Plugin.loadDependencies`—some transitive dependencies that are queried when collecting the aliases (and perhaps for other purposes, which I'm looking into). https://github.com/ucsd-progsys/liquidhaskell/pull/2550/commits/8441f2e072d7dd8c8a416e42de0563e9489373e9 made this explicit by giving an empty module alias to names coming from a transitive dependency to prevent them from mapping over an empty list in `mkAliasEnv` and getting discarded—. But as @facundominguez  pointed out:

  > If I understand correctly, the type aliases should be in scope in the modules that are imported directly. Once imported, type aliases would be accumulated transitively. If T exports a type alias that has been imported from a transitive dependency, the following should work:
  > 
  > ```
  > import qualified T as Q
  > 
  > {-@ f :: Q.TypeAlias @-}
  > f :: Int
  > f = 0
  > ```
  > 
  > Using a empty module alias gives a behavior that the user would not expect.
  
  My opinion was that having an empty module alias for transitive dependencies reflected the fact that type aliases of transitive dependencies are not included in the `LiftedSpec` of the imported module, but rather are accessed directly from the specs of those dependencies themselves (as this change fixing the error mentioned in the previous point seems to suggest). As far as we could tell by reading the code that builds the `LiftedSpec`, that is not the case. I'm currently investigating this matter.
* https://github.com/ucsd-progsys/liquidhaskell/pull/2550/commits/fab4891848509b6f19acebc09cc4170f09aa4503 fixes a bug that prevented an expected name ambiguity from arising in a regression test (namely, `names-neg`), while https://github.com/ucsd-progsys/liquidhaskell/pull/2550/commits/a6a83578ace10ee6fc48625d271cec89abd2780b was introduced as a safeguard, given that at that point no unresolved names are expected.

We speculate that a comprehensive solution to this matter would be to have a mechanism to explicitly export and re-export aliases (or other LH specs). Nevertheless, that approach exceeds the scope of my GSoC project and is left for future work.

Going forward, I'll make sure that aliases from transitive dependencies actually populate a module's `LiftedSpec` (which is the one used by clients for verification) and that local definitions can shadow them.

[^1]: The reason is that the parser initially parses the type aliases arguments as types, so the current implementation fixes that at this point.

ninioArtillero:
Now that this PR is ready for review, I'll provide an overview of the proposed changes:

In brief, we can now qualify aliases to disambiguate cases where two imports export type aliases with the same name. The approach we followed succeeded: by using the `InScopeEnv` for alias symbol lookup, we are now able to resolve qualified type aliases.

It turns out type alias _environments_ are created at four different stages:

1. By `collectTypeAliases` at the start of name resolution (`resolveLHNames`) to resolve variable names upfront, fix the term/value arguments of the aliases themselves (since all such arguments are initially parsed as type arguments), and finally—during the first name resolution pass—for resolving type constructor names.
2. By `makeRTEnv` during the construction of the `TargetSpec` (in `makeGhcSpec0`) to expand all type alias uses.
3. By `myRTEnv`, to prepare the type aliases to be added into the `LiftedSpec`. Here, both expression (predicate) and type aliases are normalized and filtered to include only locally defined aliases.
4. During `checkTargetSpec`, to throw an error on duplicate alias names among all known aliases (from the current module’s `BareSpec` and the `LiftedSpec`s of dependencies).

In all these cases, the environments were built from the aliases of the current module and all its transitive dependencies. For the qualified alias approach to work, I had to make changes at all of these sites. Accordingly:

1. We now use an `InScopeEnv` for symbol lookup (via `lookupInScopeNonReflectedEnv`) to resolve qualified names. All aliases originate either from the current module or a direct import. See `liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs`.
2. I decoupled the expansion of type and expression aliases. Now, for type aliases, we use `LHName`s as keys for an `AliasTable'` (used to order the aliases for expansion and check for cyclic dependencies among them) and, most importantly, for the `BareRTEnv`. This last environment is used to expand aliases at all use sites, ultimately by calling `expandBareType`. See `liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs`.
3. Only expression aliases are filtered to include just those defined locally (for now), while all type aliases are gathered, ensuring we keep only one per unqualified name. Among these ambiguous cases, we keep the locally defined one or the last—according to the lexicographic ordering of the modules they are imported from—. In other words, even though we may have used any such aliases when creating the specification, only one of them is stored in the resulting spec. This is reasonable compromise until a mechanism exists for explicitly exporting and importing type aliases. See `liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs`.
4. We only check for duplicate type alias definitions within the current module. See `liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Check.hs`.

## To summarize:

Previously, all transitive dependencies were considered during `collectTypeAliases` to build an environment (a symbol hash map) for name resolution. This posed no issue, as the (unqualified) symbol was intended to uniquely identify each alias. In this setting, aliases from dependencies could shadow local aliases—or aliases from other dependencies—during name resolution based on the accidental order in which their symbols were concatenated when the relevant hash map was created. This ambiguity would be caught later at `checkTargetSpec`: If no duplicate name was detected, only the locally defined aliases were stored in the final `LiftedSpec`.

Now, we ensure that _all_ type aliases from transitive dependencies are included in the `LiftedSpec`, so that during name resolution we gather type aliases from direct imports only. This allows us to disambiguate cases where two imports export an alias with the same name by qualifying its use (with the module name or a qualified import alias). Nevertheless, we had to make a compromise due to the lack of a mechanism for explicitly exporting and importing type aliases: when storing an ambiguous alias into the `LiftedSpec`, we give preference to the locally defined or pick the one from the lexicographical last module (similar to how conflicting _measures_ are handled).

tests/names/pos/ImportedTypeAlias.hs
```diff
@@ -0,0 +1,68 @@
...
+--       Nat1      Nat2      \
+--         \        /         \
+--          \      /           \
+--   QualifiedTypeAliases   TypeAliasShadowing
+--                \              /
+--                 \            /
+--                  \          /
+--                ImportedTypeAlias
+--
+-- The sole export of 'NatFoo' is a type alias of the same same.
```
facundominguez:
```suggestion
-- The sole export of 'NatFoo' is a type alias of the same name.
```

tests/names/pos/ImportedTypeAlias.hs
```diff
@@ -0,0 +1,68 @@
...
+-- As there is no conflicting name anywhere, no ambiguity arises even though
+-- its brought into scope from distinct imports. So the transitive accumulation
+-- of aliases is consistent. In this case, both qualified or unqualified ocurrances
+-- correspond to the same alias.
+--
+-- Both @Nat*@ modules define the same type alias @INat@.
+-- The implementation is such that in this case, only the definition from the
+-- last module (lexicographically) is exported by 'QualifiedTypeAliases'.
+-- So here, the only the definition of @INat@ from @Nat2@ refining 'Int32' is
```
facundominguez:
```suggestion
-- So here, only the definition of @INat@ from @Nat2@ refining 'Int32' is
```

docs/mkDocs/docs/specifications.md
```diff
@@ -377,10 +377,27 @@ see [tests/pos/Map.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/develo
...
+   This makes possible to effectively _update_ an alias definition.
+4. When imports contain aliases with identical symbols, name resolution triggers
+   an error upon ambiguous usage. Resolve this by qualifying either:
+   - The usage
+   - The import
+   - Or both
+5. When the same alias name is imported from multiple modules (with no local definition),
+   only the lexicographically last instance is exported (i.e., included in the resulting specification).
+
+see [tests/names/pos/ImportedTypeAlias.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/names/pos/ImportedTypeAlias.hs)
```
facundominguez:
```suggestion
See [tests/names/pos/ImportedTypeAlias.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/names/pos/ImportedTypeAlias.hs) for examples.
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -212,7 +212,11 @@ makeGhcSpec0 cfg ghcTyLookupEnv tcg instEnvs lenv localVars src lmap targetSpec
   tycEnv <- makeTycEnv1 env (tycEnv0, datacons) coreToLg simplifier
   let tyi      = Bare.tcTyConMap   tycEnv
   let sigEnv   = makeSigEnv  embs tyi (_gsExports src) rtEnv
+  -- This spec is used to add lifted measures.
   let lSpec1   = makeLiftedSpec1 cfg src tycEnv lmap mySpec1
+  -- 'mySpec' and 'specs' contain the result of the first lifting stanges, see [NOTE]: REFLECT-IMPORTS
```
facundominguez:
```suggestion
  -- 'mySpec' and 'specs' contain the result of the first lifting stages, see [NOTE]: REFLECT-IMPORTS
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -1418,42 +1442,48 @@ locFile = Misc.fst3 . F.sourcePosElts . F.sp_start . F.srcSpan
 
 
 --------------------------------------------------------------------------------
--- | @myRTEnv@ slices out the part of RTEnv that was generated by aliases defined
---   in the _target_ file, "cooks" the aliases (by conversion to SpecType), and
---   then saves them back as BareType.
+
+-- | @myRTEnv@ slices out the part of RTEnv that was generated by expression
+--   aliases defined in the _target_ file, removes type aliases with duplicate
+--   unqualified names, and "cooks" the type aliases by conversion to SpecType
```
facundominguez: Is the function really removing type aliases with duplicate names? What's the bit of code that removes the duplicates?

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -1418,42 +1442,48 @@ locFile = Misc.fst3 . F.sourcePosElts . F.sp_start . F.srcSpan
...
+    eAs           = filter (isLocInFile srcF . rtName) $ myAliases exprAliases
+    myAliases fld = M.elems . fld $ rtEnv
+    srcF          = _giTarget    src
+    name          = _giTargetMod src
+
+mkRTE :: [RTAlias x a] -> [RTAlias F.Symbol F.Expr] -> RTEnv x a
 mkRTE tAs eAs   = RTE
   { typeAliases = M.fromList [ (aName a, a) | a <- tAs ]
-  , exprAliases = M.fromList [ (aName a, a) | a <- eAs ]
+  , exprAliases = M.fromList [ (getLHNameSymbol $ aName a, a) | a <- eAs ]
```
facundominguez: lhNameToUnqualifiedSymbol?

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
```diff
@@ -126,18 +126,27 @@ renameRTVArgs rt = rt { rtVArgs = newArgs
     oldArgs      = rtVArgs rt
     rtArg x i    = F.suffixSymbol x (F.intSymbol "rta" i)
 
-makeRTAliases :: [Located (RTAlias F.Symbol BareType)] -> BareRTEnv -> BareRTEnv
-makeRTAliases lxts rte = graphExpand buildTypeEdges f rte lxts
+-- | Expands type aliases (recursively), by unfolding the definition of all inner
+-- aliases, and adds them to environment.
+-- We make sure that innermost aliases are unfolded and added first, or trow an
+-- error in case of cyclic dependencies.
+-- Note that, from 'makeRTEnv', the input environement contains the expanded
```
facundominguez:
```suggestion
-- Note that, from 'makeRTEnv', the input environment contains the expanded
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
```diff
@@ -147,42 +156,84 @@ graphExpand buildEdges expBody env lxts
...
+    n            = val . rtName $ a
 
-setREAlias :: RTEnv x t -> Located (RTAlias F.Symbol F.Expr) -> RTEnv x t
+-- | Inserts an expression alias in the environment. It can overwrite an alias
+-- with the same symbol.
+setREAlias :: RTEnv x t -> RTAlias F.Symbol F.Expr -> RTEnv x t
 setREAlias env a = env { exprAliases = M.insert n a (exprAliases env) }
   where
-    n            = rtName (val a)
+    n            = getLHNameSymbol . val . rtName $ a
```
facundominguez: `lhNameToUnqualifiedSymbol`?

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
```diff
@@ -147,42 +156,84 @@ graphExpand buildEdges expBody env lxts
...
 
 
 --------------------------------------------------------------------------------
-type AliasTable x t = M.HashMap F.Symbol (Located (RTAlias x t))
+type AliasTable x t = M.HashMap F.Symbol (RTAlias x t)
 
-buildAliasTable :: [Located (RTAlias x t)] -> AliasTable x t
-buildAliasTable = M.fromList . map (\rta -> (rtName (val rta), rta))
+buildAliasTable :: [RTAlias x t] -> AliasTable x t
+buildAliasTable = M.fromList . map (\rta -> (getLHNameSymbol . val . rtName $ rta, rta))
```
facundominguez: `lhNameToUnqualifiedSymbol`?

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
```diff
@@ -147,42 +156,84 @@ graphExpand buildEdges expBody env lxts
...
 
-fromAliasSymbol :: AliasTable x t -> F.Symbol -> Located (RTAlias x t)
+fromAliasSymbol :: AliasTable x t -> F.Symbol -> RTAlias x t
 fromAliasSymbol table sym
   = fromMaybe err (M.lookup sym table)
   where
     err = panic Nothing ("fromAliasSymbol: Dangling alias symbol: " ++ show sym)
 
+type AliasTable' x t = M.HashMap LHName (RTAlias x t)
```
facundominguez, line 205: Why do expression aliases use a hashmap indexed by symbols instead of LHNames?

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Measure.hs
```diff
@@ -402,7 +402,7 @@ getDefinedSymbolsInLogic env measEnv specs =
         concat (tycDCons `Mb.mapMaybe` (dataDecls spec ++ newtyDecls spec))
     getFromDataCtor decl = S.fromList $
       map lhNameToResolvedSymbol $ val (dcName decl) : (fst <$> dcFields decl)
-    getAliases spec = S.fromList $ rtName . val <$> Ms.ealiases spec
+    getAliases spec = S.fromList $ getLHNameSymbol . val . rtName <$> Ms.ealiases spec
```
facundominguez: `lhNameToUnqualifiedSymbol`?

liquidhaskell-boot/src/Language/Haskell/Liquid/Parse.hs
```diff
@@ -1308,7 +1308,10 @@ rtAliasP f bodyP
        body <- bodyP
        posE <- getSourcePos
        let (tArgs, vArgs) = partition (isSmall . headSym) args
-       return $ Loc pos posE (RTA name (f <$> tArgs) vArgs body)
+       -- The source position of the complete alias definition gets stored as its name position.
+       -- For e.g. this is reported to the user upon failure of `Bare.Expand.expandApp`.
+       -- See also `Bare.Expand.expandRTAliasApp`.
```
facundominguez: Is it more accurate to say that we do not store the source position of the complete alias definition?

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Names.hs
```diff
@@ -374,6 +379,17 @@ reflectGHCName thisModule n =
...
+
 maybeReflectedLHName :: LHName -> Maybe GHC.Name
 maybeReflectedLHName (LHNResolved (LHRLogic (LogicName _ _ m)) _) = m
 maybeReflectedLHName _ = Nothing
+
+isResolvedLogicName :: LHName -> Bool
+isResolvedLogicName (LHNResolved (LHRLogic (LogicName {})) _) = True
+isResolvedLogicName _ = False
+
+logicNameModuleOrigin :: LHName -> GHC.Module
```
facundominguez: Rename to `logicNameOriginModule`?

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Types.hs
```diff
@@ -381,11 +381,18 @@ data RTAlias x a = RTA
...
 mapRTAVars :: (a -> b) -> RTAlias a ty -> RTAlias b ty
 mapRTAVars f rt = rt { rtTArgs = f <$> rtTArgs rt }
 
-lmapEAlias :: LMap -> F.Located (RTAlias Symbol Expr)
-lmapEAlias (LMap v ys e) = F.atLoc v (RTA (F.val v) [] ys e) -- (F.loc v) (F.loc v)
+-- | Transform a logic equation to an expression alias.
+--
+-- Used when making the 'Language.Haskell.Liquid.Types.Specs.GhcSpec'
+-- where we need definitions from the logic map (e.g. Haskell inlines)
+-- for alias expansion.
```
facundominguez: Haskell inlines are not in the logic map, IIRC. `define`s are in the logic map.

tests/names/neg/Nat1.hs
```diff
@@ -1,3 +1,3 @@
-module Nat1 where
+module Nat1 () where
```
facundominguez: Is this change needed? This file could use a comment explaining that it is here only to be imported from another module.

tests/names/neg/Nat2.hs
```diff
@@ -1,4 +1,4 @@
-module Nat2 where
+module Nat2 () where
```
facundominguez: Same comment as for `Nat1.hs`.

tests/names/pos/Nat1.hs
```diff
@@ -0,0 +1,5 @@
+module Nat1 where
```
facundominguez, line 4: Add a comment mentioning that the purpose of these modules is explained elsewhere?

tests/names/pos/QualifiedTypeAliases.hs
```diff
@@ -0,0 +1,37 @@
+-- | Disambiguate between type aliases with the same name
+-- coming from distint imports using the name qualified.
+--
+-- Here, both @Nat@ modules export a type alias with
+-- the same name @INat@. The appropiate alias is used in each case.
+-- This example show that we can use both the unqualified or qualified name
```
facundominguez:
```suggestion
-- This example shows that we can use both the unqualified or qualified name
```

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -95,32 +95,40 @@ import qualified Text.Printf               as Printf
...
-        HM.fromList $ bsAliases ++ depAliases
+        unionAliasEnvs $ bsAliases : depAliases
 
 collectExprAliases
   :: BareSpecParsed
   -> TargetDependencies
   -> HS.HashSet Symbol
 collectExprAliases spec deps =
-    let bsAliases = HS.fromList $ map (rtName . val) (ealiases spec)
+    let bsAliases = HS.fromList $ map (getLHNameSymbol . val . rtName) (ealiases spec)
```
facundominguez, line 129: `lhNameToUnqualifiedSymbol`?

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -95,32 +95,40 @@ import qualified Text.Printf               as Printf
...
         depAliases =
-          [ HS.map (rtName . val) $ liftedEaliases lspec
+          [ HS.map (getLHNameSymbol . val . rtName) $ liftedEaliases lspec
```
facundominguez: `lhNameToUnqualifiedSymbol`?

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -212,24 +222,48 @@ resolveLHNames cfg thisModule localVars impMods globalRdrEnv bareSpec0 dependenc
...
+            case lookupInScopeNonReflectedEnv taliases s of
+              Right ns ->
+                case partition (\(_,lhname,_) -> isResolvedLogicName lhname) ns of
+                  -- Unresolved names correspond to type aliases defined in
+                  -- the current module. We give them priority so that the user
+                  -- can shadow those imported with the same name.
+                  (_resolved ,[(m, _, _)]) ->
+                    pure $ LHNResolved (LHRLogic $ LogicName (LH.dropModuleNames s) m Nothing) s
+                  -- Resolved type alias names come from imports.
+                  -- If we match just one we return it.
```
facundominguez:
```suggestion
                  -- If we match just one, we return it.
```

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -400,42 +445,65 @@ resolveBoundVarsInTypeAliases = updateAliases resolveBoundVars
...
+                           (map (\case
+                                   (m, LHNUnresolved _ sym, _) ->
+                                     pprint (LH.qualifySymbol (symbol . GHC.moduleNameString . GHC.moduleName $ m) sym) PJ.<+>
+                                     PJ.text "defined in current module"
+                                   (m, n@(LHNResolved _ _), _) ->
+                                     pprint (lhNameToResolvedSymbol n) PJ.<+>
+                                     PJ.text ("imported from " ++ GHC.moduleNameString (GHC.moduleName m))
+                                 ) ns)
+                         )
+                 RApp <$> pure c <*> mapM go ts <*> mapM goRef rs <*> pure r
```
facundominguez: Could we factorize some of this logic so it isn't duplicated? I'm thinking of
```Haskell
resolveTypeAlias
   :: InScopeEnv (RTAlias Symbol a)
   -> Symbol
   -> Either [Symbol] ([(GHC.Module, LHName, a)], [(GHC.Module, LHName, a)])
```
It would be a composition of `lookupInScopeNonReflectedEnv` and `partition`.

The large return type could be redefined to something isomorphic like
```Haskell
data TypeAliasResolution a
  = NoSuchTypeAlias [Symbol]
  | FoundTypeAliases
      { tarLocallyDefined :: [(GHC.Module, LHName, a)]
      , tarImported :: [(GHC.Module, LHName, a)]
      }
```
which helps navigating the meaning of the type.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -460,16 +528,24 @@ exprArg l msg = notracepp ("exprArg: " ++ msg) . go
...
+-- we 'collectTypeAliases'.
+type InScopeEnv a = SEnv [(GHC.ModuleName, (GHC.Module, LHName, a))]
+
+type InScopeNonReflectedEnv = InScopeEnv ()
+
+-- | Looks for the 'LHName's in scope with the given symbol,
+-- taking possible qualification prefixes into account.
+-- Returns a list of close but different symbols or a non-empty list
 -- with the matched names.
 lookupInScopeNonReflectedEnv
```
facundominguez: Rename to `lookupInScopeEnv`?

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -482,7 +558,7 @@ lookupInScopeNonReflectedEnv env s = do
     maybeQualifySymbol n m =
       if m == "" then n else LH.qualifySymbol m n
 
--- | Builds an environment of non-reflected names in scope from the module
+-- | Builds an environment of non-reflected names in scope from the
```
facundominguez: I think "module aliases" is more precise than "aliases" alone here. Instead of "module aliases" it could say "module imports".

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -540,46 +620,64 @@ makeLogicEnvs impAvails thisModule spec dependencies =
...
-       in fromListSEnv
-            [ (s, map (,(m, lhname)) aliases)
-              -- Note that only non-reflected names go to the InScope environment.
-              -- See the local function resolveVarName for more details.
-            | lhname@(LHNResolved (LHRLogic (LogicName s _ Nothing)) _) <- lhnames
-            ]
+    mkLogicNameEnv names =
+      LogicNameEnv
+        -- A qualified-symbol map to resolve logic names. This is called after the first
+        -- name resolution pass and before the resolution of logic names.
```
facundominguez: The second sentence strikes as rather obvious, as this is a local function called in the body of the top-level function.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -540,46 +620,64 @@ makeLogicEnvs impAvails thisModule spec dependencies =
...
+        { lneLHName = fromListSEnv [ (lhNameToResolvedSymbol n, n) | n <- names ]
+        -- A 'GHC.Name' map of all logic names comming from reflected Haskell functions.
```
facundominguez: Do these comments about the fields belong to the definition of `LogicNameEnv` in `liquidhaskell-boot/src/Language/Haskell/Liquid/Name/LogicNameEnv.hs` instead?

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -540,46 +620,64 @@ makeLogicEnvs impAvails thisModule spec dependencies =
...
+    foldl' (HM.unionWith (++)) HM.empty .
+    coerce @_ @[HM.HashMap Symbol [(GHC.ModuleName, (GHC.Module, LHName, a))]]
+
+-- | Builds a symbol-lookup environment from a list of names with the module we imported
+-- them from, adding the aliases this module has within the current module.
+mkAliasEnv:: GHC.Module -> GHC.ImportedMods -> (GHC.Module, [(LHName, a)]) -> InScopeEnv a
+mkAliasEnv thisModule impMods (m, lhnames) =
+    let aliases = moduleAliases thisModule impMods m
+     in fromListSEnv
+          [ (LH.dropModuleNames $ getLHNameSymbol lhname, map (,(m, lhname, x)) aliases)
```
facundominguez, line 669: `lhNameToUnqualifiedSymbol`?

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -589,22 +687,16 @@ makeLogicEnvs impAvails thisModule spec dependencies =
...
-      LogicNameEnv
-        { lneLHName = fromListSEnv [ (lhNameToResolvedSymbol n, n) | n <- names ]
-        , lneReflected = GHC.mkNameEnv [(rn, n) | n <- names, Just rn <- [maybeReflectedLHName n]]
-        }
-
 {- HLINT ignore collectUnhandledLiftedSpecLogicNames "Use ++" -}
 collectUnhandledLiftedSpecLogicNames :: LiftedSpec -> [LHName]
 collectUnhandledLiftedSpecLogicNames sp =
-    map (makeLocalLHName . LH.dropModuleNames . rtName . val) $ HS.toList $ liftedEaliases sp
+    map (makeLocalLHName . LH.dropModuleNames . getLHNameSymbol . val . rtName) $ HS.toList $ liftedEaliases sp
```
facundominguez: `lhNameToUnqualifiedSymbol`?

facundominguez (commented):
Quite a good piece of work! I think the standing comments won't be difficult to address.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
ninioArtillero: indeed!

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
ninioArtillero: missed that one too

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: yes

liquidhaskell-boot/src/Language/Haskell/Liquid/Parse.hs
ninioArtillero: When I removed the outer location I changed the inner (`Located LHName`) one to match it, as a safeguard in case of other used sites required the full definition location (source span). AFAICT, this seem to be the case when calling `ErrAliasApp` at `expandApp`... but is actually the only place I find the ending source position of the definition to be used. So we can go with the name location as would be expected and remove this comment.

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Names.hs
ninioArtillero: seems about right :P

tests/names/neg/Nat1.hs
ninioArtillero: It is not. It was intended as a means to make explicit that no Haskell definitions are being exported... not that there are any to export, but guess at the time it made sense.

I'll add the comment, stating it is only intended to export the type alias to be imported elsewhere.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
ninioArtillero: that is no longer the case :flushed:

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Types.hs
ninioArtillero: Messed that up, Haskell inlines and definitions from the logic map are the two cases this function is used to add new expression aliases for expansion. Fixing...

tests/names/neg/Nat2.hs
ninioArtillero: got it

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Measure.hs
ninioArtillero: Guess I was being unnecessarily cautious keeping the expression alias related things as close as possible to the previous implementation. But I see that in that in all cases the parser symbol and the unqualified symbol for the resolved name must match (as it stands currently), but by using `lhNameToUnqualifiedSymbol` we get the bonus of types guaranteeing we are dealing with a resolved name at those places.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: I think `getLHNameSymbol` is ok here, as this names are unresolved at this stage.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: yep, this one is very redundant, I'll just remove it.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: The first sentence is mentioned in the `LogicNameEnv` doc, but will take it to the field definition anyway for easy reference.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: I think not, here we expect both resolved and unresolved names.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: I like this a lot, thanks!

liquidhaskell-boot/src/Language/Haskell/Liquid/Parse.hs
```diff
@@ -1303,14 +1303,11 @@ rtAliasP :: (Symbol -> tv) -> Parser ty -> Parser (RTAlias tv ty)
...
+       posE <- getSourcePos
        args <- many aliasIdP
        reservedOp "="
        body <- bodyP
-       posE <- getSourcePos
        let (tArgs, vArgs) = partition (isSmall . headSym) args
-       -- The source position of the complete alias definition gets stored as its name position.
-       -- For e.g. this is reported to the user upon failure of `Bare.Expand.expandApp`.
-       -- See also `Bare.Expand.expandRTAliasApp`.
        return $ RTA (Loc pos posE $ makeUnresolvedLHName LHLogicNameBinder name) (f <$> tArgs) vArgs body
```
facundominguez: This could use `locBinderLogicNameP` to abbreviate.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: also factored the error logic :)

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
```diff
@@ -235,18 +251,19 @@ buildTypeEdges table = ordNub . go
...
+    go :: F.Expr -> [LHName]
     go (EApp e1 e2)   = go e1 ++ go e2
     go (ENeg e)       = go e
     go (EBin _ e1 e2) = go e1 ++ go e2
     go (EIte _ e1 e2) = go e1 ++ go e2
     go (ECst e _)     = go e
     go (ESym _)       = []
     go (ECon _)       = []
-    go (EVar v)       = go_alias v
+    go (EVar v)       = go_alias $ makeLogicLHName v  m Nothing
```
ninioArtillero: This is the reason I carried the `Ghc.Module`s all the way for the `LHName` lookup to work and get rid of the duplicate alias graph and table logic (the prime functions).

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
```diff
@@ -235,18 +251,19 @@ buildTypeEdges table = ordNub . go
     go_ref (RProp _ (RHole _)) = Nothing
     go_ref (RProp  _ t) = Just t
 
-buildExprEdges :: M.HashMap F.Symbol a -> F.Expr -> [F.Symbol]
-buildExprEdges table  = ordNub . go
+-- | Gathers all variables that match a expression alias within an expression.
```
facundominguez:
```suggestion
-- | Gathers all variables that match an expression alias within the body of
-- another expression alias.
--
-- The function is given the table of aliases, the module where the expression
-- alias is defined, and the body of the expression alias.
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
```diff
@@ -235,18 +251,19 @@ buildTypeEdges table = ordNub . go
...
+buildExprEdges :: AliasTable F.Symbol a -> Ghc.Module -> F.Expr -> [LHName]
+buildExprEdges table m  = Misc.ordNub . go
```
facundominguez: What is exactly the module `m` that should be passed when calling this function?

ninioArtillero: Here `m` is the module associated with the `LiftedSpec` where we extracted the current alias RTAlias from. It should be the module where the alias is defined, but for predicate aliases both are the same and we can use it to construct the `LHName` from the `EVar v` symbol for lookup. This is not the case when building the _edges_ for type aliases, as we get a `LHName` from the constructor, so that function remains unchanged.

facundominguez: Ok, I expanded the explanation in the suggestion to document that function.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
ninioArtillero: I went a little more verbose, but get all this across.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
ninioArtillero: Because I was focusing on type aliases alone, so to decouple both in this rather awkward way seemed appropriate at the time. In hindsight, I'd say is due to the fact that we have no `LHName`s for the lookup in the expression alias body. The way I re-merged them was to carry the `Ghc.Module` of all specs (current and dependencies) down to this point to be able to construct the `LHName`s from the body expression symbols for lookup.

facundominguez (approved):
Looks good to me. Well done @ninioArtillero!

PR #2554, opened 2025-06-19
collect type aliases unsing env

ninioArtillero:
For #2481 
I paremeterized `InScopeNonReflectedEnv` to create an environment for qualified type aliases names, this is based on #2550 . For this, I refactored `makeLogicEnvs` by moving some of its local functions to the top-level.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
@@ -106,9 +106,9 @@ collectTypeAliases
   -> TargetDependencies
   -> InScopeEnv (RTAlias Symbol ())
 collectTypeAliases impMods thisModule spec deps =
-    let bsAliases = mkAliasEnvUnfiltered thisModule impMods (thisModule, bsNames)
+    let bsAliases = mkAliasEnv thisModule impMods (thisModule, bsNames)
         bsNames = [ (val . rtName $ rta, void rta) | rta <- map val (aliases spec)]
-        depAliases = map (mkAliasEnvUnfiltered thisModule impMods) $
+        depAliases = map (mkAliasEnv thisModule impMods) $
           [ (m, depNames)
           | (sm, lspec) <- HM.toList (getDependencies deps)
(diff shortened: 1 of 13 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -540,46 +566,71 @@ makeLogicEnvs impAvails thisModule spec dependencies =
...
+moduleAliases thisModule deps impMods m =
+    case Map.lookup m impMods of
+      -- Aliases for imported modules
+      Just impBys -> concatMap imvAliases $ GHC.importedByUser impBys
+      Nothing
+        | thisModule == m ->
+          -- Aliases for the current module
+          [GHC.moduleName m, GHC.mkModuleName ""]
+          -- Dependencies not in scope get an empty alias
+        | HM.member (GHC.toStableModule m) (getDependencies deps) -> [GHC.mkModuleName ""]
```
facundominguez, line 624: If I understand correctly, the type aliases should be in scope in the modules that are imported directly. Once imported, type aliases would be accumulated transitively. If T exports a type alias that has been imported from a transitive dependency, the following should work:
```
import qualified T as Q

{-@ f :: Q.TypeAlias @-}
f :: Int
f = 0
```
Using a empty module alias gives a behavior that the user would not expect.

ninioArtillero: That behavior holds AFAIK (its the case with the use of the `PtrV` alias in the `Foreign.Concurrent_LHAssumptions` module) but it seems the original implementation depends on bringing along—in the `TargetDependencies` build at `GHC.Plugin.loadDependencies`—some transitive dependencies to query when collecting the aliases (and other purposes which I'm looking into). 

This change makes this explicit—to avoid those names from mapping over an empty list and getting discarded—, so its just reifying the current behavior in a way. In other words, it would seem that having an empty module alias for this transitive dependencies reflects the fact that type aliases accumulate transitively, which indeed is a behavior a user might not expect.

ninioArtillero:
this work was merged into #2550

PR #2561, opened 2025-07-25
Fix rendering of sublist in type alias docs

ninioArtillero:


facundominguez:
Thanks @ninioArtillero!

PR #2563, opened 2025-07-28
Remove support for multiple cores

facundominguez:
See #2562 for the details.


facundominguez:
Closing in favor of only changing the default (#2564).

PR #2566, opened 2025-08-06
Qualified predicate aliases support

### LINKED ISSUE #2481: Allow to import type and predicate aliases qualified

ninioArtillero:
This PR include some small refactorings and implements qualified predicate alias support
with comprehensive tests in place.
Together with #2550, it fixes #2481 and accomplishes the goals set in my [GSoC project](https://summerofcode.withgoogle.com/programs/2025/projects/ALSN6wwr).

A key insight was that predicate aliases get mixed with define and inline annotations when building the lifted spec
as a consequence of the alias expansion implementation.
Predicate alias are separated and placed in the _logic environments_ used for name resolution.
The resulting specification gives precedence, when creating the lifted spec, to those locally defined
and next to those from the lexicographical last dependency module name
in accordance to the handling of other annotations.

Handling inline annotations in a similar manner is left for future work.


liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -247,7 +247,8 @@ resolveLHNames cfg thisModule localVars impMods globalRdrEnv bareSpec0 dependenc
         n@(LHNUnresolved LHLogicName _) ->
           -- This one will be resolved by resolveLogicNames
           pure n
-        LHNUnresolved ns s -> lookupGRELHName [] ns lname s listToMaybe
+        LHNUnresolved ns@(LHDataConName _) s -> lookupGRELHName [] ns lname s listToMaybe
+        -- Resolved names.
         n -> pure n
```
facundominguez: In this case, the `n` case should be changed to be `n@LHNResolved{} -> pure n`. Otherwise, we get no warning here when new data constructors are added to `LHName` or name spaces.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -124,15 +124,22 @@ collectTypeAliases impMods thisModule spec deps =
...
+          , let depPredAliases = HS.filter isResolvedLogicName exprAliases
           ]
-     in
-        HS.unions $ bsAliases : depAliases
+        inlinesAndDefines = HS.unions
+          [ HS.map lhNameToResolvedSymbol depInlinesAndDefines
+          | (_, lspec) <- HM.toList (getDependencies deps)
+          , let exprAliases = HS.map (val . rtName) $ liftedEaliases lspec
+          , let depInlinesAndDefines = HS.filter isGeneratedLogicName exprAliases
+          ]
```
facundominguez: This is discarding names that are not generated or logic names. This would deserved to be explained in a note, and the note referenced at the place where GeneratedNames are defined, in the definition of ealiases, and perhaps where GeneratedNames are created.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -715,7 +722,7 @@ resolveLogicNames
   -> LocalVars
   -> LogicNameEnv
   -> HS.HashSet LocSymbol
-  -> HS.HashSet Symbol
+  -> (HS.HashSet Symbol, HS.HashSet Symbol)
   -> BareSpecParsed
   -> State RenameOutput BareSpecLHName
 resolveLogicNames cfg env globalRdrEnv lmap0 localVars lnameEnv privateReflectNames allEaliases sp = do
```
facundominguez: Replace `allEaliases` with a pattern `(predicateAliases, inlinesAndDefines)` in the parameter list?

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
```diff
@@ -65,25 +67,32 @@ import qualified Text.Printf                           as Printf
...
+        -- lmap expansion happens during elaboration
+        -- this clearly breaks things if a signature
+        -- contains lmap functions but never gets
+        -- elaborated
+        else [ e | (_, xl) <- M.toList (lmSymDefs lmap), let e = lmapEAlias xl ]
+      }
     specs = (modName, mySpec) : dependencySpecs
+    -- Here, 'Symbol's are converted to 'LHName's in expressions in an unambiguous way,
+    -- allowing to use the type alias lookup and expansion procedure for predicate aliases.
+    specs' = map (toBareSpecLHName cfg lenv . snd) $ (modName, mySpec') : dependencySpecs
```
facundominguez: Instead of converting the whole spec, we could be converting only the field ealiases, it seems. Could be left in a TODO note.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
facundominguez: It is rewritten in a later commit.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
facundominguez: Perhaps it would be less surprising to throw in here anything not considered to be a predicate alias.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: Got it!

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: Added the note and reference it in: the corresponding `BareSpec` and `LiftedSpec` fields, the function that make such names `GeneratedLogicName`, the two functions where this happen (for inlines and defines respectively), and the place in name resolution where this all converges.

I did not add it where the `GeneratedLogicName` constructor is defined, as generated names seem to be used for other purposes.

I saw a couple of different conventions for the note format, I can gladly changing it if you think we should be moving to a specific format for this kind of notes.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Expand.hs
ninioArtillero: got a implementation converting the single field that seems to be working :)

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Types.hs
```diff
@@ -375,7 +375,6 @@ data RTAlias x a = RTA
   , rtTArgs :: [x]                -- ^ type parameters
   , rtVArgs :: [Symbol]           -- ^ value parameters
   , rtBody  :: a                  -- ^ what the alias expands to
-  -- , rtMod   :: !ModName           -- ^ module where alias was defined
```
facundominguez, line 378: Such a big PR only to get rid of this commented line :stuck_out_tongue:

facundominguez (approved):
LGTM! Bravo @ninioArtillero :bouquet:

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Types.hs
ninioArtillero: The devil is in the details indeed!

PR #2567, opened 2025-08-07
CircleCI fix and document a bit more the benchmarking configuration

facundominguez:


.circleci/config.yml
```diff
@@ -72,7 +72,7 @@ commands:
           name: Test
           command: |
             mkdir -p /tmp/junit/cabal
-            LIQUID_CABAL_PROJECT_FILE=cabal.project scripts/test/test_plugin.sh
+            scripts/test/test_plugin.sh -- --ghc-options=-j1 -j2
```
facundominguez:
```suggestion
            scripts/test/test_plugin.sh -- --ghc-options=-j1 -j1
```

PR #2568, opened 2025-08-14
liquidhaskell-boot.cabal: relax upper bounds on Diff and optparse-applicative

ulysses4ever:
I checked that it builds with the latest versions.

I'd appreciate it if you could revise on hackage both 0.9.12.2 and 0.9.10.1.2.


facundominguez:
Thanks @ulysses4ever! I created the requested revisions on hackage.

PR #2570, opened 2025-08-25
Improve duplicate names error

ninioArtillero:
- **add kind for context to ErrDupNames**
- **include type alias definition location in message**

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
@@ -546,8 +546,15 @@ errResolveTypeAlias :: LocSymbol -> TypeAliasResolution (RTAlias x a) -> Error
 errResolveTypeAlias ls (FoundTypeAliases imported local) =
   ErrDupNames (LH.fSrcSpan ls) "type alias" (pprint $ val ls)
     (
-      -- Collected local type alias names are unresolved,
-      -- so we need to extract their symbol with a function that tolerates them.
+      -- Currently, multiple local definitions prevent this error from being raised,
+      -- because duplicate names are discarded when constructing the alias environment
+      -- for each individual module,
+      -- and a local alias always shadows any imported one. Such duplicates are detected
+      -- later during validation of the final target spec.
+      --
+      -- Also, note that collected local type alias names remain unresolved at this stage,
+      -- so we must extract their symbol using a function that can safely handle unresolved
+      -- names.
       map (\(_, lhn, rta) -> pprint (getLHNameSymbol lhn)
                              PJ.<+>
                              PJ.text "defined in current module at"
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
@@ -681,7 +681,10 @@ mkAliasEnv:: GHC.Module -> GHC.ImportedMods -> (GHC.Module, [(LHName, a)]) -> In
 mkAliasEnv thisModule impMods (m, lhnames) =
     let aliases = moduleAliases thisModule impMods m
      in fromListSEnv
-          [ (LH.dropModuleNames $ getLHNameSymbol lhname, map (,(m, lhname, x)) aliases)
+          -- Note that when building a name environment for the current module
+          -- we might process unresolved names here.
+          [ (LH.dropModuleNames $ getLHNameSymbol lhname
+            , map (,(m, lhname, x)) aliases)
           | (lhname, x) <- lhnames
           ]
 
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
@@ -61,6 +61,7 @@ import           Language.Haskell.Liquid.Types.Names
 import           Language.Haskell.Liquid.Types.RType
 import           Language.Haskell.Liquid.Types.RTypeOp
 
+import           Control.Monad.Except (ExceptT, runExceptT, throwError)
 import           Control.Monad ((<=<), mplus, unless, void)
 import           Control.Monad.Identity
 import           Control.Monad.State.Strict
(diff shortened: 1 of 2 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -540,20 +542,23 @@ data TypeAliasResolution a
...
 
-errResolveTypeAlias :: LocSymbol -> TypeAliasResolution b -> Error
+errResolveTypeAlias :: LocSymbol -> TypeAliasResolution (RTAlias x a) -> Error
 errResolveTypeAlias ls (FoundTypeAliases imported local) =
-  ErrDupNames (LH.fSrcSpan ls) (pprint $ val ls)
+  ErrDupNames (LH.fSrcSpan ls) "type alias" (pprint $ val ls)
     (
-      map (\(m, lhn, _) -> pprint (LH.qualifySymbol (symbol . GHC.moduleNameString . GHC.moduleName $ m)
-                                   $ getLHNameSymbol lhn)
+      map (\(_, lhn, rta) -> pprint (getLHNameSymbol lhn)
```
facundominguez, line 562: Use `lhNameToUnqualifiedSymbol` instead of `getLHNameSymbol`?

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -540,20 +542,23 @@ data TypeAliasResolution a
...
                            PJ.<+>
-                           PJ.text "defined in current module")
-      local
+                           PJ.text "defined in current module at"
+                           PJ.<+>
+                           pprint  (LH.fSrcSpan . rtName $ rta))
+          local
      ++
-     map (\(m, lhn, _) -> pprint (lhNameToResolvedSymbol lhn)
+     map (\(_, lhn, rta) -> pprint (lhNameToResolvedSymbol lhn)
```
facundominguez: We come to the point where we need to set what the qualifier should be. If `m` is the module alias, it could make sense to qualify it with it, instead of using the definition module. The definition module is already implied by the span printed below, I guess.

facundominguez (commented):
Good PR. I'm leaving a couple of comments.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: Tests pass with this change, making me suspect this variant of the error is unreachable (as we expect those names to be unresolved and needing `getLHNameSymbol`). As far as I've checked, this error triggers in the case that multiple imported definitions are found for an occurrence. The case of a single local definition with multiple imported ones is ruled out by shadowing. The case of multiple local definitions (and possibly other imported ones) triggers the error at line 169 in `Bare.Check.checkTargetSpec` instead of this one.

I guess we can either tolerate the redundancy (as a kind of safeguard) or remove it for consistency.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: That `m` is the module and it is the lookup that uses the module aliases internally. If we made aliases accessible there, we would have to pick one to qualify with in case there are multiple module aliases.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
facundominguez: Hm, is it better to use have `checkTargetSpec` to check duplicates instead of doing it in this module?

I guess it might be a different kind of check. Here the resolution environment must have only one of the local definitions.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
facundominguez: Now I'm inclining for keeping both the definition location and the module that is directly imported in the error message. 🤔

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: These two checks serve different purposes. As I understand, one reports all matching names in scope, addressing potential resolution ambiguity, while the other ensures spec consistency by detecting duplicate alias definitions. However, they overlap in the specific case I mentioned, where the latter error is the one that emerges as a consequence of laziness.

My current take is to keep both in place, since future changes could affect how these checks interact, potentially causing some errors to go undetected.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: So we might open with the unqualified name (to avoid giving the impression that the fully qualified name can be used when is not the case), followed by the module of origin and the definition location. Something like:

```haskell
```Haskell
examples/C.hs:6:19: error:
    Ambiguous specification symbol `Bar` for type alias
    Could refer to any of the names
    .
    * Bar from module B defined at examples/B.hs:3:10-39
    .
    * Bar from module B' defined at examples/B'.hs:3:10-39
  |
6 | {-@ baz :: Foo -> Bar @-}
  |                   ^^^
```

facundominguez: Sounds good to me.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -456,19 +458,20 @@ fixExpressionArgsOfTypeAliases
...
                  RApp <$> pure c <*> fixExprArgs (btc_tc c) rta (mapM go ts) <*> mapM goRef rs <*> pure r
-               -- Report ambiguos name and continue traversing.
-               _ -> do
-                 addError $ errResolveTypeAlias (s <$ lname) tar
-                 RApp <$> pure c <*> mapM go ts <*> mapM goRef rs <*> pure r
+               -- We don't report ambiguity here and just continue traversing;
+               -- the corresponding error should be thrown when resolving type
+               -- type constructor names. Otherwise it would appear duplicate
+               -- on the first phase error batch.
+               _ -> RApp <$> pure c <*> mapM go ts <*> mapM goRef rs <*> pure r
```
facundominguez: Another fix that comes to mind is stopping name resolution early when there are errors here.

If there are arguments that haven't been "fixed", then error resolution might report errors about missing aliases that it has incorrectly interpreted as type aliases.

ninioArtillero: Done, along with a refactor of the error handling in `resolveLHNames`, placing an early exit function at the appropriate places.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
ninioArtillero: > Here the resolution environment must have only one of the local definitions.

That's indeed the case, and the explanation of why this error path is death. I have documented it in place, until we see how to keep those duplicate names (and if that change is worth the effort), so we can abort earlier.

ninioArtillero:
I think this PR is ready. Force-pushed it to refine the error message to resemble a GHC error presented by @amesgen in a blog post review.

PR #2571, opened 2025-09-03
Add finite field support

clayrat:
Companion PR for https://github.com/ucsd-progsys/liquid-fixpoint/pull/755


liquid-prelude/src/Language/Haskell/Liquid/FinField.hs
```diff
@@ -0,0 +1,44 @@
...
+{-@ define add    x           y                      = (FF_add x y) @-}
+add :: FF17 -> FF17 -> FF17
+add x y = FF17 (FFld (ffToInteger (toFFld x) + ffToInteger (toFFld y) `mod` 17))
+
+{-@ assume mul :: x : FF17 -> y : FF17 -> {v:FF17 | v = FF_mul x y} @-}
+{-@ define mul    x           y                      = (FF_mul x y) @-}
+mul :: FF17 -> FF17 -> FF17
+mul x y = FF17 (FFld (ffToInteger (toFFld x) * ffToInteger (toFFld y) `mod` 17))
+
+-- TODO move to tests
```
nikivazou: Indeed!

.circleci/config.yml
```diff
@@ -27,7 +27,7 @@ commands:
       - checkout
       - add_ssh_keys
       - run: git submodule sync
-      - run: git submodule update --init
+      - run: git submodule update --init --remote
```
nikivazou: Is this required?

clayrat: This was necessary for CI to run with the underlying LF PR. Now that it's been merged, I'll roll back both this change and the one in .gitmodules. That's why I put a "please do not merge" disclaimer on every LF PR where changes to LF are required.

facundominguez:
:wave: I think that other changes in LF will require this PR to rebase ~master~`develop` in order to get CI to succeed.

tests/tests.cabal
```diff
@@ -1466,8 +1466,10 @@ executable unit-pos-2
                     , Extype
                     , Fail
                     , FailName
-                    , FF17
-                    , FF2131
+                  -- TODO the following tests are disabled because they currently can only be run with CVC5
+                  -- and we use Z3 by default (specifically in CI)
+                  --  , FF17
+                  --  , FF2131
```
facundominguez: You could use `{-@ LIQUID "--smtsolver=cvc5" @-}` in those files.

That's what `tests/benchmarks/sf/Lists.hs` does, at least.

clayrat: Oh neat, thanks!

clayrat:
Should be good now. Do we need performance charts for this? I don't think the changes are too extensive in the end.

facundominguez (approved):
LGTM!

facundominguez:
I don't think we need to check performance.

PR #2583, opened 2025-10-30
Fix name resolution of tuple expressions in type alias applications

AlecsFerra:
This PR fixes name resolution so the following example stops failing.

```haskell
module Minimal where

{-@ type Ix typ E = {v:typ | prop v = E} @-}

data Foo where
  {-@ MkFoo :: x:Int -> Ix Foo (x, x) @-}
  MkFoo :: Int -> Foo
```

This was the error produced:
```
[1 of 1] Compiling Minimal          ( Minimal.hs, Minimal.o )
Fixpoint.Types.dummyLoc:1:1: error:
    Unknown logic name `Tuple2`
    Cannot resolve name
    Maybe you meant one of: len autolen
```

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
@@ -513,7 +513,13 @@ mapMBareTypes f  = go
 --   the string `Prop (Ev (plus n n))` where `Prop` is the alias:
 --     {-@ type Prop E = {v:_ | prop v = E} @-}
 --   the parser will chomp in `Ev (plus n n)` as a `BareType` and so
---   `exprArg` converts that `BareType` into an `Expr`.
+-- | @exprArg@ converts a type to a value.
+--   
+--   At parse time the arguments of type aliases are all treated as types.
+--   This needs fixing before verification because some arguments are
+--   meant to be values. Hence, this function to correct the
+--   arguments in question. See the documentation of
+--   @fixExpressionArgsOfTypeAliases@ for some more context.
 exprArg :: SourcePos -> String -> BareTypeParsed -> ExprV LocSymbol
 exprArg l msg = notracepp ("exprArg: " ++ msg) . go
   where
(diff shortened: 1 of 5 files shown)


facundominguez:
:wave: My guess is that `exprArg` [[link]](https://github.com/ucsd-progsys/liquidhaskell/blob/e35312e93aa52fa1c7b77b366a715684c8344d45/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs#L508-L527) needs to handle tuples specially, and perhaps lists as well.

AlecsFerra:
Oh I did not see that code, thanks!

AlecsFerra:
Hi @facundominguez lmk if it makes sense, I did not do lists as the parser accepts only `[x]` as a list literal, I guess that was only meant to parse list types

AlecsFerra:
Now that I think about it, maybe tuples weren't meant to be parsed as expressions 🤔 and we are abusing the fact that they have the same syntax as their type

nikivazou (commented):
Great!!!!! 

Can you add some more things in the test file to test that 
- pair type parameters still work, e.g., `Ix (Int,Int) True`
- pairs in refinements work (not sure if this was already supported, e.g., `{v:() | (1,2) /= (3,4)}`

AlecsFerra:
Added the new tests, curious that we need PLE to see that the two tuples are different!🤔

nikivazou (approved):

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
facundominguez:
```suggestion
-- | @exprArg@ converts a type to a value.
--   
--   At parse time the arguments of type aliases are all treated as types.
--   This needs fixing before verification because some arguments are
--   meant to be values. Hence, this function to correct the
--   arguments in question. See the documentation of
--   @fixExpressionArgsOfTypeAliases@ for some more context.
```

facundominguez, line 463:
```suggestion
-- | @fixExpressionArgsOfTypeAliases taliases spec@ converts types to
-- values when they appear in value positions of type aliases according
-- to @taliases@.
--
-- The expression arguments of type aliases are initially parsed as
-- types. This function converts them to expressions.
--
-- For instance, in @Prop (Ev (plus n n))@ where `Prop` is the alias
--
-- > {-@ type Prop E = {v:_ | prop v = E} @-}
--
-- the parser builds a type for @Ev (plus n n)@, making a type
-- constructor of @Ev@ and type variables of @plus@ and @n@. But
-- @Ev@ is really a data constructor, @plus@ is a function, and @n@
-- is a value. 
```

facundominguez (approved):
LGTM! Thanks @AlecsFerra!

I contributed some amendments to the preexisting documentation.

PR #2586, opened 2025-11-03
Add new syntax for indexed types

### LINKED ISSUE #2585: Deprecating the type alias `Prop`
## Motivation
Currently the main way to specify data propositions in Liquid Haskell
is the following:
``` haskell
data Foo where
  {-@ MkFoo :: Prop (Foo 1) @-}
  MkFoo :: Foo
data FOO = Foo Int
```
Using `Prop` in `Language.Haskell.Liquid.ProofCombinators` defined as
``` haskell
{-@ type Prop A = {v:_ | prop v = A} @-}

```
While technically we don't need the `data FOO` declaration,
and we could write:

``` haskell
data Foo where
  {-@ MkFoo :: Prop 1 @-}
  MkFoo :: Foo
```
We introduce a new datatype `FOO` with a constructor with the same name
as the type, to make explicit the type that is being used in place of `_`
in the `Prop` definition.

This is somewhat verbose and confusing, especially for new users,
first is the need to introduce a new datatype just to use `Prop`,
second it leverages the fact that type constructors and data
constructors live in different namespaces, which can be confusing, and
third, since `_` it can be replaced by any type and we could
accidentally refine a different type than the one we intended to use.

## Proposal
We should deprecate the use of `Prop` and introduce a new type alias
which makes explicit the type being used to replace `_` in `Prop`.
``` haskell
{-@ type Ix t A = {v:t | prop v = A} @-}

```
Now we can rewrite the previous example as:
``` haskell
data Foo where
  {-@ MkFoo :: Ix Foo 1 @-}
  MkFoo :: Foo
```
Having the same readability as before, but without all the ceremony of
introducing a new datatype and the awkwardness of using `_`.

I'm not super attached to the name `Ix`, so if there is a better
suggestion for the name I'm open to it.

## Possible drawbacks
I can think of two possible drawbacks:
- We don't have the equivalent to the `DEPRECATED` pragma so there is
  no way to signal to users that is reading some old code that they
  should migrate to the new `Ix` alias.
- We would need to update all the documentation and examples that
  currently use `Prop` to use `Ix` instead.

AlecsFerra:
Should close #2585

### DIFF
--- a/docs/mkDocs/docs/specifications.md
+++ b/docs/mkDocs/docs/specifications.md
@@ -1157,7 +1157,7 @@ types**.
 
 ### Definition
 
-A **stratified type** is a data type refined with `Prop` from
+A **stratified type** is a data type indexed with `Ix` from
 `Language.Haskell.Liquid.ProofCombinators`, where all recursive
 occurrences of the type:
 - appear under `Prop`, and
@@ -1177,12 +1177,11 @@ data Ty = TInt | TFun Ty Ty
 
 {-@ stratified Val @-}
 data Val where
-  {-@ VInt :: Int -> Prop (Val TInt) @-}
+  {-@ VInt :: Int -> Ix Val TInt @-}
   VInt :: Int -> Val
 
   {-@ VFun :: t1:Ty -> t2:Ty
-           -> (Prop (Val t1) -> Prop (Val t2))
-           -> Prop (Val (TFun t1 t2)) @-}
+           -> (Ix Val t1 -> Ix Val t2)
+           -> Ix Val (TFun t1 t2) @-}
   VFun :: Ty -> Ty -> (Val -> Val) -> Val
-data VAL = Val Ty
 ```
--- a/docs/mkDocs/docs/specifications.md
+++ b/docs/mkDocs/docs/specifications.md
@@ -65,7 +65,7 @@ above the data definition. See, for example, [tests/pos/Map.hs](https://github.c
 
 ```haskell
 {-@
-data Map k a <l :: k -> k -> Prop, r :: k -> k -> Prop>
+data Map k a <l :: k -> k -> Bool, r :: k -> k -> Bool>
   = Tip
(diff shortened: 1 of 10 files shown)


facundominguez:
I rebased `develop`.

facundominguez (commented):
LGTM. There are plenty of tests that use Prop. Should those be adjusted? Otherwise, it is a lot of testing in the deprecated way.

AlecsFerra:
Idk, I think it’s fine to keep them as they are. The important part is that Prop and Ix share the same refinement, so it doesn’t really matter which one the tests use. I just ported the SKI test since it uses all the features, mainly to double-check that nothing sketchy was happening.

docs/mkDocs/docs/specifications.md
```diff
@@ -1157,7 +1157,7 @@ types**.
 
 ### Definition
 
-A **stratified type** is a data type refined with `Prop` from
+A **stratified type** is a data type indexed with `Ix` from
 `Language.Haskell.Liquid.ProofCombinators`, where all recursive
 occurrences of the type:
 - appear under `Prop`, and
```
facundominguez: There are still some mentions of Prop in the documentation.
```
docs/mkDocs/docs/options.md:     VV : {VV : Bool | Prop VV <=> Set_mem ?b (listElts ?a)}
docs/mkDocs/docs/specifications.md:data Map k a <l :: k -> k -> Prop, r :: k -> k -> Prop>
docs/mkDocs/docs/specifications.md:- appear under `Prop`, and
```

It is looking like those should be removed, but only the last one is related to this PR.

AlecsFerra: The last one is correct I missed it, the others I think they refer to a different Prop but honestly I have no idea what it is

AlecsFerra: I did some reverse engineering and now they should be fixed

facundominguez (approved):

PR #2588, opened 2025-11-05
Update LF after removing PGrad

facundominguez:
Follow up to https://github.com/ucsd-progsys/liquid-fixpoint/pull/789


AlecsFerra:
I think LH exposes some flags related to gradual stuff so maybe there is more code that depend on those that can be deleted

facundominguez:
Indeed. I'm doing a second pass to remove unused code.

PR #2591, opened 2025-11-08
Matching PR for LF #793

ranjitjhala:
Which adds support for some string operations, and switches `stringTheory` on by default (but still not on by default in LH) https://github.com/ucsd-progsys/liquid-fixpoint/pull/793


ranjitjhala:
I didn't see any major performance difference even using `stringTheory` on as default, but still modified LH to preserve current behavior.

PR #2594, opened 2025-11-24
Update LF after requiring braces for qualifiers

facundominguez:
Follow up to https://github.com/ucsd-progsys/liquid-fixpoint/pull/805


facundominguez:
The blogposts that discuss qualifiers are going to refer to the old syntax. One way around that could be to add the blog post examples to the test suite and link them from each blog post so the reader can get a version that works by following the link. I'm not aiming to do that in this PR, though.

PR #2596, opened 2025-11-29
Show solutions of non-cut kvars in error messages

facundominguez:
Follow up to https://github.com/ucsd-progsys/liquid-fixpoint/pull/810

This PR shows kvar solutions in error messages. As an example, where an error message used to show:
```
KVars.hs:10:1: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : GHC.Types.Int
    .
      VV : {VV##579 : GHC.Types.Int | VV##579 > 0}
    .
    Constraint id 2
   |
10 | test x = intId x
   | ^^^^^^^^^^^^^^^^
```
With this PR it says
```
KVars.hs:10:1: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : {v : GHC.Types.Int | v == x##aDm
                                && v < 0}
    .
      VV : {VV##579 : GHC.Types.Int | VV##579 > 0}
    .
    in the context
      x##aDm : {v : GHC.Types.Int | v < 0}
    Constraint id 2
   |
10 | test x = intId x
   | ^^^^^^^^^^^^^^^^
```
See the test `tests/errors/KVars.hs` added in this PR.

Modules that produce large kvar solutions are a bit slower to verify now. This is noticeable in only 5 of our benchmarks. This is because simplifying the solutions (as in `simplifyKVars`) adds a few seconds. The simplification is forced when writing `.vim.annot`, `.json`, and `.html` files in the `.liquid` folder (currently controlled by the `--no-annotations` flag). I'm guessing that these files should be stopped from being written by default.

The simplification of a kvar solution is also forced when rendering an error message where the kvar solution is needed. If no such error is shown, there is no performance overhead. 

Here are the worsening benchmarks:

* `tests/neg/StateConstraints0.hs` 7s +8s
* `tests/benchmarks/icfp15/pos/IfM.hs` 2s +5s
* `tests/benchmarks/esop2013-submission/Base.hs` 30s + 3s
* `tests/benchmarks/bytestring-0.9.2.1/Data/ByteString/LazyZip.hs` 2s + 3s
* `tests/pos/StateConstraints.hs` 2s +2s

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Liquid.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Liquid.hs
@@ -135,29 +135,29 @@ solveCs cfg tgt cgi info names = do
   finfo            <- cgInfoFInfo info cgi
   let fcfg          = fixConfig tgt cfg
   F.Result {resStatus=r0, resSolution=solCuts, resNonCutsSolution=solNonCuts} <- solve fcfg finfo
-  let sol           = HashMap.union solCuts solNonCuts
+  let sol           = HashMap.union (HashMap.map F.Delayed solCuts) solNonCuts
   let failBs        = gsFail $ gsTerm $ giSpec info
   let (r,rf)        = splitFails (S.map val failBs) r0 
   let resErr        = second (applySolution finfo sol . cinfoError) <$> r
   -- resModel_        <- fmap (e2u cfg sol) <$> getModels info cfg resErr
-  let resModel_     = cidE2u cfg sol <$> resErr
-  let resModel'     = resModel_  `addErrors` (e2u cfg sol <$> logErrors cgi)
+  let resModel_     = cidE2u cfg <$> resErr
+  let resModel'     = resModel_  `addErrors` (e2u cfg <$> logErrors cgi)
                                  `addErrors` makeFailErrors (S.toList failBs) rf 
                                  `addErrors` makeFailUseErrors (S.toList failBs) (giCbs $ giSrc info)
   let lErrors       = applySolution finfo sol <$> logErrors cgi
-  let resModel      = resModel' `addErrors` (e2u cfg sol <$> lErrors)
+  let resModel      = resModel' `addErrors` (e2u cfg <$> lErrors)
   let out0          = mkOutput cfg resModel finfo sol (annotMap cgi)
   return            $ out0 { o_vars    = names    }
                            { o_result  = resModel }
 
 
-e2u :: Config -> F.FixSolution -> Error -> UserError
-e2u cfg s = fmap F.pprint . tidyError cfg s
+e2u :: Config -> Error -> UserError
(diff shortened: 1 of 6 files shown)


facundominguez:
One other aspect to consider: the user may find now in error messages kvar solutions that don't simplify well and are very large.

Maybe we could limit the size of predicates that we show in error messages, and offer a flag to show them in full when such large predicates show up.

Another alternative is to leave it as is, and have the user assume signatures or otherwise disable verification to get rid of large error messages. This requires more familiarity with LH though.

----

And one other aspect: the user might now see predicates that contain existential quantification in error messages. These are still disabled in Liquid Haskell annotations. Let me know if you want to do something about it.

ranjitjhala:
Awesome! I wonder if we should put this behind a flag, eg there is already some short-error or similar?

Not clear what the right default should be though: the performance hit is small so acceptable imo; plus we can hide the serialization behind a flag if needed.

The question is: are the non-cuts helpful as default OR should they be opt in? Any thoughts?

facundominguez:
I expect them to be helpful in general. The original message with no predicate in the LHS does not allow to determine what facts are we missing to prove the RHS.

As another example, the original error message of https://github.com/ucsd-progsys/liquid-fixpoint/issues/780 was
```
Test.hs:110:1: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : GHC.Types.Int
    .
      VV : {VV##2013 : GHC.Types.Int | ?a > 0}
    .
    in the context
      ?a : GHC.Types.Int
    Constraint id 4
    |
110 | stFun i = do
    | ^^^^^^^^^^^^...
```
with this PR, it becomes
```
Test.hs:106:1: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : {v : GHC.Types.Int | Test.gt0 i##aS5
                                && ?d == i##aS5
                                && ?d == v
                                && v == i##aS5}
    .
      VV : {VV##1940 : GHC.Types.Int | ?d > 0}
    .
    in the context
      i##aS5 : GHC.Types.Int
       
      ?d : GHC.Types.Int
       
      ?c : (Test.State GHC.Types.Int GHC.Types.Int)
       
      ?a : (GHC.Internal.Base.Monad (Test.State GHC.Types.Int))
       
      ?b : (Test.State GHC.Types.Int ())
    Constraint id 4
    |
106 | stFun i = do
    | ^^^^^^^^^^^^...
```
which allows to appreciate the bug at that ticket: `gt0 i##aS5`, if unfolded, should demonstrate the RHS, but it is not unfolded (because it is in the solution of a kvar).

ranjitjhala:
Great -- lets just go ahead and merge then? We can cross the "too long error messages" bridge in one of two ways: a `--short-errors` flag (maybe even reusing the existing `shortNames` or `shortErrors` flags?) or your threshold idea?

facundominguez:
ok, merging then \o/

ranjitjhala:
Oh I missed this, this is great thanks!!!!

PR #2602, opened 2025-12-30
feat: add pkgconfig-depends z3

quantum9Innovation:
this fixes compatibility issues with Nix and allows discovery of the `z3` dependency needed for Liquid Haskell to work
originally patched by @youwen5

### DIFF
--- a/liquidhaskell.cabal
+++ b/liquidhaskell.cabal
@@ -85,6 +85,7 @@ library
                       ghc-prim
   default-language:   Haskell98
   ghc-options:        -Wall
+  pkgconfig-depends:  z3
 
   if flag(devel)
     ghc-options:      -Werror


quantum9Innovation:
For anyone interested, [this comment](https://github.com/input-output-hk/haskell.nix/issues/2433#issuecomment-3695362117) has the instructions for using `liquidhaskell` with Haskell.nix. Merging this PR would make step 1 unnecessary, since the main `liquidhaskell` package will have `z3` listed as a dependency.

youwen5:
Chiming in here, the way I fixed this by adding the `pkgconfig-depends` field is a complete hack and entirely invalid, pkg-config is a tool for finding _libraries to link to_, not binaries needed at build time. It just so happened that `haskell.nix` would pull in `z3` to provide it as a library, but due to the particulars of Nix it would also make Z3 available in $PATH.

The more idiomatic solution is the [build-tools](https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-field-build-tools) field, but that field has in fact been deprecated in Cabal 2.0 and removed in Cabal 3.0. As the docs mention, this is a field that may be used to specify dependencies required at compile time, although Cabal will not attempt to provide them if they're arbitrary binaries in $PATH (e.g. Z3 and other SMT solvers). This will however be picked up by `haskell.nix` which will provide Z3 to `liquidhaskell` at compile time, and ignored by Cabal (albeit with a deprecation notice) when invoked so everything works out. But this still doesn't seem to work as well as the `pkgconfig-depends` hack.

`build-tools` was deprecated in favor of [build-tool-depends](https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-field-build-depends), but it's not clear to me how (or even if) this field can be used to specify an arbitrary binary _not_ provided as the output of a Haskell package. In any case, no amount of my fiddling with this field managed to get `haskell.nix` to provide Z3 at build time, so this method has been fruitless thus far.

So I'm not sure it's appropriate to merge such changes into upstream, as the `pkgconfig-depends` approach is a hack, albeit one that works quite well, and `build-tools` is deprecated and unwieldy.

liquidhaskell.cabal
```diff
@@ -85,6 +85,7 @@ library
                       ghc-prim
   default-language:   Haskell98
   ghc-options:        -Wall
+  pkgconfig-depends:  z3
```
facundominguez, line 88:
```suggestion
  -- ignored by cabal but honored by haskell.nix
  -- See https://github.com/ucsd-progsys/liquidhaskell/pull/2602#issuecomment-3699178483
  pkgconfig-depends:  z3
```

facundominguez:
Hello! Glad to see Liquid Haskell building with haskell.nix.

I would prefer to keep this outside the repo, if possible. Being specific to haskell.nix, it is going to be difficult for maintainers to know when to remove it or how to change it if needed. Ideally haskell.nix would provide some way to patch Haskell packages that need tools in the path.

Alternatively, one could consider building `liquid-fixpoint` with the flag `link-z3-as-a-library`, so z3 is not needed in the path.

quantum9Innovation:
> Hello! Glad to see Liquid Haskell building with haskell.nix.
> 
> I would prefer to keep this outside the repo, if possible. Being specific to haskell.nix, it is going to be difficult for maintainers to know when to remove it or how to change it if needed. Ideally haskell.nix would provide some way to patch Haskell packages that need tools in the path.
> 
> Alternatively, one could consider building `liquid-fixpoint` with the flag `link-z3-as-a-library`, so z3 is not needed in the path.

I see, that seems like it could work! I'll keep the changes on my fork and keep experimenting with this; thanks for the advice!

PR #2613, opened 2026-01-12
Remove FinField module from liquid-prelude

### LINKED ISSUE #2610: CVC5 is an hard dependency if the build instructions in the readme are followed
Running `cabal build liquid-prelude` builds the finite field modules that require cvc5 to be installed or the compilation fails with a non very nice error message, I think we should have some conditional compilation.

facundominguez:
Closes #2610


facundominguez:
Closing in favor of #2614

PR #2633, opened 2026-03-14
Indicate the expected Cabal version to detect old cabal-install versions

facundominguez:


facundominguez:
I just tried it with cabal-install 3.14.2.0 and it doesn't improve the build error message. We will need to find some other way.

PR #2636, opened 2026-03-15
Add script to check and install dependencies automatically

AlecsFerra:


AlecsFerra:
No clue if there is a cross platform way to install z3

facundominguez:
Hello!

I'm not convinced that we should provide scripts for the user to install system dependencies like ghc, cabal-install or z3. These things depend on the peculiarities of each user system, and I don't think we are going to provide much value trying to accommodate them all.

I do think, though, that LH should be checking for the minimum z3 version, and cvc5 versions.

cabal-install already checks the ghc version well enough, but it doesn't check for the cabal-install version. Maybe a script similar to this one could check in CI that the cabal install version there matches what the README recommends.

AlecsFerra:
Fine with not providing the install option but clearly cabal is not good enough at reporting mismatched versions of cabal or ghc

facundominguez:
I'm closing this, as I understand we agree to not merge it. Please, let me know if this is not what you meant.

PR #2637, opened 2026-03-15
Add modern flag for LH

AlecsFerra:


liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
```diff
@@ -428,6 +428,9 @@ defConfig = Config {
     = False &= help "Dump time measures of the Liquid Haskell plugin"
           &= name "ddump-timings"
           &= explicit
+  , modern
+    = False &= help "Enable modern features (e.g. reflection, ple, etc.)"
```
facundominguez: Currently, the user has no way to know what `--modern` enables without looking at the source code.

Probably the implied flags should be listed explicitly here.

AlecsFerra: Sure, I will also update the documentation

facundominguez (approved):
LGTM!

PR #2642, opened 2026-03-20
Goodbye --exact-data-cons

### LINKED ISSUE #1284: Generate `is$...` measures regardless of `exactDC`
The following code fails with `Unbound symbol is$Blank.A`, and works with `--exactdc` but that is really odd. We ought to just _always_ generate the `is$Blank.A` tests regardless as they appear to be used in `CoreToLogic`.

```haskell
module Blank where

data Foo = A | B

data Bar = C Foo Foo

{-@ data Baz = D { db :: {v:Bar | isEqBar v} } @-}
data Baz = D Bar

{-@ measure isEqBar @-}
isEqBar :: Bar -> Bool 
isEqBar (C A A) = True 
isEqBar (C B B) = True 
isEqBar (C _ _) = False 
```

**Note:** that we can't just make `--exactdc` default because that _also_ strengthens ctors with the type:

``` 
C :: x:_ -> y: _ -> {v:_ | v = C x y}
```

which can cause problems if `x` and `y` are function sorted... c.f. https://circleci.com/gh/ucsd-progsys/liquidhaskell/2105

### LINKED ISSUE #2292: Fix crashes and obscure errors in tests when enabling --exact-data-cons
#2290 tried enabling `--exact-data-cons` by default, but this causes many tests to fail with crashes and error messages that are hard to relate to the flag. While the feature could still be disabled in the tests, we don't want users to be confronted with the problem of discovering how to get rid of such errors.

This issue is about improving those error messages.

Earlier discussion on enabling `--exactdc` by default can be found in https://github.com/ucsd-progsys/liquidhaskell/issues/1284.

------
Work in progress: https://github.com/ucsd-progsys/liquidhaskell/compare/fd/exactdc

facundominguez:
Fixes #2292
Fixes #1284

This PR addresses the failures reported in #2292, and replaces `--exact-data-cons`
and `--no-adt` with a single flag `--adt` that only controls whether data types
are declared to Liquid Fixpoint.

`--exact-data-cons` used to have the following effects:
* Create selectors and checkers for types used in type signatures of reflected
  functions and measures
* Strengthen the environment of alternatives with facts about which constructors
  had been checked in preceding alternatives
* Declared data types to Liquid Fixpoint.

The first two of these effects are now done unconditionally, while the third
effect is enabled by `--adt`. This simplifies the flags that users need to be
aware of, and eliminates the infectious nature of `--exact-data-cons`: once
enabled in a module, it needed to be enabled in the modules that imported it
transitively.

Turning the above effects unconditionally required in turn addressing the
issues discussed in #2292.

A summary of the fixes follows.

* Selectors and checkers are generated only for types whose data constructors
  are actually used in the bodies of reflected functions and measures. This
  allows to avoid generation for cases that are not needed and that could be
  problematic (e.g. because of UNPACK pragmas). 
* Added a default embedding for `Bool` to ensure that checkers returning `Bool`
  are accepted by the SMT solver, even if the user does not explicitly import
  the embedding from an LHAssumptions module.
* Now Liquid Haskell rejects data annotations where the same selector name is
  used for different field types across constructors.


facundominguez:
This is the first PR where I'm using copilot to help me write a good part of the code.

facundominguez:
CPU time comparisons do not show significant differences.
Wall-clock times do seem to improve a bit.

top
bot

AlecsFerra:
This is great!

ranjitjhala:
Amazing thank you!!!

- Ranjit.

PR #2654, opened 2026-04-16
Update fixpoint, add regression test for #2652

### LINKED ISSUE #2652: An unprovable function equality
The following program doesn't pass verification.
```
module Test where

{-@ LIQUID "--modern" @-}
{-@ LIQUID "--save" @-}

{-@ reflect f @-}
f :: Int -> Int
f x = if x >= 0 then x else x

{-@ reflect g @-}
g :: Int -> Int
g x = if x <= 0 then x else x

{-@ test :: { f = g } @-}
test :: ()
test = ()
```

Looking at the generated equalities, it looks like PLE cannot unfold `f` and `g`.

```
constraint id 6:
  GHC.Internal.Types.True
  
  (~ (GHC.Internal.Types.False))
  
  (Test.f = (\ eta##0 : int -> (Test.f eta##0)))
  
  (Test.g = (\ eta##1 : int -> (Test.g eta##1)))
  
  (VV##F##6 = GHC.Internal.Tuple.$40$$41$)
```

Can this equality be proved somehow?

AlecsFerra:
Companion of https://github.com/ucsd-progsys/liquid-fixpoint/pull/838 should fix #2652

### DIFF
--- a/tests/ple/pos/T2652.hs
+++ b/tests/ple/pos/T2652.hs
@@ -1,0 +1,15 @@
+{-@ LIQUID "--modern" @-}
+
+module T2652 where
+
+{-@ reflect f @-}
+f :: Int -> Int
+f x = if x >= 0 then x else x
+
+{-@ reflect g @-}
+g :: Int -> Int
+g x = if x <= 0 then x else x
+
+{-@ test :: { f = g } @-}
+test :: ()
+test = ()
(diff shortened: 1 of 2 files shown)


facundominguez:
LGTM. I've been thinking for alternatives without finding anything better so far. :shrug:

Thanks @AlecsFerra!

PR #2656, opened 2026-04-24
Remove the custom setup for "fast" recompilations

facundominguez:
This setup introduced a dependency on cabal which sometimes would cause troubles during upgrades. Moreover, the feature it provided for faster recompilations was seldom used.


facundominguez:
Someone is still using this and it is not difficult to keep.

PR #2660, opened 2026-04-30
Fix spurious invariant propagation for measures on type synonyms

facundominguez:
Replace tcUnifyTy with tcMatchTy in goodInvs to prevent measure invariants from being applied to binders whose type arguments are too general. Previously, a measure on [Diff [a]] (via a type synonym Hunk a) would have its Nat invariant (ms v >= 0) spuriously attached to plain [a] binders, causing elaboration failures.

The fix uses one-directional matching so that an invariant's type args must be matchable *from* the target type args (i.e., the target must be an instance of the invariant's expected type). This prevents a type variable in the target from absorbing a structured invariant arg.


ninioArtillero (commented):
LGTM! Thanks for addressing this!

Issue #2544, opened 2025-05-25
Exponential growth of constraints on the number of lemma applications


facundominguez:
The following example passes verification:
```Haskell
{-@ LIQUID "--exactdc" @-}
module Test where

import Language.Haskell.Liquid.ProofCombinators

{-@
f :: xs:[a]
  -> ys:[a]
  -> zs:[a]
  -> ws:[a]
  -> vs:[a]
  -> us:[a]
  -> { append xs (append ys (append zs (append ws (append vs us)))) =
       append (append (append (append (append xs ys) zs) ws) vs) us
     }
@-}
f :: [a] -> [a] -> [a] -> [a] -> [a] -> [a] -> ()
f xs ys zs ws vs us =
  () ? lemmaAssoc xs ys zs                                      --   1 constraint, 2 kvars
     ? lemmaAssoc xs ys (append zs (append ws (append vs us)))  --  65 constraints, 4 kvars
     ? lemmaAssoc (append xs ys) zs (append ws (append vs us))  -- 213 constraints, 6 kvars
     ? lemmaAssoc (append (append xs ys) zs) ws (append vs us)  -- 477 constraints, 8 kvars
     ? lemmaAssoc (append (append (append xs ys) zs) ws) vs us  -- 889 constraints, 10 kvars

{-@
assume lemmaAssoc
  :: xs:[a]
  -> ys:[a]
  -> zs:[a]
  -> { append xs (append ys zs) = append (append xs ys) zs}
@-}
lemmaAssoc :: [a] -> [a] -> [a] -> ()
lemmaAssoc _ _ _ = ()

append :: [a] -> [a] -> [a]
append [] ys = ys
append (x:xs) ys = x : append xs ys

{-@ reflect append @-}
```
LH reports 889 constraints were verified, and the fqout file shows 10 kvars.

If I comment the last application of `lemmaAssoc` the constraints drop roughly by half, although verification starts to fail, of course. The kvars also diminished as noted in the commments. Every time a lemma application is commented out the number of constraints halves again.

This doesn't look good for two reasons:
1. There is no inference of refinement types that could help verifying `f`, so introducing kvars doesn't seem to serve any purpose.
2. The exponential growth of constraints has a serious impact on performance in larger examples with 130 kvars and 56000 constraints.

Can we somehow cut on the number of kvars or checked constraints when introducing lemma applications?

facundominguez:
Using this definition
```Haskell
[...]

PR #2661, opened 2026-05-02
Preserve dead bindings when desugaring

facundominguez:
By marking all identifiers in the user program as exported, we can stop ghc from removing dead bindings. This allows to insert hypothesis in a direct way.

Instead of 
```
f x = () ? lemma x
```
we can say
```
f x = ()
  where
   lem1 = lemma x
```

Improves situation for #2544.

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -1,0 +1,51 @@
+name: "Copilot Setup Steps"
+
+on:
+    workflow_dispatch:
+    push:
+        paths:
+            - .github/workflows/copilot-setup-steps.yml
+    pull_request:
+        paths:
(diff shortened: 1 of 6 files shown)


ranjitjhala:
Did you have a chance to see the performance impact of the extra bindings? (With any luck not bad!)

facundominguez:
I haven't, let's see.

@copilot Please, share here comparison benchmark charts before and after this change as instructed in README.md.
Update: hm, that didn't go too well :see_no_evil:

facundominguez:
hm, some differences to check.

top
bot

facundominguez:
Running benchmarks one at a time shows no significant changes due to dead bindings. The previous run was with parallelism. I think we can merge.

top
bot

PR #2666, opened 2026-05-06
Expand defines in define bodies

### LINKED ISSUE #2548: Using `abs` in logic definition makes reflected function verification fail
The following definitions in `src/GHC/Real_LHAssumptions.hs` should be equivalent:

```haskell
define quot x y =  if x >= 0
                   then (if y >= 0 then x / y else -(x / abs y))
                   else -(abs x / y)
define rem x y = if x >= 0
                 then (if y >= 0 then x mod y else x mod (abs y))
                 else - ((abs x) mod y)

-- the same but using explicit negation instead of `abs`

define quot x y =  if x >= 0
                   then (if y >= 0 then x / y else -(x / -y))
                   else -(-x / y)
define rem x y = if x >= 0
                 then (if y >= 0 then x mod y else x mod -y)
                 else - (-x mod y)
```

Nevertheless, the following example is verified by the latter but rejected due to a **Liquid Type Mismatch** by the former.

```haskell
{-@ LIQUID "--ple" @-}

{-@ reflect intId @-}
intId :: Int -> Int
intId x = (x `quot` 2) * 2 + (x `rem` 2)

{-@ lemmaQuotRem :: x:Int -> { intId x = x } @-}
lemmaQuotRem :: Int -> ()
lemmaQuotRem _ = ()
```

But this is ok:

```haskell
{-@ lemmaQuotRem :: x:Int -> { quot x 2 * 2 + rem x 2  = x } @-}
lemmaQuotRem :: Int -> ()
lemmaQuotRem _ = ()
```

Note `abs` is defined in `src/GHC/Num_LHAssumptions.hs` as 

```haskell
define abs x = if x >=0 then x else -x
```

and—while this verification error is observable—the following spec passes verification:

```haskell
{-@ lemmaAbs :: x:Int -> { (x >= 0 && abs x =  x) || (x < 0 && abs x = -x) } @-}
lemmaAbs :: Int -> ()
lemmaAbs _ = ()
```

> [!IMPORTANT]
> To reproduce the error, it seems to be necessary to rebuild both `liquidhaskell` and `liquid-fixpoint`.

facundominguez:
A reflected function's GHC Core body is translated to fixpoint logic via `CoreToLogic`. When a function call (e.g., `quot x 2`) is encountered, `eAppWithMap` expands it using the corresponding `define` body. The old implementation performed a **single-level substitution** only — it did NOT recursively expand nested define references in the result.

This commit has `eAppWithMap` expand nested define references in the **body before substitution**.

Fixes #2548

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Types.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Types.hs
@@ -291,9 +291,47 @@ eAppWithMap :: LogicMap -> Symbol -> [Expr] -> Expr -> Expr
 eAppWithMap lmap f es expr
   | Just (LMap _ xs e) <- M.lookup f (lmSymDefs lmap)
   , length xs == length es
-  = F.subst (F.mkSubst $ zip xs es) e
+  -- Expand nested define references in the body *before* substitution.
+  -- Arguments are already fully expanded by callers, so substituting them
+  -- into the pre-expanded body produces the correct result without redundant
+  -- traversals over the argument expressions.
+  = F.subst (F.mkSubst $ zip xs es) (expandDefineBody lmap (S.singleton f) e)
   | otherwise
   = expr
+
+-- | Recursively expand references to other defines within a define body.
+-- The @visited@ set prevents infinite expansion of recursive defines.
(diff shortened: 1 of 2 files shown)


facundominguez:
cc @ninioArtillero

PR #2677, opened 2026-05-11
Document alias naming convention

### LINKED ISSUE #2673: Specify and enforce a convention for aliases names (lower-case vs upper case)
LH accepts the following lower-cased type alias definition

```haskell
{-@ type lt B = { v : Int | v < B} @-}
```

But fails when used, eg

```haskell
{-@ x :: lt 1  @-}
x = 0
```

results in 

```
../src/Path/To/File.hs:196:12: error:
    Specified type does not refine Haskell type for `Path.To.File.x` (Plugged Init types old)
The Liquid type
.
    lt {1}
.
is inconsistent with the Haskell type
.
    GHC.Internal.Types.Int
.
defined at ../src/Path/To/File.hs:198:1
.
Specifically, the Liquid component
.
    lt {1}
.
is inconsistent with the Haskell component
.
    GHC.Internal.Types.Int
.

HINT: Use the hole '_' instead of the mismatched component (in the Liquid specification)
    |
196 | {-@ x :: lt 1 @-}
    |          ^^^^^
```

The error is fixed by changing `lt` to `Lt`.

On the other hand, lower-cased predicate aliases work just fine:

```haskell
{-@ predicate gt X Y = X > Y @-}

{-@ y :: {v : Int | gt v 0}@-}
y :: Int
y = 1
```

The naming convention of both type and predicate aliases names are not specified in the [documentation](https://ucsd-progsys.github.io/liquidhaskell/specifications/#refinement-type-aliases), but the examples suggest using only upper-case.

ninioArtillero:
Closes #2673

Adds a couple of lines to the specification page on allowed names for both type and predicate aliases.

Namely:

- Type aliases names should follow the same convention as regular Haskell types (haven't checked if this is actually enforced by the parser, but some experiments point in that direction).
- Predicate aliases can start start either with lower or upper case.

### DIFF
--- a/docs/mkDocs/docs/specifications.md
+++ b/docs/mkDocs/docs/specifications.md
@@ -333,7 +333,8 @@ and compelling example.
 
 **Syntax:** The key requirements for type aliases are:
 
-- Value parameters are specified in **upper**case: `X`, `Y`, `Z` etc.
+1. It's name can start with either an upper or lower case letter.
+2. Value parameters are specified in **upper**case: `X`, `Y`, `Z` etc.
 
 ### Type Aliases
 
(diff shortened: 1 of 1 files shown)


facundominguez:
LGTM!

PR #2681, opened 2026-05-16
Document all RTypeBV constructors in RType.hs

facundominguez:


facundominguez:
This documentation is produced from debugging of my own and copilot's. Please, let me know if any of it looks amiss. The documentation is introduced in the second commit. The first commit is a clean up of Show instances for instrumentation sake.

I care about the scoping rules to eventually implement substitutions correctly.

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/RType.hs
```diff
@@ -723,58 +812,191 @@ data RTypeBV b v c tv r
...
+    --
+    -- * @rt_bind@ is the expression binder, e.g. @px@ above.
+    --   The scope of @rt_bind@ is @rt_ty@.
+    -- * @rt_allarg@ is the type of the bound variable, e.g. @a@
+    -- * @rt_ty@ is the body type, e.g. @RAllE "py" ... (Pair a b)@
+    --
+    -- Semantics: when checking @RAllE x tx t \<: t2@, the binder @x:tx@ is
+    -- added to the environment and then unified with all existing bindings of
+    -- compatible type before checking @t \<: t2@.
+    --
```
ranjitjhala: Do we still really use this? Or is this left over from the "implicit refinement types" work?

ranjitjhala (approved):
This is really most useful, thanks @facundominguez !!!

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/RType.hs
facundominguez: Thanks for having a look. It is a left over indeed. I'll get rid of this case.

Issue #2676, opened 2026-05-09
Epic: make all substitutions rapier-style


facundominguez:
Liquid Haskell uses four implementations of substitutions.

* [Language.Fixpoint.Types.Substitutions.rapierSubstExpr](https://github.com/ucsd-progsys/liquid-fixpoint/blob/1fe1ffe5fb874f40f4c456c42c00ac945a9ba3dd/src/Language/Fixpoint/Types/Substitutions.hs#L266)

* [Language.Fixpoint.Types.Refinements.Subable](https://github.com/ucsd-progsys/liquid-fixpoint/blob/1fe1ffe5fb874f40f4c456c42c00ac945a9ba3dd/src/Language/Fixpoint/Types/Refinements.hs#L1103-L1114)

* [Language.Haskell.Liquid.GHC.Play.Subable](https://github.com/ucsd-progsys/liquidhaskell/blob/f212abb6926af22d8d886eae8a237954de0b4d96/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Play.hs#L150-L152)
* [Language.Haskell.Liquid.Bare.Axiom.Subable](https://github.com/ucsd-progsys/liquidhaskell/blob/f212abb6926af22d8d886eae8a237954de0b4d96/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Axiom.hs#L430-L431)

Of these, only the first one has a reasonable chance of not introducing bugs, if implemented correctly.

All of the others require the caller to ensure somehow that the free variables they are putting in the range of the substitution do no get captured by bindings in the term where the substitution happens.

Because bugs in substitutions continue to show up after this many years of Liquid Haskell, I propose that we consolidate the substitution mechanisms into at most one Subable class, and that we make it implement *the rapier* so name captures are harder to make accidentally.

Probably we can generalize a bit the Subable class of Liquid Fixpoint so it is usable with different implementations of name sets and substitution maps.

PR #2684, opened 2026-05-25
Make explicit in types that r can have only three instances

facundominguez:
`r` parameter of RTypes is now constrained to be only one of three types.

It can be `NoReftBV b v`, or `ReftBV b v`, or `UReftBV b v`. This allows to do
case analysis on `r` values when they are polymorphic.

In order to make the instantiation assumption explicit, the `IsReft` type class
has earned methods that allow to cast between `r` and these types.

The PR does some clean up of type classes and fields to make the transition to
use the new methods.
* The `ToReft` class has been eliminated.
* `IsReft` is used instead of `ToReft`.
* Instances of `ToReft` and `IsReft` other than the identified three types have been removed.
* `UReftBV` does not have an `r` parameter anymore.
* `PVar` does not have the `parg` field anymore.
* The `PredicateCompat` class is gone.

Supports #2676

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Types/RType.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Types/RType.hs
@@ -14,6 +14,7 @@
 {-# LANGUAGE TupleSections              #-}
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE TypeOperators              #-}
+{-# LANGUAGE RoleAnnotations            #-}
 {-# LANGUAGE StandaloneDeriving         #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
 
@@ -79,7 +80,8 @@ module Language.Haskell.Liquid.Types.RType (
   , mapUReftV
   , emapUReftVM
   , NoReft
-  , NoReftB(..)
+  , NoReftB
+  , NoReftBV(..)
 
   -- * Parse-time entities describing refined data types
   , SizeFun, SizeFunV (..), szFun
@@ -393,7 +395,7 @@ emapSubstVM f m = F.toKVarSubst . M.fromList <$> mapM (traverse (emapExprVM f)) 
 
 type UsedPVar    = UsedPVarV Symbol
 type UsedPVarV v = UsedPVarBV Symbol v
-type UsedPVarBV b v = PVarBV b v (NoReftB b)
+type UsedPVarBV b v = PVarBV b v (NoReftBV b v)
 
 type Predicate = PredicateV Symbol
 type PredicateV v = PredicateBV Symbol v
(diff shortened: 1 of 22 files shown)


facundominguez:
Ideally, `RTypeBV` wouldn't have an `r` parameter. The `rt_reft` fields would instead be a sum type that allows the three types (like `ConcreteReft b v`), which is as concrete as we can hope the AST to be. But this requires more changes to Liquid Haskell, and I already achieved my objective which is supporting myself to implement the rapier-style substitutions (which needs the ability to do case analysis on the `rt_reft` field).

facundominguez:
Reviewing this PR is perhaps best done by starting with the last commits which materialize the goal, and then reading them backwards to learn about the transition.

facundominguez:
cc @ninioArtillero

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/RType.hs
```diff
@@ -1243,58 +1251,89 @@ class Semigroup r => Meet r where
...
-trueReft :: IsReft r => r
-trueReft = ofReft F.trueReft
+instance Top (NoReftBV b v) where
+  top _ = NoReft
+instance F.Binder b => Top (F.ReftBV b v) where
+  top _ = F.trueReft
+instance F.Binder b => Top (UReftBV b v) where
+  top _ = MkUReft F.trueReft pdTrue
+
+trueReft :: forall r. IsReft r => r
```
ninioArtillero, line 1304: Some documentation explaining what the purpose of this value is would be welcome, specially given that there is a clashing `F.trueReft` also in use across the codebase.

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/RTypeOp.hs
```diff
@@ -385,7 +385,7 @@ emapRefM bscp vf f γ0 (RProp ss t0) =
       <*> emapReftM bscp vf f (map fst ss ++ γ0) t0
 
 emapBareTypeVM
-  :: (Monad m, Ord v1)
+  :: (Monad m, F.Fixpoint v1)
```
ninioArtillero: I can't seem to make the connection between `F.Fixpoint` (pretty printing) and the `Ord` instance. Was the latter unused?

ninioArtillero (commented):
LGTM! This tidies up quite a big chunk of code; the breakdown helps indeed. I'll just leave a couple of comments on what caught my eye.

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/RTypeOp.hs
facundominguez: It was unused, indeed. I have removed them in a new commit. :+1:

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/RType.hs
facundominguez: I added a description in a comment. :+1:

PR #2685, opened 2026-05-27
Implement flag --allocs to plot allocations in benchmark comparisons

facundominguez:
Allocations are more convenient to measure since they depend less on the load of the machine running benchmarks.

One of the nice things about it is that the benchmarks can then run in parallel, taking only a few minutes to measure the whole testsuite.

This PR implements an `--allocs` flag in `plot-performance` which switches all the charts to show allocations instead of time.


facundominguez:
cc @ninioArtillero

ninioArtillero (commented):
LGTM!

facundominguez:
Thanks for looking at it.

Issue #2590, opened 2025-11-07
Name resolution chokes on dependent pairs + type aliases


AlecsFerra:
```haskell
import Language.Haskell.Liquid.ProofCombinators
-- {-@ type Ix ty R = { val:ty | prop val = R @-}

{-@ foo :: (v::Int, { r:Int | r = v }) @-}
foo :: (Int, Int)
foo = (3, 3)


{-@ bar :: (v::Int, Ix Int v) @-}
bar :: (Int, Int)
bar = (3, 3)
```

Gives the following error:
```
    Unknown logic name `v`
    Cannot resolve name
    Maybe you meant one of: snd len fst Max Min
   |
13 | {-@ bar :: (v::Int, Ix Int v) @-}

```

AlecsFerra:
`exprArgs` should also do something with contexts apparently when we transform types to expressions

facundominguez:
`resolveLogicNames` calls to `emapSpecM`, which computes the local variables at each location. As I understand it, `v` in this example is a local variable bound by `v :: Int`. However, it doesn't seem to be picked by `emapSpecM`.

The line that produces the error is https://github.com/ucsd-progsys/liquidhaskell/blob/1695e59ce246ee714357ebc7499aaf61a5645a45/liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs#L828
inside `resolveLogicName`, where `ss` should hold the local variables. In the failure case, `ss` is the empty list. I'm conjecturing that it could be a mistake, and that it should be made to contain `v`.

AlecsFerra:
I was able to move the error to fixpoint in [here](https://github.com/AlecsFerra/liquidhaskell/commit/4db8ac06b3883be261358c2d5a4e712f7ea0a9fd) apparenty now is the constraint generation failing

AlecsFerra:
It looks like dependent pairs are just super broken

AlecsFerra:
Oh now it shows the right error the abstract refinement is inlined wrong

```
TestSKI.hs:7:12: error:
    Illegal type specification for `SKILam.bar`
    SKILam.bar :: k:GHC.Types.Int -> (GHC.Types.Bool, {VV##0 : GHC.Types.Int | m})<\m VV -> {VV##0 : GHC.Types.Int | m}>
    Sort Error in Refinement: {VV##0 : int | m}
    Unbound symbol m --- perhaps you meant: k ?
    Just
  |
[...]

PR #2688, opened 2026-05-29
Change desugaring of dependent tuples

facundominguez:
Before we desugared `(a :: ta, tb)` as `(ta, tb) <{\a -> tb}>`.

Now it is desugared as `(ta, _) <{\a -> tb}>`, which works even if `tb` uses type aliases.

Addresses #2590.


ranjitjhala:
fantastic thanks i think i've suffered from this too!!!

PR #2694, opened 2026-06-03
Fix splitC panic with higher-kinded types

### LINKED ISSUE #2692: splitc unexpected when working with NonEmpty
I have this code that uses the `non-empty` package:
~~~~
module NonEmptyTest where

import qualified Data.NonEmpty as NonEmpty

type List0 = []
type List1 = NonEmpty.T List0

maxHead :: (Real a) => List1 (List1 a) -> (List1 a, [List1 a])
maxHead = NonEmpty.maximumKey (abs . NonEmpty.head . fst) . NonEmpty.removeEach
~~~~

Liquid Haskell 9.14.1 fails with:
~~~~
    Uh oh.
  panic, called at src/Language/Haskell/Liquid/Constraint/Split.hs:266:5 in liquidhaskell-boot-0.9.14.1-f8859d060f627da42e401b681c06e17e1d83f8ee0a8dfe73b2ae27089eb278ba:Language.Haskell.Liquid.Constraint.Split
    (Another Broken Test!!!) splitc unexpected:
(RVar a)
  <:
(RApp Data.NonEmptyPrivate.T<[]>(RApp GHC.Internal.Types.List<[p fldList]>) (RVar a))
   |
13 | maxHead = NonEmpty.maximumKey (abs . NonEmpty.head . fst) . NonEmpty.removeEach
   |                                                             ^^^^^^^^^^^^^^^^^^^
~~~~

The `NonEmpty` functions are:
~~~~
maximumKey :: (Ord b, Foldable f) => (a -> b) -> NonEmpty.T f a -> a
removeEach :: (Traversable f) => NonEmpty.T f a -> NonEmpty.T f (a, f a)
~~~~

I tried to reproduce with `NonEmpty` from `base`. But the problem does not happen there. I guess it has to do with `Foldable` and `Traversable` generalizations.

facundominguez:
There is an invariant of the RApp that is not always maintained:

The types of the RTProp predicates need to match the types of the predicates that the type constructor expects.

This patch reestablishes the invariant at a point that prevents the failure of #2692.

Fixes #2692

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Split.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Split.hs
@@ -277,8 +277,8 @@ splitC _ (SubR γ o r)
     g   = reLocal $ renv γ
 
 traceTy :: SpecType -> String
-traceTy (RVar v _)      = parens ("RVar " ++ showpp v)
-traceTy (RApp c ts _ _) = parens ("RApp " ++ showpp c ++ unwords (traceTy <$> ts))
+traceTy (RVar v _)      = parens ("RVar " ++ show (F.symbol v))
+traceTy (RApp c ts rs _) = parens ("RApp " ++ showpp c ++ unwords (traceTy <$> ts) ++ if null rs then "" else " {rprops=" ++ show (map traceRProp rs) ++ "}")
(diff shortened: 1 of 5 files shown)


facundominguez:
I don't think I can propose a correct invariant before reading more about abstract predicates. Let me retire this for now.

PR #2700, opened 2026-06-08
Detect derived type class instances with source spans and ignore annotation for methods

### LINKED ISSUE #2432: Deriving `Data` turns UNSAFE code to be SAFE
The following example is reported to be liquid safe although it isn't. The problem seems to be introduced by the `deriving (Data)` clause, as the correct liquid type mismatch gets reported when leaving it out. 

_Tested with 8c550dfa4 and GHC 9.10.1._

```haskell
{-@ LIQUID "" @-}
{-# LANGUAGE DeriveDataTypeable #-}
{-# OPTIONS_GHC -fplugin=LiquidHaskell #-}

module T2432 where

import Data.Data (Data)

{-@ data T = A | B  @-}
data T = A | B
  deriving (Data)

{-@ type A = { v : T | v == A } @-}

{-@ toInt :: v : A -> { i : Nat | i == 2 } @-}
toInt :: T -> Int
toInt A = 0
toInt _ = error ""
```

LH reports a cryptic warning at least
```
WARNING: Found false in Test.hs:1:1
WARNING: Found false in Test.hs:7:14-17
```

facundominguez:
This PR detects derived classes using SrcSpans in the Haskell source code.

We only check now the instance methods that are written by the user.

One consequence of this change is that we no longer produce errors for undefined methods. This was implemented by checking generated methods that threw errors. GHC still warns about instances with undefined methods.

This change surfaced methods written by the user that were not previously checked, and now they fail to pass verification. In order to disable checking these, this PR implements ignore annotations for instance methods.

Lastly, we fixed the parsing of --expect-error-containing in LIQUID annotations, which previously discarded all words except the first.

Fixes #2432


liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs
```diff
@@ -70,6 +70,7 @@ import Language.Haskell.Liquid.Types.Types
 import Language.Haskell.Liquid.Types.Visitors
 import Language.Haskell.Liquid.UX.Config
 import Language.Haskell.Liquid.UX.Tidy
+import qualified GHC as Ghc
```
ninioArtillero: Doesn't this disregards the centralization (explicit selection) of the GHC API intended at the `Liquid.GHC.API` module imported above?

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs
```diff
@@ -80,38 +81,33 @@ classCons :: Maybe [ClsInst] -> [Id]
...
+
+-- | Collect the source spans of all 'ClsInstDecl' nodes in the renamed AST.
+-- These cover the full body of every manually-written instance declaration.
+manualInstSpans :: Ghc.TcGblEnv -> [Ghc.SrcSpan]
+manualInstSpans tcg = case Ghc.tcg_rn_decls tcg of
+  Nothing  -> []
+  Just grp ->
+    [ Ghc.getLocA b
+    | d <- Ghc.hsGroupInstDecls grp
+    , Ghc.ClsInstD _ inst <- [Ghc.unLoc d]
```
ninioArtillero, line 106:
```haskell
let Ghc.ClsInstD _ inst = Ghc.unLoc d
```
?

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -121,6 +122,19 @@ plugin = GHC.defaultPlugin {
     parsePlugin opts ms parsedResult = liquidPlugin opts parsedResult $ \cfg ->
       parsedHook cfg ms parsedResult
 
+    -- | Ensure that 'tcg_rn_decls' is populated for use in 'typeCheckResultAction'.
+    -- GHC only keeps renamed syntax when generating HIE files or haddock docs;
+    -- we need it to distinguish manually-written from auto-derived instances.
+    -- We prime 'tcg_rn_decls' with an empty group here so that GHC's own
+    -- accumulation code does: appendGroups emptyRnGroup rn_decls = rn_decls.
```
ninioArtillero: The phrase is odd, perhaps you intend something like

"... here, like what GHC's own accumulation code does: ..." ?

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -600,8 +615,9 @@ makeTargetSrc :: Config
               -> ModGuts
               -> HscEnv
               -> GlobalRdrEnv
+              -> [GHC.SrcSpan]
               -> IO TargetSrc
-makeTargetSrc cfg file modGuts hscEnv rdrEnv = do
+makeTargetSrc cfg file modGuts hscEnv rdrEnv manualSpans = do
```
ninioArtillero: Perhaps `manualSpans` -> `manInstSpans`?

liquidhaskell-boot/src/Language/Haskell/Liquid/WiredIn.hs
```diff
@@ -249,30 +245,4 @@ mkps_ (n:ns) (t:ts) ((f, x):xs) args ps = mkps_ ns ts xs (a:args) (p:ps)
...
-  , show ''Monad
-  , show ''Applicative
-  , show ''Functor
-  , show ''Foldable
-  , show ''Traversable
-  , show ''Fractional
-  -- , "GHC.Enum.Bounded"
-  -- , "GHC.Base.Monoid"
-  ]
+-- Deprecated: Use manualInstSpans from Interface.hs instead
```
ninioArtillero: Why leave this here? I would remove it or add more context about what has been deprecated.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs
```diff
@@ -107,6 +108,18 @@ manualInstSpans tcg = case Ghc.tcg_rn_decls tcg of
...
+-- | Collect the full source spans of instance declarations.
+-- Unlike 'manualInstSpans' which collects spans of individual bindings,
+-- this collects the span of the entire instance declaration.
+instDeclSpans :: Ghc.TcGblEnv -> [Ghc.SrcSpan]
+instDeclSpans tcg = case Ghc.tcg_rn_decls tcg of
+  Nothing  -> []
+  Just grp ->
+    [ Ghc.getLocA d
+    | d <- Ghc.hsGroupInstDecls grp
+    , Ghc.ClsInstD _ _inst <- [Ghc.unLoc d]
```
ninioArtillero, line 119:
```haskell
, let Ghc.ClsInstD _ _inst = Ghc.unLoc d
```
?
Not if this matching also filters as the generator does.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -605,6 +611,71 @@ makeSpecVars cfg src mySpec env measEnv = do
       | b <- Ghc.bindersOfBinds (_giCbs src)
       ]
 
+-- | Partition ignore annotations into those inside instance declarations
+-- and those at the top level, based on source position.
+partitionIgnores :: [Ghc.SrcSpan] -> S.HashSet (F.Located LHName) -> ([F.Located LHName], [F.Located LHName])
```
ninioArtillero: A haddock for each output field role would be nice.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -605,6 +611,71 @@ makeSpecVars cfg src mySpec env measEnv = do
...
+          matchingVars = case matchingInstSpan of
+            Nothing -> []
+            Just instSp ->
+              [ b
+              | b <- Ghc.bindersOfBinds (_giCbs src)
+              , Ghc.occNameString (Ghc.getOccName b) == cName
+              , Ghc.getSrcSpan b `Ghc.isSubspanOf` instSp
+              ]
+      case matchingVars of
+        (c:_) -> Right c
```
ninioArtillero: What are we to make of other matching variables?

liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
```diff
@@ -593,15 +593,33 @@ withPragmas cfg ps action
...
+    -- we're inside double quotes (dq) or single quotes (sq).
+    go acc _  _  []     = [reverse acc | not (null acc)]
+    go acc dq sq (c:cs)
+      | c == '"'  && not sq = go acc (not dq) sq      cs
+      | c == '\'' && not dq = go acc dq       (not sq) cs
+      | isSpace c && not dq && not sq =
+          if null acc then go [] False False cs
+                      else reverse acc : go [] False False cs
+      | otherwise            = go (c : acc) dq sq cs
+    isSpace c = c == ' ' || c == '\t'
```
ninioArtillero: Shouldn't other whitespace characters (e.g. newline) be part of this as well?

liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
```diff
@@ -593,15 +593,33 @@ withPragmas cfg ps action
        liftIO $ setFxVerbosity (loggingVerbosity cfg) -- restore
        pure res
 
+-- | Split a string into words like a shell would: unquoted whitespace
+-- separates tokens, and single or double quoted substrings are kept as
+-- one token (with the quotes stripped).  This allows option values that
+-- contain spaces, e.g. @--expect-error-containing="Cannot ignore m"@.
```
ninioArtillero, line 599:
```haskell
-- Note: Nested quoting, combining both quote types, is disregarded.
```

Or something similar

liquidhaskell-boot/ghc-api-tests/GhcApiTests.hs
```diff
@@ -308,3 +313,86 @@ hasLetNamed name = go
...
+        (_, tcMod) <- typecheckSourceCode "InstMethods" inputSource
+        let (tcg, _) = GHC.tm_internals_ tcMod
+        (iSpan, mBinds) <- liftIO $ case tcg_rn_decls tcg of
+          Nothing  -> fail "No renamed declarations found"
+          Just grp ->
+            case [ ( GHC.getLocA d
+                   , cid_binds inst
+                   )
+                 | d <- hsGroupInstDecls grp
+                 , ClsInstD _ inst <- [GHC.unLoc d]
```
ninioArtillero, line 348:
```haskell
, let ClsInstD _ inst = GHC.unLoc d
```
?

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs
```diff
@@ -80,38 +81,44 @@ classCons :: Maybe [ClsInst] -> [Id]
...
-unfoldDep _                           = []
-
-exprDep :: CoreExpr -> [Id]
-exprDep = freeVars S.empty
+    isGeneratedBinding v =
+      let occ = Ghc.getOccName v
+       in
+          Ghc.isDerivedOccName occ &&
+          -- default methods are always written by the user
+          not (isPrefixOf "$dm" $ Ghc.occNameString occ) &&
```
ninioArtillero: A nitpick

```haskell
not  ("$dm" `isPrefixOf` Ghc.occNameString occ) &&
```
and similarly bellow

ninioArtillero (commented):
LGTM, kudos for the tests and the docs update. Left some comments with my findings (mostly nitpicks and doc suggestions), with a couple of possibly relevant questions.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs
facundominguez: It can be another data constructor than `ClsInstD` and then the pattern matching would cause the computation of the list comprehension to fail.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs
facundominguez: Indeed. I reverted this.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
facundominguez: Rephrased.

liquidhaskell-boot/src/Language/Haskell/Liquid/WiredIn.hs
facundominguez: Removed.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs
facundominguez: I think it changes the behavior of the list comprehension.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
facundominguez: Documented.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
facundominguez: I explained it in a comment now.

liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
facundominguez: Changed to `Data.Char.isSpace`.

liquidhaskell-boot/ghc-api-tests/GhcApiTests.hs
facundominguez: Similar answer as for Interface.hs.

liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
facundominguez: Hm, the note would be necessary if there are expectations about nested quoting. But that doesn't seem to be the case for `shellWords` as shells do not parse nested quoting either.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs
facundominguez: Adopted.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
facundominguez: I renamed manualSpans to methodSpans, and instDeclSpans to instanceSpans.

facundominguez:
Thanks @ninioArtillero!

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Interface.hs
ninioArtillero: Thanks for clearing that out!

PR #2701, opened 2026-06-08
Add `Diff` library static checks to test suite

ninioArtillero:
I've been working on writing LH static checks for several invariants in the [`Diff` package](https://hackage.haskell.org/package/Diff) as part of a project at Tweag. It is still unknown if they would be merged upstream; nevertheless, they leverage recent fixes and exercise much functionality, so they might be a good fit for the test-suite.

### DIFF
--- a/tests/tests.cabal
+++ b/tests/tests.cabal
@@ -1093,6 +1093,7 @@ executable datacon-neg
 flag Diff
      default: False
 
+-- Test annotations for the Diff package
 executable Diff
     import:           common-ghc-options
     main-is:          Main.hs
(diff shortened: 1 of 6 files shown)


facundominguez (approved):
LGTM!

tests/tests.cabal
```diff
@@ -1090,6 +1090,28 @@ executable datacon-neg
     hs-source-dirs:   app
                     , datacon/neg
 
+flag Diff
+     default: False
+
+executable Diff
```
facundominguez, line 1097:
```suggestion
-- Test annotations for the Diff package
executable Diff
```

PR #2706, opened 2026-06-11
preserve signatures of data constructors over modules; reopen

nikivazou:
Before, in `tests/import/lib/ReproDef.hs`, the refinements of the signature of `Refuter` were not imported, and verification of `contradictionThere` was failing.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
@@ -33,7 +33,7 @@ import qualified Language.Fixpoint.Types                    as F
 import qualified Language.Haskell.Liquid.Misc               as Misc -- (nubHashOn)
 import qualified Language.Haskell.Liquid.GHC.Misc           as GM
 import qualified Liquid.GHC.API            as Ghc
-import           Language.Haskell.Liquid.GHC.Types          (StableName, mkStableName)
+import           Language.Haskell.Liquid.GHC.Types          (StableName)
 import           Language.Haskell.Liquid.LHNameResolution
 import           Language.Haskell.Liquid.Types.Errors
 import           Language.Haskell.Liquid.Types.DataDecl
@@ -1525,17 +1525,11 @@ makeLiftedSpec name src env refl sData sig qual myRTE lSpec0 = lSpec0
   , Ms.qualifiers = filter (isLocInFile srcF) (gsQualifiers qual)
   }
   where
-    myDCs         = mkConSigs (gsCtors sData)
+    myDCs         = filter (isLocalName . val . fst) $ mkSigs (gsCtors sData)
     mkSigs xts    = [ toBare (x, t) | (x, t) <- xts
                     , not (S.member x reflVars) && isExportedVar (toTargetSrc src) x
                     ]
-    mkConSigs xts = [ toBareCon (x, t) | (x, t) <- xts
-                    , not (S.member x reflVars)
-                    , isLocalDataConVar x
-                    , isExportedDataConVar (toTargetSrc src) x
-                    ]
     toBare (x, t) = (makeGHCLHNameLocatedFromId x, Bare.specToBare <$> t)
-    toBareCon (x, t) = (makeGHCLHNameLocated x, Bare.specToBare <$> t)
     xbs           = toBare <$> reflTySigs
     reflTySigs    = [(x, t) | (x,t,_) <- gsHAxioms refl]
     reflVars      = S.fromList (fst <$> reflTySigs)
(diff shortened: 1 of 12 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Misc.hs
```diff
@@ -337,10 +337,7 @@ isTupleId :: Id -> Bool
 isTupleId = maybe False Ghc.isTupleDataCon . idDataConM
 
 idDataConM :: Id -> Maybe DataCon
-idDataConM x = case idDetails x of
-  DataConWorkId d -> Just d
-  DataConWrapId d -> Just d
-  _               -> Nothing
+idDataConM = Ghc.isDataConId_maybe
```
facundominguez: I think it is better to inline `idDataConM` than to have an alias for an operation from the GHC API.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin/SpecFinder.hs
```diff
@@ -67,14 +68,18 @@ findRelevantSpecs lhAssmPkgExcludes hscEnv mods = do
...
     loadRelevantSpec eps currentModule = do
-      res <- liftIO $ runMaybeT $
-        lookupInterfaceAnnotations eps (ue_hpt $ hsc_unit_env hscEnv) (hsc_NC hscEnv) currentModule
-      case res of
-        Nothing         -> do
-          mAssm <- loadModuleLHAssumptionsIfAny currentModule
-          return $ fromMaybe (SpecNotFound currentModule) mAssm
-        Just specResult ->
-          return specResult
+      liftIO (lookupLiquidLib currentModule) >>= \case
```
facundominguez: This lookup looks redundant. The new test succeeds with it at least. If this code remains in the end, it would be helpful to have a comment explaining the need for the additional lookup.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -131,15 +131,27 @@ makeTargetSpec cfg localVars lnameEnv lmap targetSrc bareSpec dependencies = do
...
       return lspec { liftedAsmSigs = S.filter (exportedAssumption . val . fst) (liftedAsmSigs lspec) }
 
+    constructorIsExported :: S.HashSet StableName -> Ghc.DataCon -> Bool
+    constructorIsExported exportedNames dc =
+      mkStableName (Ghc.getName dc) `S.member` exportedNames
+
     ghcSpecToLiftedSpec = toLiftedSpec . toBareSpecLHName cfg lnameEnv . _gsLSpec
 
+dataConIdMaybe :: Ghc.Var -> Maybe Ghc.DataCon
+dataConIdMaybe = Ghc.isDataConId_maybe
```
facundominguez: This is yet another alias to isDataConId_maybe. It would be better to inline it to see the real call where it is used.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -131,15 +131,27 @@ makeTargetSpec cfg localVars lnameEnv lmap targetSrc bareSpec dependencies = do
     removeUnexportedLocalAssumptions lspec = do
       tcg <- Ghc.getGblEnv
       let exportedNames = Ghc.availsToNameSet (Ghc.tcg_exports tcg)
+          exportedStableNames = availsToStableNameSet (Ghc.tcg_exports tcg)
           exportedAssumption (LHNResolved (LHRGHC n) _) =
-            case Ghc.nameModule_maybe n of
-              Nothing -> Ghc.elemNameSet n exportedNames
-              Just m -> m /= Ghc.tcg_mod tcg || Ghc.elemNameSet n exportedNames
+            case Ghc.lookupTypeEnv (Ghc.tcg_type_env tcg) n of
```
facundominguez: This lookup is not needed by the new test either. If this code remains in the end, it would be helpful to have a comment explaining the need for the additional lookup.

facundominguez (changes requested):
Thanks @nikivazou.

Only the change of `Constraint/Init.hs` makes the test pass.

The other changes seem redundant. I guess you either have to strengthen the testsuite or remove the extra bits.

In either case it would be helpful to have more explanation in code comments at some places.

nikivazou:
Thanks @facundominguez !
I remove most unrequired edits. Let me know if I missed anything.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare.hs
```diff
@@ -1525,11 +1525,17 @@ makeLiftedSpec name src env refl sData sig qual myRTE lSpec0 = lSpec0
   , Ms.qualifiers = filter (isLocInFile srcF) (gsQualifiers qual)
   }
   where
-    myDCs         = filter (isLocalName . val . fst) $ mkSigs (gsCtors sData)
+    myDCs         = mkConSigs (gsCtors sData)
```
facundominguez: This change also seems to make no difference to the tests.

facundominguez (changes requested):
:wave: Checked again. There is still one change that looks untested in Bare.hs.

nikivazou:
Correct, it is not needed for the original example either.... 

Thanks a lot! When circle says ok, I will merge!

facundominguez (approved):

PR #2664, opened 2026-05-05
Optimize and document (?)

facundominguez:
This PR introduces a core pass which runs after ANF normalization and before constraint generation. It replaces fully-applied `(?) @A @b e1 e2` expressions with just `e1`.

The pass looks up (?) in the GlobalRdrEnv (tcg_rdr_env) to obtain its Name, then uses Name equality for fast identification in Core. If the module does not have (?) in scope, the pass is skipped entirely.

This avoids introducing KVars for the applications of (?), while the lemma argument (e2) remains as an ANF binding so its postcondition still enters the verification environment.

This eliminates the growth of queries for the SMT solver with each (?) application without --reflection and makes T1660 independent of (?)'s LH signature.

A second commit documents (?) and eliminates the now redundant refinement type signature.

Fixes #2544.
Fixes #2617.

PR #2708, opened 2026-06-16
add Refcore calculus, extraction and test

clayrat:
Extract LiquidHaskell's reflected definitions and refinement types into an intermediate LiquidHaskell calculus (ILH) representation and serialize it to a binary `.ilhb` file in the companion `.liquid` for consumption by downstream tools.

Added:
* `RefCore` modules in `liquidhaskell-boot`: `Calculus`, `CoreToLH`, `SpecToLH`, `Parse`, `Simplify`, `Extract`, `Names`, `Misc`.
* New flags/pragmas: `--refcore` (extracts the calculus and writes the binary .ilhb) and `--refcore-text` (additionally writes the human-readable `.ilh` text dump for debugging).
* A `Half.hs` sanity test for `--refcore`.

Based on the work of @ColinRothgang, @lowyck and @nikivazou


clayrat:
Should we bump hlint to 3.10 on CI to keep up with the GHC version?

facundominguez:
> Should we bump hlint to 3.10 on CI to keep up with the GHC version?

:wave: Probably.

Is the use case for this PR described somewhere? I'm also curious about the description of RefCore.

nikivazou:
Correct! Can you add some documentation on how to run it, what it returns, etc using the `Half.hs` as the example? 

Otherwise, I believe the edits in the current code are indeed minimal! So, it is in great direction! Thanks!

clayrat:
> Is the use case for this PR described somewhere? I'm also curious about the description of RefCore.

I've added a "Refcore Calculus Extraction" section to the options documentation (`docs/mkDocs/docs/options.md`) describing the flags and their usage. 

The use case is that it extracts reflected definitions and proof terms into a small intermediate calculus and serializes those into a binary file. This can then be ingested by other tools to produce translations into other proof languages, e.g., Rocq. `Refcore` is the name we came up with for this calculus and the tooling around it; the idea is to decouple as much as possible from LH/GHC internals to capture the essence of LH proofs.

docs/mkDocs/docs/options.md
```diff
@@ -695,6 +695,30 @@ typeclasses. One limitation is that proofs cannot be written directly
 within the instance definition unless the `--aux-inline` flag is
 turned on as well.
 
+## Refcore Calculus Extraction
+
+**Options:** `refcore`, `refcore-text`
+
+LiquidHaskell can extract a module's reflected definitions and refinement
+types into an intermediate ILH representation, to be consumption by
```
facundominguez:
```suggestion
types into an intermediate ILH representation, to be consumed by
```

tests/refcore/Half.hs
```diff
@@ -0,0 +1,32 @@
+{-@ LIQUID "--refcore" @-}
```
facundominguez, line 1: I'm guessing that you will need a stronger check to keep the translations from bit-rotting.

facundominguez (commented):
I had a quick look. The testing seems to be in need of some work to keep the new code functioning.

tests/refcore/Half.hs
clayrat: We do have additional tests for the full translation pipeline, but those probably don't belong in the LiquidHaskell codebase. Testing the `Binary` serialize/parse round-trip wouldn't show much either, since it's auto-derived.

What we could do is: add the `--refcore-text` pragma to `Half.hs`, commit a "golden" version of the extracted calculus, and diff against it on a test run. One issue with that idea is that the calculus includes autogenerated GHC names (like `ds_d10r`) that would most likely drift between runs, so some normalization would need to happen before the diff. Is this something like what you had in mind?

facundominguez: Maybe that helps. Surely names can be printed without the suffix.

The tests are going to define the scope of the support. `Half.hs` is pretty small at the moment. Is that using all of the features that the use cases will demand?

nikivazou: Well, the thing is that to really test the output we need a theory prover (e.g., Rocq we use right now, maybe an alternative in the future). So, this toy test is sufficient just to check that the output is produced.

clayrat:
Regarding the CI breakage, there are two orthogonal issues:

* We're using HLint 3.8 which is too old to handle `OrPatterns`, the fix is to switch to HLint 3.10
* On June 18, 2026 Megaparsec 9.8.0 was released which apprently uses a different heuristic for error reporting. This broke the two error reporting test cases in `liquidhaskell-parser` test suite. The fix is to change the expected error string to what 9.8 is producing. 

Do we want to fix these first or deal with them after the current PR?

clayrat:
Oh, I see @facundominguez has already added the megaparsec fix in https://github.com/ucsd-progsys/liquidhaskell/commit/f94a5a8e314409c45ca8cef8b727728b2c548047. I'll open a separate PR for HLint update and then rebase this PR on both to get a green CI.

clayrat:
We discovered that PR #2664 optimises away `?` hints before RefCore has a chance to process them, which breaks the translation. I've now added a mechanism that saves a version of CoreBinds before optimisation when the `--refcore` pragma is enabled.

Issue #2662, opened 2026-05-03
Document using lemmas to prove termination properties


facundominguez:
In the following example some lemmas are necessary to prove termination.

```Haskell
{-@ LIQUID "--ple" @-}

module Test1 where

import Language.Haskell.Liquid.ProofCombinators

{-@ reflect ms @-}
{-@ ms :: [[Int]] -> Nat @-}
ms :: [[Int]] -> Int
ms [] = 0
ms (x:xs) = 1 + m x + ms xs

{-@ reflect m @-}
{-@ m :: xs:[Int] -> {v:Nat | len xs = v}  @-}
m :: [Int] -> Int
m [] = 0
m (_:xs) = 1 + m xs

{-@ f :: xs:[[Int]] -> Int / [ms xs] @-}
f :: [[Int]] -> Int
f [] = 0
f (x:xs) =
  case drop 1 x of
    [] -> 1 + f (xs ? ms xs ? m x)
    y  -> 1 + f ((y : xs) ? ms (y : xs) ? m y ? m x ? ms xs)
```

This example is using reflected functions instead of measures since measures would automatically add the missing facts.

The point of the example is that **if** lemmas are needed, they need to be provided inside the **argument** of the recursive calls. This needs to be documented.

An alternative that is less picky about the exact location of `?` is to use dead bindings as enabled by #2661.

```Haskell
f (x:xs) =
  case drop 1 x of
    [] -> 1 + f xs
    y  -> 1 + f (y : xs)
      where
        lem1 = ms (y : xs)
        lem2 = m y
  where
    lem3 = m x
    lem4 = ms xs
```

Dead bindings are blunter, as they are inserted in all constraints arising from locations where the bindings are in scope, so less precision is required to specify where they should be used.

This method also has the advantage that the lemmas appear with their given names in the environments that show up in error messages.

PR #2709, opened 2026-06-18
Document the use of lemmas

ninioArtillero:
Addresses part of #2662

### DIFF
--- a/docs/mkDocs/docs/specifications.md
+++ b/docs/mkDocs/docs/specifications.md
@@ -1191,15 +1191,15 @@ data Val where
 
 ## Using Lemmas
 
-Sometimes additional context information is need for LH to verify a specification.
+Sometimes additional context information is need to verify a specification.
 To accomplish this, we can use [lemmas](https://en.wikipedia.org/wiki/Lemma_(mathematics)).
 
 The `Language.Haskell.Liquid.ProofCombinators` module provides the `?` combinator,
 which is an infix variant of `const` optimized to pass the refined type of its
-second argument to the constraint context of the first argument.
+second argument to the generated constraint context of the first argument.
 
-For example, `e ? lemma` lets you use the type of the expression `lemma`
-when checking that expression `e` has some expected refinement type.
+For example, `e ? lemma` makes the refined type of the expression `lemma`
+available when checking expression `e`.
 
 Schematically, `?` allows to use the `q x` to check that `e` satisfies `p x e`
 in the following function.
@@ -1219,7 +1219,7 @@ to accomplish this as well:
 f :: x:t0 -> {v:t1 | p x v}
 f x = e
   where
-    -- the lemma0 binding is unused by e but it is still available
+    -- the lemma0 binding is unused by e, but it is still available
     -- for Liquid Haskell to use it
     lemma0 = lemma x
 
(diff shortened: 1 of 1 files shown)


docs/mkDocs/docs/specifications.md
```diff
@@ -1187,4 +1187,50 @@ data Val where
...
+
+f :: x:t0 -> {v:t1 | p x v}
+f x = let lemma x {- dead binding -} in e
+```
+
+There are some differences to both approaches:
+
+- Scope: while the former scopes only over `e`,
+  the latter inserts the type in all constraints arising from locations
+  where the binder is in scope.
```
facundominguez: `(f e) ? lemma` only inserts the lemma type in subtyping constraints at `f e`.

If a subtyping constraint at `e` needs the lemma too, you would need to write `(f (e ? lemma)) ? lemma`.

docs/mkDocs/docs/specifications.md
```diff
@@ -1187,4 +1187,50 @@ data Val where
...
+```haskell
+f :: x:t0 -> {v:t1 | p x v}
+f x = e
+  where
+    lemma0 = lemma x -- dead binding
+
+-- alternatively
+
+f :: x:t0 -> {v:t1 | p x v}
+f x = let lemma x {- dead binding -} in e
```
facundominguez:
```suggestion
f x =
  let -- the lemma0 binding is unused by e but it is still available
      -- for Liquid Haskell to use it
      lemma0 = lemma x
   in e
```

docs/mkDocs/docs/specifications.md
```diff
@@ -1187,4 +1187,50 @@ data Val where
...
+```
+
+Dead bindings inside a `where` clause or `let` expression can be used
+to accomplish this as well:
+
+```haskell
+f :: x:t0 -> {v:t1 | p x v}
+f x = e
+  where
+    lemma0 = lemma x -- dead binding
```
facundominguez:
```suggestion
    -- the lemma0 binding is unused by e but it is still available
    -- for Liquid Haskell to use it
    lemma0 = lemma x
```

docs/mkDocs/docs/specifications.md
```diff
@@ -1187,4 +1187,50 @@ data Val where
...
+
+```haskell
+f :: x:t0 -> {v:t1 | p x v}
+f x = e ? lemma x
+
+lemma :: y:t0 -> { q y }
+lemma = ...
+```
+
+Dead bindings inside a `where` clause or `let` expression can be used
```
facundominguez:
```suggestion
Bindings inside a `where` clause or `let` expression can be used
```

docs/mkDocs/docs/specifications.md
```diff
@@ -1187,4 +1187,61 @@ data Val where
...
+
+- Scope: The former inserts the lemma type in the subtyping constraints of
+  the expression value (`e` in the examples above) only, for example:
+  `(f e) ? lemma` inserts the `lemma` type at `f e`, but if a subtyping
+  constraint at `e` needs the `lemma` too, you would need to write
+  `(f (e ? lemma)) ? lemma`.
+  In comparison, the latter option is blunter: it inserts the type in all
+  locations where the binder is in scope, so a single binding inserts it into
+  both `e` and `f e` in this example.
+- The dead binding approach also has the advantage that the lemmas appear
```
facundominguez:
```suggestion
- The binding approach also has the advantage that the lemmas appear
```

facundominguez (approved):
LGTM!

Note that using lemmas for termination properties requires an additional note showing where to insert the lemmas exactly.

PR #2711, opened 2026-06-19
Add docs on using lemmas for termination

### LINKED ISSUE #2662: Document using lemmas to prove termination properties

ninioArtillero:
A complement to #2709 to close #2662

### DIFF
--- a/docs/mkDocs/docs/specifications.md
+++ b/docs/mkDocs/docs/specifications.md
@@ -1251,14 +1251,14 @@ Note that there are some differences between the two approaches:
 - The binding approach has the advantage that the lemmas appear
   with their given names in the environments that show up in error messages.
 
-### For Termination
+### Lemmas for termination checks
 
-When using reflected functions in a [termination metric](#termination-metrics)
-you might need to explicitly pass their post-conditions on the terms
-that intervene as arguments to a recursive call.
+Sometimes proving termination with [termination metrics](#termination-metrics)
+might need to introduce lemmas too. This might be because the metrics use non-linear
+arithmetic, or because the proof obligations require facts not available by
+default.
 
-Let's do a manual calculation to show why the following example fails to
-check termination on `f` and how to fix it:
+Consider the following example, where termination of function `f` fails.
 
 ```haskell
 {-@ LIQUID "--ple" @-}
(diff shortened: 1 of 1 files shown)


docs/mkDocs/docs/specifications.md
```diff
@@ -830,6 +830,15 @@ Decreasing expressions can be arbitrary refinement expressions, e.g.,
 states that at each recursive call of `merge` the _sum of the lengths_
 of its arguments will decrease.
 
+When using a reflected function instead of a measure (like `llen` above) in a metric,
+the reflected function is not automatically applied to the subterms
+of a pattern-matched argument, possibly making the termination check fail.
```
facundominguez: Probably the distinction between reflected functions and measures should be left to be discussed elsewhere. It would need considerably more explanation to be accessible.

docs/mkDocs/docs/specifications.md
```diff
@@ -1246,3 +1255,61 @@ Note that there are some differences between the two approaches:
   so `let lemma0 = lemma in f e` applies the `lemma` to both `e` and `f e`.
 - The binding approach has the advantage that the lemmas appear
   with their given names in the environments that show up in error messages.
+
+### For Termination
```
facundominguez: A manual calculation of the required proofs would make the lemmas more approachable.

docs/mkDocs/docs/specifications.md
```diff
@@ -1246,3 +1250,123 @@ Note that there are some differences between the two approaches:
...
+==  -- unfolding of m (via ple)
+1 + m y + ms xs
+==  -- unfolding of ms (via ple)
+ms (y:xs)
+```
+
+Liquid Haskell can only check condition (4) without further lemmas though.
+In the case of conditions (1) and (2), Liquid Haskell needs
+to be given the fact that `ms xs` produces a `Nat`. Condition (3)
+requires the fact that `m x` produces a `Nat`.
```
facundominguez: This paragraph is what I can infer from the manual proofs, but it is deviating from what Liquid Haskell actually requires empirically.

Try adding only one of the missing lemmas, and you will see that the conditions that remain unchecked are not the ones described here. Also, condition (4) needs the lemmas even though it doesn't appear so in the manual proof.

This needs some investigation, either manual or with copilot.

facundominguez (commented):
I contributed some edits to the section on lemmas for termination checks. Note though, that Liquid Haskell is not behaving exactly as described.

docs/mkDocs/docs/specifications.md
facundominguez: Fixed the proofs, which fixed the paragraph.

facundominguez (approved):
LGTM!

PR #2712, opened 2026-06-23
Document the inlineAux transformation

facundominguez:
This transformation seems to be relevant for type class support.


facundominguez:
cc @JPYamamoto

PR #2715, opened 2026-07-09
Add specification for `Data.Foldable.find`

ninioArtillero:


shell.nix
```diff
@@ -0,0 +1,17 @@
+{
```
facundominguez: Thanks @ninioArtillero. Perhaps this shell.nix file is unintended.

src/Data/Foldable_LHAssumptions.hs
```diff
@@ -1,12 +1,14 @@
-{-# OPTIONS_GHC -fplugin=LiquidHaskellBoot #-}
 {-# OPTIONS_GHC -Wno-unused-imports #-}
+{-# OPTIONS_GHC -fplugin=LiquidHaskellBoot #-}
+
 module Data.Foldable_LHAssumptions where
 
 import Data.Foldable
-import GHC.Types_LHAssumptions()
+import GHC.Types_LHAssumptions ()
```
facundominguez: The reordering of OPTIONS_GHC pragmas and spacing are perhaps unintended.

shell.nix
ninioArtillero: Indeed, it sneaked under.

src/Data/Foldable_LHAssumptions.hs
```diff
@@ -1,12 +1,14 @@
...
 import Prelude hiding (length, null)
 
 {-@
 assume length :: Foldable f => forall a. xs:f a -> {v:Nat | v = len xs}
 assume null   :: Foldable f => forall a. v:(f a) -> {b:Bool | (b <=> len v = 0) && (not b <=> len v > 0)}
+assume find   :: forall f a. Foldable f =>  pred:(a -> Bool) -> f a -> (Maybe {x:a | pred x})
```
facundominguez: I think the forall can be removed.

facundominguez (approved):

Issue #2450, opened 2024-11-28
Revive type class elaboration


facundominguez:
We discovered recently ([#2434](https://github.com/ucsd-progsys/liquidhaskell/pull/2434)) that some of the code to reason about type classes is not tested in the test suite. As discussed in the comments there, we have disabled tests that could be reenabled, and we have other tests that could be considered for inclusion.

A project idea has spontaneously materialized, then, where these tests could guide fixing of the code and improving the documentation. In addition, we also have a paper discussing how the code is intended to work [[link]](https://www.cs.umd.edu/~mwh/papers/lh-typeclasses.pdf).

The code in question is an implementation of section 3.2 of [this paper](https://www.cs.umd.edu/~mwh/papers/lh-typeclasses.pdf). The paper should provide the context necessary to understand the purpose of this part of Liquid Haskell.

After the tests are passing, we could assess whether these tests and the others reasonably cover the paper's promises.

PR #2716, opened 2026-07-16
Restore typeclass support

jpyamamoto:
This PR re-enables the type class infrastructure for GHC 9, fixing several errors that prevented the `--typeclass` tests from passing. The changes include name resolution, measure generation, and expression elaboration.

This solves the issue https://github.com/ucsd-progsys/liquidhaskell/issues/2450 which is one of the main goals for my [GSoC 2026 project](https://summerofcode.withgoogle.com/myprojects/details/lnBMUfdU), but there's still room for improvement in the test suite and the documentation. I expect to work on that too during the coming weeks.

**What changed:**

Changes stream from the addition of the new name resolution mechanism in LH, how the GHC API retrieves dictionary arguments of dictionary data constructors, and minor variation in the result of desugaring during the elaboration pass that produces dictionaries from type information.

- GHC 9 API migration: Updated calls to functions that changed or were removed in GHC 9, e.g. `dataConSig`, `GREInfo`, etc.

- Measure generation for class dictionary constructors: `bkDataCon` and `makeSimplify` in `ToFixpoint.hs` were double-counting or dropping the superclass dictionary fields that class dict constructors carry. The fixes ensure the correct arities.

- Data constructor plugging: `makePluggedDataCon` now includes class constraints from `dataConSig`'s `theta` in the constructor's argument list.

- Evidence bindings are inlined: Evidence bindings (dictionaries, superclass selectors) are now inlined into the elaborated core expression, this is to align with the way this used to work in GHC 8, and make the "wrong number of binders" errors.

- **Tests**: Re-enable `typeclasses/pos` tests. These tests used to work correctly in GHC 8, but were disabled in GHC 9 due to the many errors that were present.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Plugged.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Plugged.hs
@@ -103,7 +103,10 @@ makePluggedDataCon allowTC embs tyi ldcp
                           }
   where
     (tArgs, tRes)     = plugMany allowTC  embs tyi ldcp (das, dts, dt) (dcVars, dcArgs, dcpTyRes dcp)
-    (das, _, dts, dt) = {- F.notracepp ("makePluggedDC: " ++ F.showpp dc) $ -} Ghc.dataConSig dc
+    (das, theta, originDts, dt) = {- F.notracepp ("makePluggedDC: " ++ F.showpp dc) $ -} Ghc.dataConSig dc
+    consClassArgs     = if allowTC then filter (not . GM.isEmbeddedDictType) theta else []
+    dts               = consClassArgs ++ originDts
+
     dcArgs            = reverse $ filter (not . (if allowTC then isEmbeddedClass else isClassType) . snd) (dcpTyArgs dcp)
     dcVars            = if isGADT
                           then padGADVars $ L.nub (dcpFreeTyVars dcp ++ concatMap (map ty_var_value . freeTyVars) (dcpTyRes dcp:(snd <$> dcArgs)))
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Plugged.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Plugged.hs
@@ -149,7 +149,15 @@ plugMany allowTC embs tyi ldcp (hsAs, hsArgs, hsRes) (lqAs, lqArgs, lqRes)
     ((xs,_,ts,_), t) = bkArrow (val pT)
     pT               = plugHoles allowTC (Bare.LqTV dcName) embs tyi (const killHoles) hsT (F.atLoc ldcp lqT)
     hsT              = foldr (Ghc.mkFunTy Ghc.FTF_T_T Ghc.ManyTy) hsRes hsArgs'
-    lqT              = foldr (uncurry (rFun' (classRFInfo allowTC))) lqRes lqArgs'
+    lqT              = foldr (uncurry (rFun' (classRFInfo allowTC))) lqRes' lqArgs'
+    -- Substitute the GHC type variables (hsAs) in lqRes with the corresponding
+    -- LQ type variables (lqAs). This is needed because lqRes (= dcpTyRes) may
+    -- have been derived from ofType/dataConOrigResTy and therefore carries the
+    -- actual GHC TyVar objects, while lqAs (= dcpFreeTyVars) was built from
+    -- symbolRTyVar and therefore carries distinct synthetic RTyVars.
+    -- Without this substitution, runMapTyVars sees the same GHC type variable
+    -- mapped to two different Liquid RTyVars and throws an ErrMismatch.
+    lqRes'           = subts (zip (rTyVar <$> hsAs) lqAs) lqRes
     hsArgs'          = [ Ghc.mkTyVarTy a               | a <- hsAs] ++ hsArgs
     lqArgs'          = [(F.dummySymbol, RVar a mempty) | a <- lqAs] ++ map (Bifunctor.first lhNameToUnqualifiedSymbol) lqArgs
     nTyVars          = length hsAs -- == length lqAs
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Plugged.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Plugged.hs
@@ -98,14 +98,19 @@ makePluggedDataCon allowTC embs tyi ldcp
   | mismatchTyVars    = Ex.throw (err "type variables")
   | otherwise         = F.atLoc ldcp $ F.notracepp "makePluggedDataCon" $ dcp
                           { dcpFreeTyVars = dcVars
+                          , dcpTyConstrs  = constrs
                           , dcpTyArgs     = reverse tArgs
                           , dcpTyRes      = tRes
                           }
   where
     (tArgs, tRes)     = plugMany allowTC  embs tyi ldcp (das, dts, dt) (dcVars, dcArgs, dcpTyRes dcp)
     (das, theta, originDts, dt) = {- F.notracepp ("makePluggedDC: " ++ F.showpp dc) $ -} Ghc.dataConSig dc
-    consClassArgs     = if allowTC then filter (not . GM.isEmbeddedDictType) theta else []
+    consClassArgs     = if allowTC && isClassDc then filter (not . GM.isEmbeddedDictType) theta else []
     dts               = consClassArgs ++ originDts
+    isClassDc         = Ghc.isClassTyCon (Ghc.dataConTyCon dc)
+    -- For class dictionary constructors, the dictionary arguments correspond to
+    -- constructor fields and are included in `dts`, and not in `constrs`.
+    constrs           = if allowTC && isClassDc then [] else dcpTyConstrs dcp
 
     dcArgs            = reverse $ filter (not . (if allowTC then isEmbeddedClass else isClassType) . snd) (dcpTyArgs dcp)
     dcVars            = if isGADT
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Typeclass.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Typeclass.hs
@@ -111,7 +111,7 @@ compileClasses src env (name, spec) rest =
   instClss =
     [ (inst, cls)
     | inst <- mconcat . Mb.maybeToList . _gsCls $ src
-    , Ghc.moduleName (Ghc.nameModule (Ghc.getName inst)) == getModName name
+    , (Ghc.moduleName <$> Ghc.nameModule_maybe (Ghc.getName inst)) == Just (getModName name)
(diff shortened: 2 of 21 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Misc.hs
```diff
@@ -769,18 +769,20 @@ elabRnExpr rdr_expr = do
              tcInferRho rn_expr
 
     -- Generalise
+    -- The outputs of simplifyInfer are dead, but simplifyInfer has some side
+    -- effect that changes the desugaring of full_expr below. Without this
+    -- side effect, the desugaring of full_expr is missing some dictionary
+    -- bindings
```
facundominguez: Maybe this comment is outdated now. We use the `evbs` output in this commit at least.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Misc.hs
```diff
@@ -793,7 +795,31 @@ elabRnExpr rdr_expr = do
       Just e -> do
         when (errorsOrFatalWarningsFound ds_msgs)
           failM
-        return e
+        return $ fst $ Ghc.collectArgs $ inlineEvidenceLets e
+
+inlineEvidenceLets :: CoreExpr -> CoreExpr
```
facundominguez, line 811: This needs a comment explaining what it does.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Misc.hs
```diff
@@ -793,7 +795,31 @@ elabRnExpr rdr_expr = do
       Just e -> do
         when (errorsOrFatalWarningsFound ds_msgs)
           failM
-        return e
+        return $ fst $ Ghc.collectArgs $ inlineEvidenceLets e
```
facundominguez, line 799: I have recollections of `fst . Ghc.collectArgs` being equivalent to a more direct function in the GHC API defined near `Ghc.collectArgs`.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Typeclass.hs
```diff
@@ -427,3 +417,12 @@ substAuxMethod dfun methods = F.notracepp "substAuxMethod" . go
...
         go e = F.notracepp "LEAF" e
+
+mkSymbol :: Ghc.Var -> F.Symbol
+mkSymbol x =
+  case Ghc.getOccString x of
+    '$' : 'c' : 'p' : rest -> F.symbol ('$' : 'p' : rest)
+    '$' : 'c' : rest       -> F.symbol rest
+    occ
+      | Ghc.isDictonaryId x -> F.mappendSym "$" (F.dropSym 2 (F.symbol occ))
+      | otherwise           -> F.symbol occ
```
facundominguez: The changes in this file deserve to be put in a different commit, I think.

It is also necessary to investigate why this wasn't required in GHC 8.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/ToFixpoint.hs
```diff
@@ -272,7 +272,9 @@ makeSimplify (var, t)
       Ghc.DataConWorkId dc ->
         let nCoerce = length $ filter (Ghc.isSimplePredTy . Ghc.irrelevantMult)
                               $ Ghc.dataConRepArgTys dc
-        in  drop nCoerce (ty_binds trep)
+        in if Ghc.isClassTyCon (Ghc.dataConTyCon dc)
```
facundominguez, line 277: Maybe add a comment here:
```
-- Typeclass dictionaries can have other dictionaries as arguments
```

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -856,7 +856,7 @@ resolveLogicNames cfg thisModule env globalRdrEnv lmap0 localVars lnameEnv priva
           Left alts ->
             -- If names are not in the environment, they must be data constructors,
             -- or they must be reflected functions, or they must be in the logicmap.
-            case resolveDataConName ls `mplus` resolveVarName lmap0 ls of
+            case resolveClassMethodName ls `mplus` resolveDataConName ls `mplus` resolveVarName lmap0 ls of
```
facundominguez: The comment above this line needs to be updated to mention class methods.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -931,6 +931,28 @@ resolveLogicNames cfg thisModule env globalRdrEnv lmap0 localVars lnameEnv priva
                  (map (PJ.text . GHC.showPprUnsafe) es)
               )
             return $ makeLocalLHName $ val s
+    resolveClassMethodName s =
+      case GHC.lookupGRE globalRdrEnv (mkLookupGRE (LHVarName LHAnyModuleNameF) $ val s) of
+        [e] ->
+          case (GHC.greInfo e, GHC.greParent_maybe e) of
+            (GHC.Vanilla, Just pName) ->
```
facundominguez: Maybe use patterns guards (`-XPatternGuards`) here, to remove the inner `case` and the inner `if`.
```haskell
  [e]
    | GHC.Vanilla <- GHC.greInfo e
    , Just pName <- GHC.greParent_maybe e
    , isClassTyCon pName ->
      ...
  _ ->
      Nothing
```

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
```diff
@@ -931,6 +931,28 @@ resolveLogicNames cfg thisModule env globalRdrEnv lmap0 localVars lnameEnv priva
...
+              if isClassTyCon pName
+              then Just $ do
+                let n = GHC.greName e
+                let lhName = makeLogicLHName (symbol $ GHC.getOccString n) (GHC.nameModule n) Nothing
```
facundominguez: I'm curious about the last argument. It is `Just` in the case of data constructors, why does it need to be `Nothing` here?

It would deserve a comment in the code.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Plugged.hs
```diff
@@ -106,6 +107,8 @@ makePluggedDataCon allowTC embs tyi ldcp
     (das, theta, originDts, dt) = {- F.notracepp ("makePluggedDC: " ++ F.showpp dc) $ -} Ghc.dataConSig dc
     consClassArgs     = if allowTC then filter (not . GM.isEmbeddedDictType) theta else []
     dts               = consClassArgs ++ originDts
+    isClassDc         = Ghc.isClassTyCon (Ghc.dataConTyCon dc)
+    constrs           = if allowTC && isClassDc then [] else dcpTyConstrs dcp
```
facundominguez, line 113: Please, propagate the conjunct `&& isClassDc` to the `if` in `consClassArgs`, and check that the tests pass. It is likely an unintended omission.

If possible, before fixing it, add a test that rehearses the problematic case: probably a GADT with a typeclass constraint in a file with `--typeclass` enabled. A coding agent might be able to discover such a test if the obvious thing doesn't work.

facundominguez, line 113: Consider adding a comment here like
```
-- We need to avoid introducing the same constraints in `constrs` and `dts`.
-- Otherwise, terms and types generated from these fields would have duplicated
-- occurrences of the constraint.
```

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
facundominguez: Should methods be allowed here if `--typeclass` is inactive? Maybe LH would crash with an ungraceful error message in such a case.

facundominguez (commented):
Please, remember to update the website documentation on type-classes now that type signatures are not required for methods.

liquidhaskell-boot/src/Language/Haskell/Liquid/LHNameResolution.hs
facundominguez: Maybe if the names resolves successfully as a method but `--typeclass` is disabled, the error message could suggest enabling `--typeclass`.

jpyamamoto:
I made the requested changes. I'm using fixup commits so the conversations in this PR aren't lost (though I see a few of them are tagged "outdated").

I also didn't extract the `mkSymbol` change to its own commit, because that would require a rebase. I will do that, and run the `--autosquash`, once the conversations are resolved.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Misc.hs
jpyamamoto: There's the [collectFunSimple](https://hackage.haskell.org/package/ghc-9.14.1/docs/GHC-Core.html#v:collectFunSimple) which is almost equivalent, except that it looks inside Casts and Ticks.

I didn't make that change, but I left a comment documenting this decision.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Plugged.hs
jpyamamoto: I did. A similar fix was required in `Measure.stitchArgs` too.

I added the `ConstrainedGADT` test to check this execution path.

facundominguez:
I contributed some edits to comments. The PR is looking good :raised_hands:

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Typeclass.hs
```diff
@@ -427,3 +417,18 @@ substAuxMethod dfun methods = F.notracepp "substAuxMethod" . go
...
+-- `$p(n)(Class)` selector symbols.  Such bindings reach this function both as
+-- dictionary identifiers and as ordinary variables, so their prefix must be
+-- handled before checking `Ghc.isDictonaryId`.
+mkSymbol :: Ghc.Var -> F.Symbol
+mkSymbol x =
+  case Ghc.getOccString x of
+    '$' : 'c' : 'p' : rest -> F.symbol ('$' : 'p' : rest)
+    '$' : 'c' : rest       -> F.symbol rest
+    occ
+      | Ghc.isDictonaryId x -> F.mappendSym "$" (F.dropSym 2 (F.symbol occ))
```
facundominguez: This case doesn't seem to be used for anything. All tests pass if removed.

jpyamamoto: Turns out, the following case (the `otherwise`) isn't used either. The `mkSymbol` function is only given class methods names as input, which always have the `$c` prefix.

I changed the comment in the code to document that condition.

facundominguez:
Looks good to squash!

facundominguez (approved):
LGTM. Thanks @jpyamamoto!

PR #2724, opened 2026-07-24
Add additional string operations to WiredIn

### LINKED ISSUE #2723: Question about using string operations from `liquid-fixpoint`
Hi,

While trying to define a SMT-backed string type in Liquid Haskell, I noticed that `liquid-fixpoint` defines several string operations, including `strLen`, `strConcat`, `strSubstr`, and others:
[Language/Fixpoint/Smt/Theories.hs#L208](
https://github.com/ucsd-progsys/liquid-fixpoint/blob/a34ed38cf5f48c7c272b2aa9c28e53592775cab6/src/Language/Fixpoint/Smt/Theories.hs#L208)

However, in Liquid Haskell, only `strLen` is exposed in `WiredIn.hs`: [Language/Haskell/Liquid/WiredIn.hs#L126](https://github.com/ucsd-progsys/liquidhaskell/blob/2e9a590fce72dfa5d718c43f57043e3cb27bc78f/liquidhaskell-boot/src/Language/Haskell/Liquid/WiredIn.hs#L126).

When I added the other functions to this file locally, they became available as logic names in Liquid Haskell.

My questions are:
1) Is modifying Liquid Haskell the only way to use these additional string operations? If so, I'd be happy to submit a PR exposing them.
2) What is the difference between `concatString` and `strConcat` ?

Thanks

tsung-ju:
Resolves #2723

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/WiredIn.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/WiredIn.hs
@@ -124,6 +124,11 @@ wiredTheorySortedSyms =
       , "FF_mul"
 
       , "strLen"
+      , "subString"
+      , "strConcat"
+      , "strPrefixOf"
+      , "strSuffixOf"
+      , "strContains"
       ]
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


facundominguez:
Thanks @tsung-ju! Can you add a regression test that checks this is accomplishing what you want?

tsung-ju:
Added tests

facundominguez (approved):
LGTM!

ranjitjhala:
Fantastic thank you!!

PR #2719, opened 2026-07-19
cmdline: export shellWords and add regression tests for LIQUIDHASKELL_OPTS parsing (#1990)

SNAPKITTYWEST:
## Summary

Closes #1990.

The multi-argument fix for `LIQUIDHASKELL_OPTS` was already applied as part of the GetOpt migration (the env var is now parsed with `shellWords` which correctly splits on whitespace). However, issue #1990 remained open because:

1. `shellWords` was not exported, making it impossible to unit-test independently
2. No regression tests existed to prevent future breakage

This PR:
- Exports `shellWords` from `Language.Haskell.Liquid.UX.CmdLine`
- Adds a `testShellWords` group to the existing `liquidhaskell-parser` test suite covering: empty input, single flag, multiple flags, flag-with-value, leading/trailing spaces, and quoted values

## Test plan

```
cabal test liquidhaskell-parser
```

All new `shellWords` cases should pass, confirming that `LIQUIDHASKELL_OPTS="--ple --reflection"` correctly produces two separate flags rather than one.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

PR #2731, opened 2026-08-05
cmdline: export shellWords and add regression tests for LIQUIDHASKELL_OPTS parsing (#1990)

### LINKED ISSUE #1990: LIQUIDHASKELL_OPTS can hold only one argument
By inspecting the code in src/Language/Haskell/Liquid/UX/CmdLine.hs we see that the contents of the env var is parsed as a single argument. This should be changed to deal with multiple arguments. And we should confirm this bug, since we didn't discover it with a test but by code inspection.

SNAPKITTYWEST:
Closes #1990. Replaces #2719 (accidentally closed by branch deletion).

The multi-argument fix for LIQUIDHASKELL_OPTS was already applied as part of the GetOpt migration (shellWords correctly splits on whitespace). This PR adds the missing test coverage:

- Exports shellWords from Language.Haskell.Liquid.UX.CmdLine
- Adds testShellWords group to the liquidhaskell-parser test suite (empty input, single flag, multiple flags, flag-with-value, leading/trailing spaces, quoted values)

Test plan: cabal test liquidhaskell-parser

Review history preserved in #2719.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
@@ -35,6 +35,9 @@ module Language.Haskell.Liquid.UX.CmdLine (
    -- * Diff check mode
    , diffcheck
 
+   -- * Shell-word splitting (used by LIQUIDHASKELL_OPTS parsing)
+   , shellWords
+
 ) where
 
 import Prelude hiding (error)
--- a/liquidhaskell-boot/tests/Parser.hs
+++ b/liquidhaskell-boot/tests/Parser.hs
@@ -20,8 +20,9 @@ import           Data.Generics.Aliases
 import           Data.Generics.Schemes
 
 import           Language.Fixpoint.Types.Spans
-import qualified Language.Haskell.Liquid.Parse as LH
-import qualified Language.Fixpoint.Types       as F
+import qualified Language.Haskell.Liquid.Parse           as LH
+import qualified Language.Fixpoint.Types                 as F
+import           Language.Haskell.Liquid.UX.CmdLine      (shellWords)
 
 import           Text.Megaparsec.Error
 import           Text.Megaparsec.Pos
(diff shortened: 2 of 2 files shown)


facundominguez (approved):
LGTM! Thanks @SNAPKITTYWEST

SNAPKITTYWEST:
I found what appears to be an inconsistency in the LiquidHaskell
`GHC.Real_LHAssumptions` specifications for the Integral operations.

The current specification contains:

rem :: x:a -> y:{v:a | v /= 0}
  -> {v:a | (v = rem x y) &&
            ((v >= 0) && (v < y))}

However, the corresponding logical definition is:

define rem x y =
  if x >= 0 then x mod y else -((- x) mod y)

The definition therefore permits a negative result when x < 0,
while the refinement requires:

v >= 0

This appears inconsistent with the intended semantics of Haskell's
`rem`, where the remainder has the sign of the dividend.

There is also a related issue with the logical definition:

define mod x y =
  if y > 0 then x mod y else -((- x) mod y)

The definition appears to refer to `mod` recursively rather than
providing an independent terminating logical model for the operation.

Similarly:

define div x y =
  if y > 0 then x / y else -((- x) / y)

appears to use `/` while modeling the Integral `div` operation.

Minimal example

A concrete semantic counterexample to the `rem` refinement is:

(-7) `rem` 3 = -1

The refinement requires:

v >= 0
which cannot hold for this result.


Please advise whether this is an intentional abstraction in the
GHC.Real_LHAssumptions` model or an error in the specifications /
logical definitions.

facundominguez:
:wave: Hello!

```
define mod x y = if y > 0 then x mod y else -((- x) mod y)
```

* `mod` on the left-hand side refers to `Prelude.mod`, whereas in the right-hand side it refers to an infix `mod` provided by SMT solvers.
* Similarly with `div` and `/`.
* I think the refinement type of `rem` needs to be adjusted.

There is some discussion about how to improve the naming of the operators for integer division in https://github.com/ucsd-progsys/liquidhaskell/issues/2735

PR #2732, opened 2026-08-06
Eliminate uses of const when eliminating (?) as well

facundominguez:
When working on the Diff package it was convenient to use the `?` operator to introduce lemmas. However, this would mean that Diff modules had to import `Language.Haskell.Liquid.ProofCombinators` which would make Diff depend on `liquid-prelude`.

While this could be handled with conditional compilation, I thought it would be simpler to just optimize `const` in the same way that LH optimizes `?` to solve #2544. This is what this PR does.

In addition, the PR fixes the name lookup of `const` and `?` in the GlobalRdrEnv, which could fail sometimes depending on how the identifiers were brought in scope.

Benchmarks show only positive changes, and this happens only on two files which unsurprisingly use `const` instead of `?` in proofs. These are `tests/typeclasses/pos/Lemma.hs` and `tests/typeclasses/pos/All.hs`.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Transforms/QuestionMark.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Transforms/QuestionMark.hs
@@ -25,15 +25,14 @@ module Language.Haskell.Liquid.Transforms.QuestionMark (eliminateQuestionMark) w
 import qualified Language.Haskell.TH.Syntax as TH (nameModule)
 import Liquid.GHC.API as Ghc
 
--- | Look up the '?' operator in the 'GlobalRdrEnv'. If found, eliminate all
--- its applications from the Core program. If not found (module doesn't import
--- ProofCombinators), return the bindings unchanged.
+-- | Look up the 'FunctionNames' in the 'GlobalRdrEnv'. If found, eliminate all
+-- their applications from the Core program. If not found, return the bindings unchanged.
 eliminateQuestionMark :: GlobalRdrEnv -> [CoreBind] -> [CoreBind]
 eliminateQuestionMark rdrEnv cbs =
   case lookupFunctionNames rdrEnv of
     Nothing    -> cbs
     Just names -> map (goBind names) cbs
-
+-- | A record of function names whose applications are to be removed form the core program.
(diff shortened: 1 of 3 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/Transforms/QuestionMark.hs
```diff
@@ -13,38 +14,74 @@
...
+-- liquid-prelude is unwanted and therefore `?` is unavailable.
 
 module Language.Haskell.Liquid.Transforms.QuestionMark (eliminateQuestionMark) where
 
+import qualified Language.Haskell.TH.Syntax as TH (nameModule)
 import Liquid.GHC.API as Ghc
 
 -- | Look up the '?' operator in the 'GlobalRdrEnv'. If found, eliminate all
 -- its applications from the Core program. If not found (module doesn't import
 -- ProofCombinators), return the bindings unchanged.
```
ninioArtillero:
```suggestion
-- | Look up the 'FunctionNames' in the 'GlobalRdrEnv'. If found, eliminate all
-- their applications from the Core program. If not found, return the bindings unchanged.
```

Or unfold the `FunctionNames` is preferred.

liquidhaskell-boot/src/Language/Haskell/Liquid/Transforms/QuestionMark.hs
```diff
@@ -13,38 +14,74 @@
...
 eliminateQuestionMark :: GlobalRdrEnv -> [CoreBind] -> [CoreBind]
 eliminateQuestionMark rdrEnv cbs =
-  case lookupQuestionMark rdrEnv of
-    Nothing   -> cbs
-    Just name -> map (goBind name) cbs
+  case lookupFunctionNames rdrEnv of
+    Nothing    -> cbs
+    Just names -> map (goBind names) cbs
+
+data FunctionNames = FunctionNames
```
ninioArtillero:
```suggestion
-- | A record of function names whose applications are to be removed form the core program.
data FunctionNames = FunctionNames
```

ninioArtillero (commented):
Looks good; just proposed some amends to documentation.

Issue #2547, opened 2025-05-26
Logic `/` and `mod` don´t match Haskell's `div` and `mod`.


ninioArtillero:
Logic `/` and `mod` (from `liquid-fixpoint`) don´t match Haskell's `div` and `mod` behavior, as they should given that the former is used to define the latter in `src/GHC/Real_LHAssumptions.hs`.

``` haskell
{-@
...

define div x y        = (x / y)
define mod x y        = (x mod y)

...
@-}
```

As a consequence, in the next example both functions fail verification but should not:

``` haskell
{-@exampleHaskellDiv :: {x:Int | x = 2} @-}
exampleHaskellDiv :: Int
exampleHaskellDiv = div (-5) (-2)

{-@ exampleHaskellMod :: {x:Int | x = -1} @-}
exampleHaskellMod :: Int
exampleHaskellMod = mod (-5) (-2)
```

On the other hand, this is ok:

``` haskell
{-@ exampleLogicDiv :: { 3 = (-5) / (-2)} @-}
exampleLogicDiv :: ()
exampleLogicDiv = ()

{-@ exampleLogicMod :: { 1 = (-5) mod (-2)} @-}
exampleLogicMod :: ()
exampleLogicMod = ()
```

ninioArtillero:
More examples of the logic operations (for reference):

``` haskell
{-@ exampleLogicDiv :: {
                         3 = (-5) / (-2)
                      && 2 = (5) / (2)
                      && -3 = (-5) / (2)
                      && -2 = (5) / (-2)
                      && 1 = (-1) / (-2)
                      && 0 = (1) / (2)
                      && -1 = (-1) / (2)
                      && 0 = (1) / (-2)
                       } @-}
exampleLogicDiv :: ()
exampleLogicDiv = ()

{-@ exampleLogicMod :: {
                         1 = (-5) mod (-2)
                      && 1 = (5) mod (2)
                      && 1 = (-5) mod (2)
                      && 1 = (5) mod (-2)
                      && 1 = (-1) mod (-2)
                      && 1 = (1) mod (2)
                      && 1 = (-1) mod (2)
                      && 1 = (1) mod (-2)
                       } @-}
exampleLogicMod :: ()
exampleLogicMod = ()
```

tsung-ju:
With the current "euclidean" `/` and `mod` we can define
```haskell
define div x y        = if y > 0 then x / y else -((- x) / y)
[...]

PR #2733, opened 2026-08-07
Change div/mod definition to match Haskell rounding semantics

tsung-ju:
Workaroud for #2547

### DIFF
--- a/src/GHC/Real_LHAssumptions.hs
+++ b/src/GHC/Real_LHAssumptions.hs
@@ -18,11 +18,11 @@ class Num a => Fractional a where
   fromRational :: Ratio Integer -> a
 
 class (Real a, Enum a) => Integral a where
-  quot :: x:a -> y:{v:a | v /= 0} -> {v:a | (v = (x / y)) &&
+  quot :: x:a -> y:{v:a | v /= 0} -> {v:a | (v = quot x y) &&
                                                      ((x >= 0 && y >= 0) => v >= 0) &&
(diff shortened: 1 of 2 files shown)


tsung-ju:
It seems the test is failing because it's still using the old `div` definition. I ran into a similar issue locally, but running `cabal clean` fixed it.

facundominguez:
Thanks @tsung-ju!

It is likely that when edits only concern comments, GHC skips recompilation and therefore the interface files do not get the new annotations.

I erased the cache to try the tests again. Another way would have been to add some dummy haskell definition in Real_LHAssumptions.hs to force the rebuild.

tsung-ju:
Thanks for the tips!

The remaining failing test case is
```haskell
{-@ myDiv :: x:Int -> y:{Int | y != 0} -> {v:Int | v = x / y} @-}
myDiv :: Int -> Int -> Int
myDiv = div
```
which makes sense because now `div` and `/` are different.

If existing code already relies on `div` and `/` being the same, perhaps the better approach is to change how `/` is translated to SMT.

facundominguez:
For integers, `/` is translated to `SMTLIB_OP_DIV` which is defined like this in the smt solver.
```
(define-fun SMTLIB_OP_DIV ((x Int) (y Int)) Int (div x y))
```

For reals the smt solver has an `/` operator and LH might use it when `/` is applied to reals in expressions.

We could restrict `/` to only work for reals, so we don't have two different integer divisions in the logic. But perhaps that can be done as a separate issue.

PR #2734, opened 2026-08-09
Change div/mod definition to match Haskell rounding semantics

facundominguez:
A continuation of #2733.

Now LiquidHaskell has operations `div` and prefix `mod` in the logic. We should consider renaming `/` for integer division and infix `mod` in a separate issue (#2735).


facundominguez:
cc @tsung-ju

PR #2738, opened 2026-08-16
Qualify the autosize measure name and report mixed-type recursive groups

### LINKED ISSUE #2736: autosize on mutually recursive data types makes the solver die: Constraint with free vars [autolen]
Verified with liquidhaskell 0.9.12.2.1 on ghc 9.12.2 (liquid-fixpoint 0.9.6.3.5) and with liquidhaskell 0.9.14.1.1 on ghc 9.14.1 (liquid-fixpoint 0.9.6.3.7), as a GHC plugin, with z3 5.0.0 on linux. Both versions behave the same.

```haskell
module Repro where

{-@ LIQUID "--reflection" @-}

data Tree = Leaf Int | Node Forest
  deriving (Eq, Show, Read)

data Forest = Forest
  { label :: Int
  , trees :: [Tree] }
  deriving (Eq, Show, Read)

{-@ autosize Tree @-}
{-@ autosize Forest @-}

sizeTree :: Tree -> Int
sizeTree (Leaf _)   = 1
sizeTree (Node frs) = 1 + sizeForest frs

sizeForest :: Forest -> Int
sizeForest (Forest _ ts) = 1 + sizeTrees ts

sizeTrees :: [Tree] -> Int
sizeTrees []       = 0
sizeTrees (t : ts) = sizeTree t + sizeTrees ts
```

Building this module makes liquid-fixpoint die during solving, with no source location:

```
**** LIQUID: ERROR *************************************************************
<no location info>: error:
    HasCallStack backtrace:
  throw, called at src/Language/Fixpoint/Types/Errors.hs:151:7 in liquid-fixpoint-0.9.6.3.5-...:Language.Fixpoint.Types.Errors
  die, called at src/Language/Fixpoint/Solver.hs:248:22 in liquid-fixpoint-0.9.6.3.5-...:Language.Fixpoint.Solver

:1:1-1:1: Error
  Constraint with free vars
  2
  [autolen]
  Try using the --prune-unsorted flag
```

Adding `{-@ LIQUID "--prune-unsorted" @-}` as the message suggests avoids the crash, but the module is then rejected as UNSAFE: the pruning appears to drop the autolen facts, so the automatic sizes are not usable for termination either way.

Expected: either the automatic size handles mutually recursive types (here one of them a record with a list field), or a located error saying it cannot. Explicit size measures written by hand work fine on the same types, so this is only about `autosize`.

shalashaska117:
Fixes #2736. Longer diagnosis in https://github.com/ucsd-progsys/liquidhaskell/issues/2736#issuecomment-5309137197.

Two independent bugs, the second only reachable once the first is out of the way.

`lenLocSymbol` emitted the bare symbol `autolen`, while the measure is declared in `GHC.Base_LHAssumptions` and resolves to `GHC.Base_LHAssumptions.autolen`, so any constraint that reached the solver with an `autolen` term was rejected with `Constraint with free vars [autolen]`. The existing `autosize` tests pass only because their constraints are sliced away before that check runs, so this was never about mutual recursion: `tests/pos/T2736.hs` is a self-recursive list with one signature that keeps a constraint alive, and it crashes on `develop`. `WiredIn.hs` has spelled the list size function `GHC.Types_LHAssumptions.len` since 5991f007d.

`consCBSizedTys` takes the size function from the type of one decreasing parameter and applies it to the decreasing parameter of every binder in the group, so a group whose parameters have different types produced a term the solver cannot sort check, and elaboration panicked with `Uh oh. elaborate solver failed on ...`. `checkEqTypes` already detected the condition and reported `The decreasing parameters should be of same type`, but its result was discarded and the metric was built regardless. It now suppresses the metric, so the error is what the user sees. This half has nothing to do with `autosize`: a hand-written size measure on the same shape panicked identically.

Tests: `tests/pos/T2736.hs` for the first, `tests/errors/T2736.hs` (the module from the issue) for the second. Docs: the restriction on decreasing-parameter types is reachable from user code now that it is an error rather than a panic, so `specifications.md` states it.

Left out on purpose: `makeSizedDataCons` counts only constructor arguments whose sort equals the result sort, so for mutually recursive types the size is constant (`autolen (Node f) = 1` for `Node :: Forest -> Tree`). Counting arguments of the other autosized types would make it grow across the mutual edge, but it changes `autolen` for existing code and does not change the outcome for the module in the issue, where the `[Tree]` hop is the blocker. Happy to do it in a separate PR if you want it.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Termination.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Termination.hs
@@ -184,9 +184,15 @@ consCBSizedTys consBind γ xes
        autoenv  <- gets autoSize
        ts       <- forM ts' $ T.mapM refreshArgs
        let vs    = zipWith collectArgs' ts es
-       is       <- mapM makeDecrIndex (zip vars ts) >>= checkSameLens
+       is0      <- mapM makeDecrIndex (zip vars ts) >>= checkSameLens
+       sameTys  <- checkEqTypes =<< mapM checkIndex (zip4 vars vs ts is0)
+       -- The size function of a group is taken from the type of one decreasing
+       -- parameter and applied to all of them, so a group whose parameters have
+       -- different types has no metric to build. Drop it: building one anyway
+       -- produces a refinement the solver cannot sort check, and the error from
+       -- checkEqTypes never gets seen.
+       let is    = if sameTys then is0 else Nothing <$ is0
        let xeets = zipWith (\v i -> [((v,i), x) | x <- zip3 vars is $ map unTemplate ts]) vs is
-       _        <- mapM checkIndex (zip4 vars vs ts is) >>= checkEqTypes
        let rts   = (recType autoenv <$>) <$> xeets
        γ'       <- foldM extender γ (zip vars ts)
        let γs    = zipWith makeRecInvariants [γ' `setTRec` zip vars rts' | rts' <- rts] (filter (not . noMakeRec) <$> vs)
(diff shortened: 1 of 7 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Termination.hs
```diff
@@ -198,19 +204,21 @@ consCBSizedTys consBind γ xes
...
-       checkAllVsHead :: Eq b => Error -> (a -> b) -> [a] -> CG [a]
-       checkAllVsHead _   _ []          = return []
-       checkAllVsHead err f (x:xs)
-         | all (== f x) (f <$> xs) = return (x:xs)
-         | otherwise               = addWarning err >> return []
+       allVsHead :: Eq b => Error -> (a -> b) -> [a] -> CG Bool
+       allVsHead _   _ []          = return True
+       allVsHead err f (x:xs)
+         | all (== f x) (f <$> xs) = return True
+         | otherwise               = addWarning err >> return False
```
facundominguez: Perhaps this is easier to read

```haskell
checkEqTypes =
  if allEqual (map toRSort $ catMaybes x) then
    return True
  else
    addWarning err1 >> return False

checkSameLens xs =
   if allEqual (map length xs) then
     return xs
   else
     addWarning err2 >> return []

...

-- All elements of the input are equal to each other
allEqual [] = True
allEqual (x:xs) = all (== x) xs
```

facundominguez (approved):
LGTM!

I left one coding style suggestion. Also, there are some conflicts after merging #2739.

> Happy to do it in a separate PR if you want it

Sure, it is good to discuss it separately.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Termination.hs
shalashaska117: I wrote it with a Prolog-like style, I will change it tomorrow

shalashaska117: Changed to `allEqual`, and rebased on `develop`.

PR #2739, opened 2026-08-16
Reject ple annotations when PLE is disabled

### LINKED ISSUE #2737: ple and automatic-instances annotations are silently ignored unless --ple-local is set
Verified with liquidhaskell 0.9.12.2.1 on ghc 9.12.2 and with liquidhaskell 0.9.14.1.1 on ghc 9.14.1, as a GHC plugin. Both versions behave the same.

```haskell
module Repro where

{-@ LIQUID "--reflection" @-}

{-@ reflect double @-}
double :: Int -> Int
double x = x + x

{-@ automatic-instances prop @-}
{-@ prop :: {v:() | double 2 == 4} @-}
prop :: ()
prop = ()
```

This module is rejected as UNSAFE. The annotation parses without complaint, but it has no effect, so the obligation fails exactly as if PLE were off. The same happens with `{-@ ple prop @-}` in place of `automatic-instances`.

Two changes make the same code SAFE: adding `{-@ LIQUID "--ple-local" @-}` next to the per-function annotation, or replacing everything with module-wide `{-@ LIQUID "--ple" @-}`.

The options documentation describes `automatic-instances` together with `--ple-local`, so requiring the flag may be by design. The problem is the silence: an accepted annotation that does nothing looks exactly like a failing proof, and nothing points at the missing flag. It took a while to tell this apart from a genuine proof failure. A warning when `ple` or `automatic-instances` occurs in a module that does not set `--ple-local`, or letting the annotation imply the flag, would remove the trap.

shalashaska117:
Fixes #2737. Discussion in https://github.com/ucsd-progsys/liquidhaskell/issues/2737#issuecomment-5309137273.

`{-@ ple f @-}` and `{-@ automatic-instances f @-}` fill `gsAutoInst`, but `doExpand` reads it only under `--ple-local`, so a module that sets neither `--ple-local` nor `--ple` drops the annotation and the proof fails exactly as if PLE were off. `makeAutoInst` now rejects the annotation in that case, located at the binder:

```
tests/errors/T2737.hs:12:25: error:
    PLE is not enabled for `prop`
    Add {-@ LIQUID "--ple-local" @-} to run PLE on the annotated binders,
    or {-@ LIQUID "--ple" @-} to run it on the whole module.
   |
12 | {-@ automatic-instances prop @-}
   |                         ^^^^^
```

`--ple` on its own is still accepted. PLE runs for the whole module there, so the annotation is redundant rather than dead, and six tests pair the two: `tests/ple/pos/IndPal0.hs`, `IndPal00.hs`, `IndPalindrome.hs` and `tests/benchmarks/popl18/ple/pos/{Lists,Overview,ApplicativeList}.hs`. Say so if you would rather reject the redundant case as well, and I will drop the annotations from those tests.

Test: `tests/errors/T2737.hs`. Docs: `options.md`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - Reject `ple` and `automatic-instances` annotations in modules that enable
   neither `--ple-local` nor `--ple`
   [#2737](https://github.com/ucsd-progsys/liquidhaskell/issues/2737)
+  [#2739](https://github.com/ucsd-progsys/liquidhaskell/pull/2739)
 
 ## 0.9.14.1.1 (2026-06-04)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Next
 
+- Reject `ple` and `automatic-instances` annotations in modules that enable
+  neither `--ple-local` nor `--ple`
+  [#2737](https://github.com/ucsd-progsys/liquidhaskell/issues/2737)
+
 ## 0.9.14.1.1 (2026-06-04)
 
 - Stop shadowing reflected signatures with assumed signatures
--- a/docs/mkDocs/docs/options.md
+++ b/docs/mkDocs/docs/options.md
@@ -60,6 +60,9 @@ liquid annotation
 {-@ automatic-instances theorem @-}
 ```
 
+The directive is read only under `--ple-local`. A module that sets neither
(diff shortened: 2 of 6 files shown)


facundominguez (commented):
LGTM!

Thanks @shalashaska117

PR #2742, opened 2026-08-20
Count the fields of every autosized type in the size measure

shalashaska117:
Follow-up to #2738, where this was left out on purpose.

`makeSizedDataCons` counted only the fields whose sort equals the type being defined, so in a group of mutually recursive autosized types the size was a constant: `autolen (Node f) = 1` for `Node :: Forest -> Tree`, and nothing that recurses on `Tree` through the `Forest` could be shown to terminate. It now counts the fields whose own type is autosized, so the size grows along the mutual edge:

```haskell
Node :: f:Forest -> {v:Tree | autolen v = 1 + autolen f}
Cons :: t:Tree -> f:Forest -> {v:Forest | autolen v = 1 + autolen t + autolen f}
```

This changes `autolen` for existing code: a constructor with a field of another autosized type gets a larger size than before. The constructor index and the non-negativity invariant are unchanged.

A field whose own type is not autosized, `[Tree]` in the module of #2736, still contributes nothing, so that module stays rejected for the reason given in #2738.

Tests: `tests/pos/AutoSizeMutual.hs`, which fails on `develop` with `autolen ... == 1` for the `Node` case. Docs: the autosize section of `specifications.md`.

The branch sits on top of #2738, so that PR's commits show up here until it merges.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Init.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Init.hs
@@ -110,6 +110,13 @@ addPolyInfo t = mkUnivs (go <$> as) ps t'
 makeDataConTypes :: Bool -> Var -> CG (Var, SpecType)
 makeDataConTypes allowTC x = (x,) <$> trueTy allowTC (varType x)
 
+-- | Refines the data constructors of every autosized type with an equation for
+--   the @autolen@ measure, and returns one invariant per refined type stating
+--   that the size is non-negative.
+--
+--   @dcts@ gives the unrefined type of every constructor, @specenv@ the types
+--   declared @autosize@ and @dcs@ the constructors in scope. Constructors of a
+--   type that was not declared @autosize@ are left alone.
 makeAutoDecrDataCons :: [(Id, SpecType)] -> S.HashSet TyCon -> [Id] -> ([LocSpecType], [(Id, SpecType)])
 makeAutoDecrDataCons dcts specenv dcs
   = (simplify rsorts, tys)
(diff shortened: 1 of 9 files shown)


liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Init.hs
```diff
@@ -131,15 +131,19 @@ idTyCon = fmap dataConTyCon . Ghc.isDataConId_maybe
 lenOf :: F.Symbol -> F.Expr
 lenOf x = F.mkEApp lenLocSymbol [F.EVar x]
 
-makeSizedDataCons :: [(Id, SpecType)] -> DataCon -> Integer -> (RSort, (Id, SpecType))
-makeSizedDataCons dcts x' n = (toRSort $ ty_res trep, (x, fromRTypeRep trep{ty_res = tres}))
+makeSizedDataCons :: S.HashSet TyCon -> [(Id, SpecType)] -> DataCon -> Integer -> (RSort, (Id, SpecType))
```
facundominguez, line 149: Maybe lets seize the opportunity to document what `makeSizedDataCons` and `makeAutoDecrDataCons` do, if you feel confident enough to explain it.

facundominguez (approved):
Looks great!

I suggested adding a piece of code documentation.

liquidhaskell-boot/src/Language/Haskell/Liquid/Constraint/Init.hs
shalashaska117: Done in 44e347f18. `makeAutoDecrDataCons` now says which types it refines and where the non-negativity invariant comes from, and `makeSizedDataCons` says what the constructor index is and which fields contribute to the size. The inline note about autosized fields moved into that Haddock.

Issue #2727, opened 2026-07-29
Higher-kinded class parameters are elaborated with wrong kind


jpyamamoto:
**Describe the bug**
Type classes with higher-kinded parameters are elaborated with kind `*`.

**To Reproduce**

## Example 1

```haskell
{-@ LIQUID "--typeclass" @-}
module HK where

class M m where
  {-@ ret :: a -> m a @-}
  ret :: a -> m a

class M m => VM m where
  {-@ lawRet :: x:a -> y:m a ->
        { ((y == ret x) => (y == ret x)) && ((y == ret x) => (y == ret x)) } @-}
  lawRet :: a -> m a -> ()
```
Verifying the previous code yields the following error:
```
HK.hs:1:1: error: [GHC-83865]
    • Expecting one more argument to ‘m’
      Expected a type, but ‘m’ has kind ‘* -> *’
    • In an expression type signature:
        forall m. VM m => forall a. a -> m -> a -> () -> Bool
      In the expression:
          (\ x
             -> (\ y -> (\ VV -> (==>) (==) y HK.ret x (==) y HK.ret x))) ::
            forall m. VM m => forall a. a -> m -> a -> () -> Bool
  |
1 | {-@ LIQUID "--typeclass" @-}
  | ^
```

The type `m` appears unapplied despite being of kind `* -> *`. The argument `a` is being dropped.

Running the same test using the [last version built with GHC 8](https://github.com/ucsd-progsys/liquidhaskell/releases/tag/v8.10.2-final) throws the following error:

```
tests/typeclasses/basic/HK.hs:7:1: error:
    * Illegal type specification for `HK.C:VM`
    HK.C:VM :: forall m .
               (M<[]> m) =>
               HK.lawRet##HK.C:VM:(forall a .
                                   x:a -> y:m a -> {VV : () | y == HK.ret HK.$p1VM##HK.C:VM x => y == HK.ret HK.$p1VM##HK.C:VM x}) -> {VV : (HK.VM m) | HK.$p1VM VV == HK.$p1VM##HK.C:VM}
    Sort Error in Refinement: {VV : Tuple | y == HK.ret HK.$p1VM##HK.C:VM x => y == HK.ret HK.$p1VM##HK.C:VM x}
    Unbound symbol HK.ret --- perhaps you meant: HK.C:VM ?
    Just constructor
    * 
  |
7 | class M m => VM m where
  | ^^^^^^^^^^^^^^^^^^^^^^^...

[...]

PR #2747, opened 2026-08-26
Fix elaboration of specs in classes with higher-kinded parameters

jpyamamoto:
This PR solves issue https://github.com/ucsd-progsys/liquidhaskell/issues/2727. This is a secondary issue found after merging PR https://github.com/ucsd-progsys/liquidhaskell/pull/2716, and is relevant to issue https://github.com/ucsd-progsys/liquidhaskell/issues/2450.

Tests were added too, including the minimal failing example, now passing correctly.

Specs mentioning higher-kinded class parameters (e.g. `y : m a` with `m :: Type -> Type`) failed under `--typeclass` with a GHC kind panic.

There were two bugs contributing to this situation.

## 1. `specTypeToLHsType` rendered type application as a function arrow

To elaborate a refinement, LH converts the spec into a Haskell expression with a type signature and sends it to GHC. `specTypeToLHsType` translated `RAppTy` (type application) using `nlHsFunTy` instead of `nlHsAppTy`.

## 2. `goPlug` inserted wrong binders

`plugHoles` merges the GHC type with the Liquid spec. When the Liquid spec is missing a quantifier that the GHC type has, `goPlug` would insert it using the binder from the GHC type. But since the body produced uses Liquid type variables, the inserted binder binds nothing, and the occurrences it should bind are free. `tyCompat` then rejects the type.

The fix renames the inserted binder to its Liquid counterpart, using a mapping between GHC and LH type variables.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Elaborate.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Elaborate.hs
@@ -695,7 +695,7 @@ specTypeToLHsType = \case
     REx _ tin tout -> nlHsFunTy (specTypeToLHsType tin) (specTypeToLHsType tout)
     RAppTy _ (RExprArg _) _ ->
       impossible Nothing "RExprArg should not appear here"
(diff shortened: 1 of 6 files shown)


facundominguez (approved):
LGTM!

PR #2754, opened 2026-08-29
Do not expand newtype constructors to their representations

### LINKED ISSUE #2332: Newtypes are unrolled when using embed
Currently, if we want to embed a specialization of some type in a LF theory, both `type` and `newtype` definitions get unrolled, breaking the abstraction barrier. E.g., defining bags as follows:

```haskell
import qualified Data.Map      as M

{-@ embed   Bag as Bag_t @-}
{-@ measure Bag_empty :: Int -> Bag a @-}
{-@ measure Bag_count :: Bag a -> a -> Int @-}

newtype Bag a = Bag { toMap :: M.Map a Int }

{-@ assume empty :: {v:Bag k | v = Bag_empty 0} @-}
empty :: Bag k
empty = Bag M.empty

{-@ assume get :: (Ord k) => k:k -> b:Bag k -> {v:Nat | v = Bag_count b k}  @-}
get :: (Ord k) => k -> Bag k -> Int
get k b = M.findWithDefault 0 k (toMap b)
```

generates an .fq file where `empty` and `get` internally reference `Data.Map` symbols, which causes trouble down the line when we want to interop with SMT Bag theory. To maintain the `Bag_t` abstraction, we must define Bags as `data Bag a = Bag { toMap :: M.Map a Int }`. This gives us proper theory definitions at the cost of runtime performance, as single-constructor `data` definitions are not optimized away in the same way as `newtype`s. Intuitively, we should probably only unroll `type` definitions, and treat `newtype` ones equally to `data`.

facundominguez:
It is unclear why the code was doing this, and it was causing sorts to be incorrect at least in the case of #2332.

There is a test `tests/errors/ElabLocation3.hs` that started failing. It was introduced in #2121, and presumably it is testing locations in error messages. But it relied on an elaboration error that no longer happens after this fix.

Fixes #2332


facundominguez:
@ranjitjhala, would you like to have any corrections done to ElabLocation3.hs or shall we remove the test?

If it is to stay, I think it would need a more explicit comment explaining what it is testing and how.

ranjitjhala:
Hi @facundominguez let's just remove the test if it works fine now!

Issue #2551, opened 2025-06-08
Liquid Haskell should accept `UnicodeSyntax`


0cjs:
Setting `{-# LANGUAGE UnicodeSyntax #-}` in a source file allows us to use, e.g., `→` and `∷` as `->` and `::`, but this doesn't apply to Liquid Haskell code, which even with `UnicodeSyntax` enabled still accepts only the latter. It would be nice if we could use `UnicodeSyntax` in Liquid Haskell as well.

(Given that we can't at the moment, I've had to abandon `UnicodeSyntax` for my projects that use Liquid Haskell, as having a mix of the two in the same file makes searching and the like more difficult.)

facundominguez:
Thanks @0cjs. 

[Here's](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/unicode_syntax.html) the list of syntax symbols that GHC supports. Of these, perhaps only `->`, `::`, and `=>` are relevant to LH.

It needs to be seen whether this only affects the LH parser, or LF parser is involved too.

AlecsFerra:
Can't we reuse the parser from GHC since we are a plugin? Honestly it feels pretty weird having syntax different from Haskell in the refinements

facundominguez:
The GHC parser would allow for more expressions than LH can handle. We would need to produce errors when the user uses unsupported syntax, produce good location info, and translate the Haskell AST to LH's.

The GHC parser doesn't allow for the syntax of abstract predicates and other decorations, so it could only be used to parse predicates, I suppose.

One inconvenience is that then parsing would be more affected by changes to the GHC AST.

Maybe it can be done. It deserves another issue.

AlecsFerra:
Uhm ok it sounds like a pain then I don't know if its worth it

vishwaksen-1:
Hi @facundominguez I'm working on this. I just submitted a PR on LF with my edits on the parser.
And I'm submitting this one now. 
Please check.

vishwaksen-1:
Hi @facundominguez 
What else should I add to address this issue fully?

facundominguez:
Hello! I think the implementation is as good as we needed for this issue. Thanks!

PR #2759, opened 2026-09-02
Tests for: LF-Support unicode syntax

vishwaksen-1:
Helping resolve issue #2551 

Adding necessary tests that complement with my latest PR in LF project where I added Unicode support. 

Please Check.

Thank you.

### DIFF
--- a/liquidhaskell-boot/tests/Parser.hs
+++ b/liquidhaskell-boot/tests/Parser.hs
@@ -286,6 +286,26 @@ testSucceeds =
        parseSingleSpec "x :: k:Int -> B (k + 1)" @?==
           "x :: k:Int -> (B {k + 1})"
 
+    , testCase "unicode :: " $
+       parseSingleSpec "assume foo ∷ a -> a " @?==
+          parseSingleSpec "assume foo :: a -> a "
+
+    , testCase "unicode -> " $
+       parseSingleSpec "assume foo :: a → b " @?==
+          parseSingleSpec "assume foo :: a -> b "
+
+    , testCase "unicode => " $
+       parseSingleSpec "assume foo :: a ⇒ b " @?==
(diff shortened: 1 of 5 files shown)


vishwaksen-1:
I think the github actions are failing because this PR(new tests) strongly depends on the code addition in LF submodule.

facundominguez (commented):
The parser tests look good, but I imagine that you also will want to add some test in `tests/basic/pos` to demonstrate that unicode operators can be used in LiquidHaskell annotations.

vishwaksen-1:
Hi @facundominguez 
How does it look now? 
I didn't really understand what the tests in `tests/basic/pos` or `neg` do but I think they are like examples to demonstrate the new capability. Correct me if I'm wrong.

Thank you.

facundominguez (approved):
LGTM!

liquidhaskell-boot/tests/Parser.hs
```diff
@@ -549,7 +569,7 @@ testErrorReporting =
               , "1 | ff :: {v:Nat  v >= 0 }"
               , "  |         ^"
               , "unexpected ':'"
```
facundominguez: CI says
```
<test>:1:17:
  |
1 | ff :: {v:Nat  v >= 0 }
  |                 ^^
  | unexpected ">="
  | expecting "->", "<:", "=>", '|', '→', '⇒', bareTyArgP, mmonoPredicateP, or monoPredicateP
```
perhaps differences in library versions could change the error message. The test doesn't fail for me in my local development environment.

facundominguez: I can reproduce the failure locally with
```
cabal update
cabal test --constraint="megaparsec>9.8.1" liquidhaskell-boot:liquidhaskell-parser
```

Issue #2756, opened 2026-09-01
"safeFromList with duplicates makeTyConInfo" on type instance with associated type


amigalemming:
Consider this module:
~~~~
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE EmptyDataDecls #-}
{-# LANGUAGE GADTs #-}
module Data.Array.Comfort.Shape.Check where

data Checked_
data Unchecked_

class Checking check where
   data Result check a
   switchCheck :: f Checked_ -> f Unchecked_ -> f check

data CheckSingleton check where
   Checked :: CheckSingleton Checked_
   Unchecked :: CheckSingleton Unchecked_


-- LiquidHaskell-9.14 fails on this one:
instance Checking Checked_ where
   newtype Result Checked_ a = CheckedResult {getChecked :: Either String a}
   switchCheck f _ = f

instance Checking Unchecked_ where
   newtype Result Unchecked_ a = UncheckedResult {getUnchecked :: a}
   switchCheck _ f = f
~~~~

When loaded with liquid-haskell HEAD at #01b76f5cd634e6664db4dc266537cecd31b8706e I get this error:
~~~~
**** ERROR *********************************************************************
safeFromList with duplicates [Data.Array.Comfort.Shape.Check.Result] makeTyConInfo
**** ERROR *********************************************************************
~~~~

This is a simplified version of:
https://hackage-content.haskell.org/package/comfort-array-0.5.6/candidate/src/src/Data/Array/Comfort/Shape/Check.hs

facundominguez:
Thanks, @amigalemming. I'm trying a fix in #2757.

amigalemming:
On Tue, 1 Sep 2026, Facundo Domínguez wrote:

> [1200667?s=20&v=4] facundominguez left a comment (ucsd-progsys/liquidhaskell#2756)
> 
> Thanks, @amigalemming. I'm trying a fix in #2757.

I still encounter the "duplicate" error with commit 
6cd79f384236cf801bbe0f22a377b39e2f29d590.

facundominguez:
Any chance that some artifact needs to be rebuilt in your development environment? I cannot reproduce it anymore and this example is part of the testsuite since #2758.

amigalemming:
On Tue, 1 Sep 2026, Facundo Domínguez wrote:

> [1200667?s=20&v=4] facundominguez left a comment (ucsd-progsys/liquidhaskell#2756)
> 
[...]

PR #2760, opened 2026-09-02
Change versioning policy so the git repo is ahead of hackage

facundominguez:
This is motivated by #2756.


facundominguez:
I need to remember merging the version bump for liquid-fixpoint ahead of merging this.

cc @amigalemming

PR #2762, opened 2026-09-03
Fix class methods starting with 'p' being mistaken for superclass bindings

jpyamamoto:
Under `--typeclass`, using an instance whose method name starts with 'p' failed with an unbound-symbol error:

```
typeclasses/pos/MethodPrefixP.hs:12:10: error:
    Illegal type specification for `MethodPrefixP.$fPNumBool`
    MethodPrefixP.$fPNumBool :: {VV : (MethodPrefixP.PNum GHC.Internal.Types.Bool) | VV == MethodPrefixP.$fPNumBool
                                                                                     && VV == MethodPrefixP.C:PNum $cpnat##a46p}
    Sort Error in Refinement: {VV : (MethodPrefixP.PNum bool) | VV == MethodPrefixP.$fPNumBool
                                                                && VV == MethodPrefixP.C:PNum $cpnat##a46p}
    Unbound symbol $cpnat##a46p --- perhaps you meant: papp4, papp6, Set_cap ?
    Just reflect
   |
12 | instance PNum Bool where
   |          ^^^^^^^^^

typeclasses/pos/MethodPrefixP.hs:13:3: error:
    Illegal type specification for `$cpnat`
    $cpnat :: lq1:GHC.Internal.Types.Bool -> {VV : GHC.Internal.Types.Int | VV == $cpnat##a46p lq1}
    Sort Error in Refinement: {VV : int | VV == $cpnat##a46p lq1}
    Unbound symbol $cpnat##a46p --- perhaps you meant: papp4, papp6, Set_cap ?
    Just assume
   |
13 |   pnat False = 0
   |   ^^^^
```

The reason is that `mkSymbol` in the `Bare.Typeclass` module will identify all names starting with `$cp` as superclass dictionary bindings. This is incorrect when a method's name starts with 'p'. We use a stricter check, using the fact that superclass bindings will include a number after `$cp` indicating the number of levels above the superclass is in.

A test is included.

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Typeclass.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Typeclass.hs
@@ -15,6 +15,7 @@ where
 
 import           Control.Monad                 ( forM, guard )
 import           Data.Bifunctor                (second)
+import           Data.Char                     (isDigit)
 import qualified Data.List                     as L
 import qualified Data.HashSet                  as S
 import           Data.Hashable                  ()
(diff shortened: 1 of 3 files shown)


facundominguez (commented):
Nice fix!

I don't think it is needed in this PR, but it would be nice to eventually test the assumptions on how GHC names dictionary fields in `liquidhaskell-boot/ghc-api-tests/GhcApiTests.hs`. That way, when GHC makes changes, we can discover it quickly.

Issue #1583, opened 2020-01-13
Information lost on dependent pair syntax


nikivazou:
The following is SAFE: 

```
{-@ safe :: ({v:Int| v <= 0 }, ()) -> {v:Int| v <= 0 } @-}
safe :: (Int, ()) -> Int
safe (xs, _) = xs
```

When dependent pair syntax is introduced it gets UNSAFE: 

```
{-@ unsafe :: (xs::{v:Int| v <= 0 }, ()) -> {v:Int| v <= 0 } @-}
unsafe :: (Int, ()) -> Int
unsafe (xs, _) = xs 
```

Desugaring the above syntax to abstract refinements is again safe: 

```
{-@ safe' :: ({v:Int| v <= 0 }, ())<{\_ _ -> true}> -> {v:Int| v <= 0 } @-}
safe' :: (Int, ()) -> Int
safe' (xs, _) = xs
```

ranjitjhala:
Yes I’ve seen strange things like this happen with the pair syntax ...
would be great to fix

PR #2764, opened 2026-09-08
Keep the refinements when parsing dependent pairs into abstract refinements

nikivazou:
fix for #1583


tests/pos/T1583.hs
```diff
@@ -0,0 +1,20 @@
...
+incr :: Int -> (Int, Int)
+incr x = (x+1, x+2)
+
+{-@ incr' :: n:Int -> ({v:Int | v > n}, Int) <{\x y -> y > x}> @-}
+incr' :: Int -> (Int, Int)
+incr' x = (x+1, x+2)
+
+{-@ greater :: n:Int -> m:{Int | m > n} @-}
+greater :: Int -> Int
+greater x = y where (y,_) = incr x -- changing incr' to incr breaks the proof
```
facundominguez: The comment doesn't seem to be true with the current fix.
```suggestion
greater x = y where (y,_) = incr x
```

tests/pos/T1583.hs
```diff
@@ -0,0 +1,20 @@
+{-# OPTIONS_GHC -fplugin=LiquidHaskell #-}
+{-@ LIQUID "--ple" @-}
+
+module T1583 where
```
facundominguez, line 5:
```suggestion
-- | Test some variations of dependent tuples for #1583
module T1583 where
```

facundominguez (approved):
LGTM!

PR #2765, opened 2026-09-11
checkRefinements

nikivazou:
Checking that the predicates in the refinement types do not violate their specifications. 
To implement it, the predicates are turned into Boolean expressions and we check that they type check. 

Currently it is on by default, but it is flagged (if we want to keep it, we could make it optional).

### DIFF
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Elaborate.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Elaborate.hs
@@ -16,6 +16,7 @@
 module Language.Haskell.Liquid.Bare.Elaborate
   ( fixExprToHsExpr
   , elaborateSpecType
+  , elaborateSpecTypeWith
   -- , buildSimplifier
   )
 where
@@ -287,7 +288,18 @@ elaborateSpecType
   -> SpecType
   -> TcRn SpecType
 elaborateSpecType coreToLogic simplifier t = GM.withWiredIn $ do
-  (t', xs) <- elaborateSpecType' (pure ()) coreToLogic simplifier t
+  elaborateSpecTypeWith (\_ _ -> pure ()) coreToLogic simplifier t
+
+-- | Elaborate a specification and expose the Core expression generated for
+-- every non-trivial refinement together with its refined checking type.
+elaborateSpecTypeWith
+  :: (CoreExpr -> SpecType -> TcRn ())
+  -> (CoreExpr -> F.Expr)
+  -> (CoreExpr -> TcRn CoreExpr)
+  -> SpecType
+  -> TcRn SpecType
+elaborateSpecTypeWith collect coreToLogic simplifier t = GM.withWiredIn $ do
+  (t', xs) <- elaborateSpecType' collect (pure ()) coreToLogic simplifier t
   case xs of
     _ : _ -> panic
       Nothing
@@ -295,13 +307,14 @@ elaborateSpecType coreToLogic simplifier t = GM.withWiredIn $ do
     _ -> pure t'
 
 elaborateSpecType'
-  :: PartialSpecType
+  :: (CoreExpr -> SpecType -> TcRn ())
+  -> PartialSpecType
   -> (CoreExpr -> F.Expr) -- core to logic
   -> (CoreExpr -> TcRn CoreExpr)
   -> SpecType
   -> TcRn (SpecType, [F.Symbol]) -- binders for dictionaries
                    -- should have returned Maybe [F.Symbol]
-elaborateSpecType' partialTp coreToLogic simplify t =
+elaborateSpecType' collect partialTp coreToLogic simplify t =
   case F.notracepp "elaborateSpecType'" t of
     RVar (RTV tv) (MkUReft reft@(F.Reft (vv, _oldE)) p) -> do
       elaborateReft
@@ -315,8 +328,8 @@ elaborateSpecType' partialTp coreToLogic simplify t =
       let partialFunTp =
             Free (RFunF bind i (wrap $ specTypeToPartial tin) (pure ()) ureft) :: PartialSpecType
           partialTp' = partialTp >> partialFunTp :: PartialSpecType
-      (eTin , bs ) <- elaborateSpecType' partialTp coreToLogic simplify tin
-      (eTout, bs') <- elaborateSpecType' partialTp' coreToLogic simplify tout
+      (eTin , bs ) <- elaborateSpecType' collect partialTp coreToLogic simplify tin
+      (eTout, bs') <- elaborateSpecType' collect partialTp' coreToLogic simplify tout
       let buildRFunContTrivial
             | isClassType tin, dictBinder : bs0' <- bs' = do
               let (eToutRenamed, canonicalBinders) =
@@ -368,6 +381,7 @@ elaborateSpecType' partialTp coreToLogic simplify t =
     -- support for RankNTypes/ref
     RAllT (RTVar tv ty) tout ureft@(MkUReft ref@(F.Reft (vv, _oldE)) p) -> do
       (eTout, bs) <- elaborateSpecType'
+        collect
         (partialTp >> Free (RAllTF (RTVar tv ty) (pure ()) ureft))
         coreToLogic
         simplify
@@ -387,6 +401,7 @@ elaborateSpecType' partialTp coreToLogic simplify t =
     -- todo: might as well print an error message?
     RAllP pvbind tout -> do
       (eTout, bts') <- elaborateSpecType'
+        collect
         (partialTp >> Free (RAllPF pvbind (pure ())))
         coreToLogic
         simplify
@@ -398,7 +413,7 @@ elaborateSpecType' partialTp coreToLogic simplify t =
       | isClass tycon -> pure (t, [])
       | otherwise -> do
         args' <- mapM
-          (fmap fst . elaborateSpecType' partialTp coreToLogic simplify)
+          (fmap fst . elaborateSpecType' collect partialTp coreToLogic simplify)
           args
         elaborateReft
           (reft, t)
@@ -407,8 +422,8 @@ elaborateSpecType' partialTp coreToLogic simplify t =
             pure (RApp tycon args' pargs (MkUReft (F.Reft (vv, ee)) p), bs')
           )
     RAppTy arg res ureft@(MkUReft reft@(F.Reft (vv, _)) p) -> do
-      (eArg, bs ) <- elaborateSpecType' partialTp coreToLogic simplify arg
-      (eRes, bs') <- elaborateSpecType' partialTp coreToLogic simplify res
+      (eArg, bs ) <- elaborateSpecType' collect partialTp coreToLogic simplify arg
+      (eRes, bs') <- elaborateSpecType' collect partialTp coreToLogic simplify res
       let (eResRenamed, canonicalBinders) =
             canonicalizeDictBinder bs (eRes, bs')
       elaborateReft
@@ -424,8 +439,8 @@ elaborateSpecType' partialTp coreToLogic simplify t =
         )
     -- todo: Existential support
     REx bind allarg ty -> do
-      (eAllarg, bs ) <- elaborateSpecType' partialTp coreToLogic simplify allarg
-      (eTy    , bs') <- elaborateSpecType' partialTp coreToLogic simplify ty
+      (eAllarg, bs ) <- elaborateSpecType' collect partialTp coreToLogic simplify allarg
+      (eTy    , bs') <- elaborateSpecType' collect partialTp coreToLogic simplify ty
       let (eTyRenamed, canonicalBinders) = canonicalizeDictBinder bs (eTy, bs')
       pure (REx bind eAllarg eTyRenamed, canonicalBinders)
     -- YL: might need to filter RExprArg out and replace RHole with ghc wildcard
@@ -463,6 +478,7 @@ elaborateSpecType' partialTp coreToLogic simplify t =
             (hsTypeToHsSigWcType (specTypeToLHsType querySpecType))
         eeWithLamsCore <- GM.elabRnExpr exprWithTySigs
         eeWithLamsCore' <- simplify eeWithLamsCore
+        collect eeWithLamsCore' querySpecType
         let
           (_, tyBinders) =
             collectSpecTypeBinders
--- a/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
+++ b/liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
@@ -17,6 +17,8 @@ module Language.Haskell.Liquid.GHC.Plugin (
 
 import qualified Liquid.GHC.API         as O
 import           Liquid.GHC.API         as GHC hiding (Type)
+import           GHC.Data.FastString    (LexicalFastString (..))
+import           GHC.Types.Name         (setNameLoc)
 import qualified Text.PrettyPrint.HughesPJ               as PJ
 import qualified Language.Fixpoint.Types                 as F
 import qualified Language.Haskell.Liquid.GHC.Misc        as LH
@@ -611,7 +613,9 @@ processModule LiquidHaskellContext{..} = do
 
       Right ((warnings, targetSpec, liftedSpec), bareSpec) -> do
         liftIO $ mapM_ (printWarning logger) warnings
-        let targetInfo = TargetInfo targetSrc targetSpec
+        (normalizedTargetSrc, normalizedTargetSpec) <- liftIO $
+          normalizeRefinementChecks moduleCfg hscEnv modGuts targetSrc targetSpec
+        let targetInfo = TargetInfo normalizedTargetSrc normalizedTargetSpec
 
         debugLog $ "bareSpec ==> "   ++ show bareSpec
         debugLog $ "liftedSpec ==> " ++ show liftedSpec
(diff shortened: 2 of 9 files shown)


tests/neg/CheckRefinements.hs
```diff
@@ -0,0 +1,34 @@
+{-# OPTIONS_GHC -fplugin=LiquidHaskell #-}
+
+{- LIQUID "--check-refinements" @-}
```
facundominguez, line 3: As it stands, this is a test which doesn't fail. If testing for failure, `--expect-error-containing=...` should be used.

liquidhaskell-boot/src/Language/Haskell/Liquid/GHC/Plugin.hs
```diff
@@ -629,6 +633,41 @@ processModule LiquidHaskellContext{..} = do
       `Ex.catch` (\(e :: Error) -> reportErrs [e])
       `Ex.catch` (\(es :: [Error]) -> reportErrs es)
 
+normalizeRefinementChecks :: Config -> HscEnv -> ModGuts -> TargetSrc -> TargetSpec -> IO (TargetSrc, TargetSpec)
```
facundominguez, line 636: It would be needed a comment explaining what normalizing a refinement check is.

It also would help linking to an explanation of what refinement checks are.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Elaborate.hs
```diff
@@ -287,21 +288,33 @@ elaborateSpecType
   -> SpecType
   -> TcRn SpecType
 elaborateSpecType coreToLogic simplifier t = GM.withWiredIn $ do
-  (t', xs) <- elaborateSpecType' (pure ()) coreToLogic simplifier t
+  elaborateSpecTypeWith (\_ _ -> pure ()) coreToLogic simplifier t
+
+-- | Elaborate a specification and expose the Core expression generated for
+-- every non-trivial refinement together with its refined checking type.
```
facundominguez, line 294:
```suggestion
-- | Elaborate a specification to make type class dictionaries explicit,
--  and expose the Core expression generated for every non-trivial
--  refinement together with its refined checking type.
```

liquidhaskell-boot/src/Language/Haskell/Liquid/UX/Config.hs
```diff
@@ -88,6 +88,7 @@ data Config = Config
   , reflection               :: Bool       -- ^ Allow "reflection"; switches on "--higherorder" and "--adt"
   , compileSpec              :: Bool       -- ^ Only "compile" the spec -- into .bspec file -- don't do any checking.
   , typeclass                :: Bool        -- ^ enable typeclass support.
+  , checkRefinements         :: Bool        -- ^ check refinement predicates with Liquid's type checker.
```
facundominguez, line 91: Link here to the implementation of the Liquid's type checker.

liquidhaskell-boot/src/Language/Haskell/Liquid/UX/CmdLine.hs
```diff
@@ -332,6 +333,8 @@ lhOptions =
       "Only compile specifications (into .bspec file); skip verification"
   , opt [] ["typeclass"] (NoArg $ fm $ \c -> c { typeclass = True })
       "Enable typeclass support"
+  , opt [] ["check-refinements"] (NoArg $ fm $ \c -> c { checkRefinements = True })
+      "Check refinement predicates with Liquid's refinement type checker"
```
facundominguez, line 337: Maybe link here to the user documentation explaining what the refinement type checker does.

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Specs.hs
```diff
@@ -255,9 +256,18 @@ data GhcSpecSig = SpSig
   , gsTexprs   :: ![(Var, LocSpecType, [F.Located F.Expr])]  -- ^ Lexicographically ordered expressions for termination
   , gsRelation :: ![(Var, Var, LocSpecType, LocSpecType, RelExpr, RelExpr)]
   , gsAsmRel   :: ![(Var, Var, LocSpecType, LocSpecType, RelExpr, RelExpr)]
+  , gsReftChecks :: ![RefinementCheck]             -- ^ Synthetic Core checks for refinement predicates
   }
   deriving Show
 
+data RefinementCheck = RefinementCheck
```
facundominguez, line 263: Explain what a refinement check is and what the representation means (?)

facundominguez (commented):
Hello! I made a pass over the implementation to flag the places where more explanation would be helpful.

User documentation and an overview note in the implementation explaining how the feature is implemented at the various places would be needed.

But I have the impression that I might have jumped too eagerly to review. My apologies if so.

nikivazou:
Thanks for the comments @facundominguez ! I am still looking into repairing the text suite! Will update once they are fixed!

tests/pos/Rest.hs
```diff
@@ -113,7 +113,7 @@ elemAtThroughAppend 0 (_:_) ys = ()
 elemAtThroughAppend i (_:xss) ys = elemAtThroughAppend (i - 1) xss ys
 
 {-@
-predicate WellTyped E CTX = checkBindings CTX E && numFreeVarsExp E <= length CTX
+predicate WellTyped E CTX = numFreeVarsExp E <= length CTX && checkBindings CTX E
```
facundominguez, line 116: The test still seems to pass without this change.

tests/pos/Rest.hs
```diff
@@ -159,11 +159,12 @@ exprType (BoolE _) = TBool
 reflect checkBindings
 checkBindings
   :: ctx : [Ty]
-  -> { e : Exp | numFreeVarsExp e <= length ctx }
+  -> e : Exp
```
facundominguez, line 162: Reverting this causes an error that is very hard to diagnose:
<details>

```
tests/pos/Rest.hs:180:6: error:
    Liquid Type Mismatch
    .
    The inferred type
      VV : {lq_tmp$x##2225 : Rest.Exp | lq_tmp$x##2225 == e##a22e
                                        && Rest.numFreeVarsExp lq_tmp$x##2225 >= 0}
    .
      VV : {VV##2224 : Rest.Exp | Rest.numFreeVarsExp VV##2224 <= Rest.length ctx0##a22c}
    .
    in the context
      e##a22e : {e##a22e : Rest.Exp | Rest.numFreeVarsExp e##a22e >= 0}
       
      ctx1##a22d : {ctx1##a22d : [Rest.Ty] | GHC.Types_LHAssumptions.len ctx1##a22d >= 0
                                             && Rest.length ctx1##a22d >= 0}
       
      ?a : {?a : GHC.Internal.Types.Int | ?a == Rest.numFreeVarsExp e##a22e
                                          && ?a >= 0}
       
      ?b : {?b : GHC.Internal.Types.Int | ?b == Rest.length ctx0##a22c
                                          && ?b >= 0}
       
      ctx0##a22c : {ctx0##a22c : [Rest.Ty] | GHC.Types_LHAssumptions.len ctx0##a22c >= 0
                                             && Rest.length ctx0##a22c >= 0}
    Constraint id 25
    |
180 |   :: ctx0 : [Ty]
    |
```
</details>

I'm concerned about users being frustrated by this, and I think it could be a good reason to turn off the feature by default. But this needs to be weighed against the advantages of the current default, which I'm not familiar with yet.

I find the error hard because the location is imprecise, and makes hard to know where the constraint originates from. An approachable error message could be, for instance:

```
... same constraint ...
... same environment extended with:

    WellTyped : {e:Exp | numFreeVarsExp e <= length ctx} -> ctx:[Ty] -> Bool

    Constraint id 25
    |   -> e : Exp
183 |   -> { WellTyped e ctx0 <=>
    |                  ^
```

tests/neg/CheckRefinements.hs
```diff
@@ -0,0 +1,35 @@
+{-# OPTIONS_GHC -fplugin=LiquidHaskell #-}
+{-@ LIQUID "--expect-any-error" @-}
+
+{- LIQUID "--check-refinements" @-}
```
facundominguez: While refinement checks are enabled by default, this comment is confusing and should probably be removed.

tests/neg/CheckRefinements.hs
```diff
@@ -0,0 +1,35 @@
+{-# OPTIONS_GHC -fplugin=LiquidHaskell #-}
+{-@ LIQUID "--expect-any-error" @-}
```
facundominguez: If the test start failing for a reason unrelated to refinement checks, this expectation won't detect it. `--expect-error-containing=...` is needed to make the expectation explicit.

tests/pos/Rest.hs
nikivazou: I agree that the error message is clear and that the flag should be optional. 
Also because it is an expensive check that currently turns the specs to Haskell code. If the user wants, they can write the specs as code to begin with. 
I was just super curious to see how many of the tests fail.

nikivazou:
So, I need to: 
- [x] make the flag optional 
- [x] Give specific error to [tests/neg/CheckRefinements.hs](https://github.com/ucsd-progsys/liquidhaskell/pull/2765/files/1c33542fb5903c17181e13db25c543a305a0a38b#diff-94682e38789dd8d4cce2b2a35269bf46ed395acf201449b2fd93f0aa42f53dad)
- [x] Revert the [tests/pos/Rest.hs](https://github.com/ucsd-progsys/liquidhaskell/pull/2765/files/1c33542fb5903c17181e13db25c543a305a0a38b#diff-06e499d2acd31d27de754fb1c0e0177f3aee14a88a8762280b229fda29c16d82)
- [x] add documentation in `docs/mkDocs/docs/options.md`

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Elaborate.hs
```diff
@@ -287,21 +288,34 @@ elaborateSpecType
...
   -> TcRn SpecType
 elaborateSpecType coreToLogic simplifier t = GM.withWiredIn $ do
-  (t', xs) <- elaborateSpecType' (pure ()) coreToLogic simplifier t
+  elaborateSpecTypeWith (\_ _ -> pure ()) coreToLogic simplifier t
+
+-- | Elaborate a specification to make type class dictionaries explicit,
+-- and expose the Core expression generated for every non-trivial
+-- refinement together with its refined checking type.
+elaborateSpecTypeWith
+  :: (CoreExpr -> SpecType -> TcRn ())
```
facundominguez: Would be worth making explicit what this parameter does, and what it gets as inputs.

liquidhaskell-boot/src/Language/Haskell/Liquid/Bare/Elaborate.hs
```diff
@@ -287,21 +288,34 @@ elaborateSpecType
   -> SpecType
   -> TcRn SpecType
 elaborateSpecType coreToLogic simplifier t = GM.withWiredIn $ do
-  (t', xs) <- elaborateSpecType' (pure ()) coreToLogic simplifier t
+  elaborateSpecTypeWith (\_ _ -> pure ()) coreToLogic simplifier t
+
+-- | Elaborate a specification to make type class dictionaries explicit,
+-- and expose the Core expression generated for every non-trivial
+-- refinement together with its refined checking type.
```
facundominguez, line 295:
```suggestion
-- | Elaborate a specification to make type class dictionaries explicit,
-- and expose the Core expression generated for every non-trivial
-- refinement together with its refined checking type.
-- 
-- @elaborateSpecTypeWith collect coreToLogic simplify (x0:t0 -> ... -> v:tn)@ calls
-- @collect et (\<env> -> ti')@ for each @i@, where @ti'@ is @ti@ with the
-- refinement predicate replaced by @true@, @et@ is the elaborated Core
-- expression of @\<env> -> ti@, and @env@ are the local bindings @xi@, @v@ in
-- scope.
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Specs.hs
```diff
@@ -255,9 +256,28 @@ data GhcSpecSig = SpSig
...
+  , gsReftChecks :: ![RefinementCheck]             -- ^ Synthetic Core checks for refinement predicates
   }
   deriving Show
 
+-- | Checking that the predicates of the refinements do not violate their specifications.
+--
+-- For each refinement that can be translated to Haskell, elaboration produces
+-- a synthetic function whose arguments are the values in scope at the point of
+-- the refinement and whose result is the predicate as a 'Bool'.  For example,
+-- the refinement @x:Int -> {v:Int | v > x}@ generates @\x v -> v > x@.
```
facundominguez, line 271:
```suggestion
-- the refinement @x:Int -> {v:Int | v > x}@ generates @\x v -> v > x@.
-- Elaboration is implemented in 'Language.Haskell.Liquid.Bare.Elaborate.elaborateSpecTypeWith'.
```

liquidhaskell-boot/src/Language/Haskell/Liquid/Types/Specs.hs
```diff
@@ -255,9 +256,28 @@ data GhcSpecSig = SpSig
   , gsTexprs   :: ![(Var, LocSpecType, [F.Located F.Expr])]  -- ^ Lexicographically ordered expressions for termination
   , gsRelation :: ![(Var, Var, LocSpecType, LocSpecType, RelExpr, RelExpr)]
   , gsAsmRel   :: ![(Var, Var, LocSpecType, LocSpecType, RelExpr, RelExpr)]
+  , gsReftChecks :: ![RefinementCheck]             -- ^ Synthetic Core checks for refinement predicates
   }
   deriving Show
 
+-- | Checking that the predicates of the refinements do not violate their specifications.
+--
+-- For each refinement that can be translated to Haskell, elaboration produces
```
facundominguez, line 268:
```suggestion
-- Since constraint generation is only done by analyzing Core, checking
-- predicates of refinement types requires translating them to Core.
--
-- For each refinement that can be translated to Haskell, elaboration produces
```

facundominguez (commented):
> - [ ] Revert the tests/pos/Rest.hs

Your changes are sensible there, though. I think the test is in better shape with them.

I suggested some more edits to the documentation.

Also, `docs/mkDocs/docs/options.md` could have an example explaining
the effect of the flag. I find the flag explanation a bit too abstract for me to grok it immediately without looking at the tests:

    Check refinement predicates do not violate the refined pecifications

Is LH unsound without these checks? I couldn't find a way to prove false without them :)

nikivazou:
Thanks!!!! (Changed my todo list). 

I do not think it is unsound, no. 
It is more that, it is very difficult to explain the meaning of such (not well typed) specification... 

The https://github.com/ucsd-progsys/liquidhaskell/blob/develop/tests/benchmarks/popl18/ple/pos/NatInduction.hs#L9
is a good example found by this check.... The inductive hypothesis, is indeed "intended" to be applied only to positive numbers. (If I am not wrong) nothing bad can happen if you apply to zero too...

AlecsFerra:
Can we make this enabled by `--modern` too if it is not going to be enabled by default?

AlecsFerra:
Re unsoundness: it is not unsound as you can build a model of LH only requires the refinement to be well-typed wrt to the base type system (actually i think you can get by with even less but i haven't checked all the details).

One effect that having ill-typed refinements can make PLE loop even in transparent (in the sense of refinement reflection) environments, interestingly it stil won't produce unsound equations.

nikivazou:
I actually expected PLE to diverge in the following, though it does not.... It just says unsafe.

```
foo :: Nat -> Int 
foo i
  | i == 0        = 0 
  | i >= 0        = foo (i-1) 
  | otherwise = foo i 

bar :: () -> { foo (-42) == 0}
bar () = () 
```

Also, not making it part of `--modern` now, let's test it/understand it a little first...

nikivazou:
I believe I did all the requested edits, so if CI agrees, we can merge!

docs/mkDocs/docs/options.md
```diff
@@ -476,6 +476,30 @@ with the `--no-termination` option.
...
+test _ = undefined
+
+{-@ get :: n:{Int | 0 <= n } -> {v:List Int | n < mylen v} -> Int @-}
+```
+
+The call `get 0 Nil` in the refinement of `test` violates the specification of `get`. This error will only be detected when the `check-refinements` flag is on. 
+
+No checking such refinements cannot lead to unsoundness, since the predicate 
+language is interpreted as a logical language, thus all functions are total. 
+Yet, it can lead to unexpected behaviours and make PLE diverge.
```
facundominguez, line 499: I'm curious about how PLE would diverge.

docs/mkDocs/docs/options.md
```diff
@@ -476,6 +476,30 @@ with the `--no-termination` option.
...
+
+{-@ get :: n:{Int | 0 <= n } -> {v:List Int | n < mylen v} -> Int @-}
+```
+
+The call `get 0 Nil` in the refinement of `test` violates the specification of `get`. This error will only be detected when the `check-refinements` flag is on. 
+
+No checking such refinements cannot lead to unsoundness, since the predicate 
+language is interpreted as a logical language, thus all functions are total. 
+Yet, it can lead to unexpected behaviours and make PLE diverge. 
+Since there cannot be unsoundness and such checks slow down verification time, they are not done by default.
```
facundominguez: It is important to mention all aspects of how experience will degrade when this is enabled.
```suggestion
The checks slow down verification time, the error messages can be difficult to understand still, and soundness is not at stake, so the feature is disabled by default. 
```

facundominguez:
> I actually expected PLE to diverge in the following, though it does not.... It just says unsafe.

After finding `foo (-42) == 0`, it would unfold `foo` to get `foo (-42) == 0` again. The equality is already among the known equalities, and therefore it would conclude that a fixpoint set of equalities was reached.

facundominguez (approved):
LGTM. I suggested one more edit in `options.md`.

nikivazou:
How about 

```
{-@ reflect foo @-}
{-@ foo :: Nat -> Int @-}
foo :: Int -> Int 
foo i
  | i == 0        = 0 
  | i >= 0        = foo (i-1) 
  | otherwise = foo (i-1) 

{-@ bar :: () -> { foo (-42) == 0} @-}
bar :: () -> () 
bar () = () 
```

The idea is that termination is proved only for the arguments that satisfy their preconditions (in my example `0 <= i`), while PLE sees all the Haskell function definitions (`i < 0`).

facundominguez:
Well, in that one PLE does not terminate indeed :)
Thanks for the example.

